PR #21, opened 2017-06-30
more functions

zzhangjii:


ashutoshbondre:
29th June, more than 5 buttons added, still building it up.

PR #160, opened 2017-08-29
fixed bugs and nbdiff pre-setup for ipynb 

zzhangjii:
fixed the json.loads() function conflicts for python3.5 and lower;
for git diff, plan to use nbdiff for ipynb files, pre-setup the environment by Vidar's help,


package.json
```diff
@@ -32,6 +32,7 @@
     "@types/react": "^16.0.0",
     "@types/react-dom": "^15.5.2",
     "jquery": "^2.2.0",
+    "nbdime": "^0.4.0",
```
vidartf: Nice to see this PR, but I think this line should be removed (the dependency goes the other way).

PR #186, opened 2018-03-22
Improve error message

TejasQ:
## In this PR
- Added basic style updates to error message when current folder is not a `git` repository.
- Updated error message to be more explanatory for a better experience.


zzhangjii:
Thx for the PR

PR #250, opened 2018-10-17
Git clone functionality: frontend

jaipreet-s:
### Notes
Git clone functionality: frontend  …
This change includes the frontend changes to enable the "Cloning" functionality in the Git plugin

* Includes a button added to the File Browser toolbar
* Includes a Modal that is invoked on clicking the said button.
* Includes the API call invoked and the form data collection for the clone URI
* Error handling when the Git Clone call fails.

Some notes:
* No syntax validation of URL done. It will be done on the server-side
* Server-side API is not fully fleshed out (Next task). This impl joins the various moving parts.
* Disabling the button when the current directory is a Git directory will be a separate commit.

### Screenshots of UX

git_clone_button

git_clone_modal

git_clone_error


package.json
```diff
@@ -13,7 +13,8 @@
   "scripts": {
     "build": "tsc",
     "clean": "rimraf lib",
-    "prepublish": "npm run build"
+    "prepublish": "npm run build",
+    "watch": "tsc -w --jsx react"
```
blink1073: This config should be part of `tsconfig.json` instead if it isn't already there

src/git.ts
```diff
@@ -122,10 +122,36 @@ function httpGitRequest(
...
 
+    /** Placeholder for the Git Clone API. */
+    async clone(path: string, url: string): Promise<GitCloneResult> {
+        try {
+            let response = await httpGitRequest('/git/clone', 'POST', {
+                current_path: path,
+                clone_url: url
+            });
+            if (response.status !== 200) {
+                return response.text().then(data => {
```
blink1073: This function is already `async`, I think this would be cleaner as `const data = await response.text()`

jaipreet-s: Fair point -  This was a pattern in the other Git methods as well; I refactored some of them along with this.

package.json
jaipreet-s: The `--jsx react` option can be omitted since its already in `tsconfig.json`, but I still need a `watch` target here for `jlpm watch` to work, right?

```suggestion
    "watch": "tsc -w"
```

blink1073:
Thanks!

PR #264, opened 2018-10-23
Fixes #263 by putting path in double quotes

### LINKED ISSUE #263: For directories with whitespace in the name, this command fails.
https://github.com/jupyterlab/jupyterlab-git/blob/e93f8267307c83b534990e0eb0e12820b99dc794/src/gitMenuCommands.ts#L70

Suggest to escape the string before executing the command.

neelamgehlot:
### Notes
Fixes #263 by putting path in double quotes which would escape any special character.

#### Tested with following directory names:
```
directory .  ( with  )  * spacial ? chars
directory with    irregular   spaces
directory-with-"double" quotes and space
directory-with-""double" quotes and space
directory-with-'single' quote and space
```


neelamgehlot:
Fixes #263

Issue #245, opened 2018-10-15
Improve history overview


weihwang:
Proposal for full width tabbed history view, which is related to how we may think about push/pull actions. We need a more detailed higher level view of where the latest changes are in local/remote branches. Fly-out full detail on mouse-over provides an easier "browse" experience. Adding a link at the bottom of a default 10 item list gives users a link to the full log if needed.

https://www.figma.com/file/UyNpeQNuTj6Uv74MoEJDh3ji/Git-extension?node-id=517%3A1718

![image](https://user-images.githubusercontent.com/551333/46969738-dfd29c80-d06b-11e8-9451-4a9dc7a7d33f.png)

dhirschfeld:
I was thinking that it might be very cool you you could select two nodes in the history to view the diff between them.

tgeorgeux:
I put a similar response in #244, what is the upside of fly-out versus just expanding/collapsing each history field?

tgeorgeux:
I definately agree with the onclick. I think the fly-out as an overview has
merit, but the interactions should stay within the sidebar.

@dhirschfeld , I think that diff idea is a great idea! Can you open a
separate issue with the 'enhancement' tag, so we can track where this is at
and will be in the existing milestones without losing that suggestion?

jaipreet-s:
cc @saulshanabrook

saulshanabrook:
I would be in favor of the on click over the popout on hover (I find hovering sorta fiddly often).

But am happy to get started on either approach if we consensus on which to pursue.

tgeorgeux:
@weihwang @jaipreet-s Do you have additional thoughts on popout vs on-click expanding?

weihwang:
If it's not possible today to create a popout, and the general feedback leans towards inline expansion of details, I think we can start with that version. I have a mock below to expand-on-click. I needed to add more details (usesrname, commit hash) into the collapsed view in this case to create a more seamless "expand" to minimize disrupting content when going to the detail state.

jaipreet-s:
Inline expansion looks better to me as well.

In general, I'd suggest taking this one incrementally

1. Get the "summary view" in place (This'll also unblock the push/pull task)
2. Add "detailed view" on inline expansion
3. Add extra details like the `View mode in Terminal option`

tgeorgeux:
Sounds like a good plan to me.

dhirschfeld:
[...]

PR #270, opened 2018-11-21
Improve history overview

saulshanabrook:
This addresses #245 by making a seperate history tab.

- [x] Add ability to expand commit details
- [x] Add screenshots of new behavior

It implements all the features requested by #245 except for the graphical branch view on the left hand side. 

It also is not a pixel perfect translation of the mockups. I reused as much of the existing styles and components as possible. As a result, there should likely be some cleanup after this to reflect the component name updates. (i.e. `HistorySideBar` is no longer a side bar).

### DIFF
--- a/src/components/HistorySideBar.tsx
+++ b/src/components/HistorySideBar.tsx
@@ -21,28 +21,12 @@ export interface IHistorySideBarProps {
   diff: any;
 }
 
-/** Interface for PastCommits component state */
-export interface IHistorySideBarState {
-  activeNode: number;
-}
+
 
 export class HistorySideBar extends React.Component<
   IHistorySideBarProps,
-  IHistorySideBarState
+  {}
 > {
-  constructor(props: IHistorySideBarProps) {
-    super(props);
-
-    this.state = {
-      activeNode: -1
-    };
-  }
-
-
-  updateActiveNode = (index: number): void => {
-    this.setState({ activeNode: index });
-  };
-
   render() {
     if (!this.props.isExpanded) {
       return null;
--- a/src/components/HistorySideBar.tsx
+++ b/src/components/HistorySideBar.tsx
@@ -4,20 +4,15 @@ import { SingleCommitInfo } from "../git";
 
 import {
   historySideBarStyle,
-  historySideBarExpandedStyle
 } from "../componentsStyle/HistorySideBarStyle";
 
-import { classes } from "typestyle";
 
 import * as React from "react";
 
 /** Interface for PastCommits component props */
 export interface IHistorySideBarProps {
-  currentFileBrowserPath: string;
   pastCommits: SingleCommitInfo[];
   isExpanded: boolean;
-  setShowList: Function;
-  getPastCommit: Function;
 }
 
 /** Interface for PastCommits component state */
@@ -37,11 +32,6 @@ export class HistorySideBar extends React.Component<
     };
   }
 
-  getSideBarClass(): string {
-    return this.props.isExpanded
-      ? classes(historySideBarExpandedStyle, historySideBarStyle)
-      : historySideBarStyle;
-  }
 
   updateActiveNode = (index: number): void => {
     this.setState({ activeNode: index });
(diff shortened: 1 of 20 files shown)


src/components/HistorySideBar.tsx
```diff
@@ -1,23 +1,24 @@
-import { PastCommitNode } from './PastCommitNode';
+import { PastCommitNode } from "./PastCommitNode";
 
-import { SingleCommitInfo } from '../git';
```
jaipreet-s: This looks great! I'm starting to review this more however I'd love to see some Unit Tests for this component.  We already have some for the branch component so you could build off that.

src/components/PastCommitNode.tsx
```diff
@@ -1,114 +1,139 @@
...
     return (
-      <div key={this.props.index}>
-        <div
-          className={this.getPastCommitNodeClass()}
-          onClick={() => this.handleClick()}
-        >
-          <span className={this.getContentClass()}>
-            {this.props.isVisible && this.getContent()}
-          </span>
+      <div
```
jaipreet-s: Same for this component.

src/components/SinglePastCommitInfo.tsx
```diff
@@ -65,10 +59,31 @@ export class SinglePastCommitInfo extends React.Component<
...
+      deletionCount: "",
+      list: [],
+    
     };
+    this.showPastCommitWork();
   }
 
+  showPastCommitWork = async () => {
+    let gitApi = new Git();
+    let detailedLogData = await gitApi.detailedLog(this.props.data.commit, this.props.topRepoPath);
```
jaipreet-s: How do you feel about handling the error callback from this API? I'd like to not fail silently, at-least.

jaipreet-s:
@weihwang If you want to have a look the implemented UX

src/components/PastCommitNode.tsx
```diff
@@ -1,114 +1,139 @@
...
+        onClick={() => {!this.state.expanded && this.expand()}}
+        className={this.getNodeClass()}
+      >
+        <div className={pastCommitHeaderStyle}>
+          <div className={pastCommitHeaderItemStyle}>
+            {this.props.pastCommit.author}
+          </div>
+          <div className={pastCommitHeaderItemStyle}>
+            {this.props.pastCommit.commit.slice(0, 9)}
```
neelamgehlot: Is there a reason for showing SHA1 with 9 digits instead of 7 digits which seems to be the default length for short SHA1?

src/components/SinglePastCommitInfo.tsx
```diff
@@ -38,23 +34,21 @@ import { classes } from 'typestyle/';
...
 }
 
 export interface ISinglePastCommitInfoState {
   displayDelete: boolean;
   displayReset: boolean;
+  info: string;
+  filesChanged: string;
+  insertionCount: string;
+  deletionCount: string;
+  list: Array<CommitModifiedFile>;
```
neelamgehlot: Could you give this more descriptive name? Maybe ```modifiedFiles``` ?

src/components/PastCommitNode.tsx
```diff
@@ -1,114 +1,139 @@
...
+    const idAbrev = this.props.pastCommit.commit.slice(0, 7);
+    const branches = [];
+    for (let i = 0; i < this.props.data.length; i++) {
+      const branch = this.props.data[i];
+      // tag sent from describe command. Must unparse to find commit hash
+      // https://git-scm.com/docs/git-describe#git-describe-ltcommit-ishgt82308203
+      if (!branch.tag) {
+        continue;
+      }
+      const tagParts = branch.tag.split("-");
```
neelamgehlot: Do you think it would be more appropriate to return the top commit SHA1 for the branch in the branch API payload instead of determining it from the tag?

In which case the top_commit_sha would always be defined for a given branch.

```
"branches": [
    {
        "is_current_branch":false,
        "is_remote_branch":true,
        "name":"branch-name", 
        "upstream":"upstream-branch-name",
        "tag":"branch-tag",
        "top_commit_sha":"abcd123",
    }
]
```

src/components/PastCommitNode.tsx
saulshanabrook: I believe there were  9 digits in the mockup. But yeah, makes sense to use 7 instead of 9, since that does seem to be the default.

src/components/PastCommitNode.tsx
saulshanabrook: Yeah it would be more appropriate. I tried not to modify the backend API in this PR, to minimize breaking changes.

Should I add that field?

neelamgehlot: This PR is already very large, so If you don't mind could you send another PR as follow up for this comment?

saulshanabrook: Will do, opened https://github.com/jupyterlab/jupyterlab-git/issues/274 to track this

PR #275, opened 2018-11-26
 Fix git repo determination when there are no tags

jaipreet-s:
When there are no tags, `git describe` returns `fatal: No names found, cannot describe anything`. This fails the entire `/git/all_history` API call and we fail to identify the Git repo completely.

This commit doesn't fail in this case but returns an empty list of tags and also uses the `--tag` option which returns non-annotated tags as well.

---
### **Testing**

Unit Tested: pytest test/
Tested via local JupyterLab installation and with a fresh GitLab repo

---


jaipreet-s:
@saulshanabrook Are you available to review this today?

PR #280, opened 2018-11-27
Send full commit SHA1 for each branch

### LINKED ISSUE #274: Send full commit SHA1 for each branch
As [suggested by @neelamgehlot](https://github.com/jupyterlab/jupyterlab-git/pull/270#discussion_r235574371), if we sent the SHA1 for each branch, then we wouldn't need to do parse the branch description as done in #270

neelamgehlot:
Fixes #274

### DIFF
--- a/tests/test-components/PastCommitNode.spec.tsx
+++ b/tests/test-components/PastCommitNode.spec.tsx
@@ -31,7 +31,7 @@ describe("PastCommitNode", () => {
       is_remote_branch: false,
       name: "name3",
       upstream: "upstream",
-      top_commit: "2414721hijklmnopqrstuvwxyz01234567890123",
+      top_commit: "2414721b194453f058079d897d13c4e377f92dc6",
       tag: "v1.0.4-14-g2414721"
     },
     {
@@ -39,7 +39,7 @@ describe("PastCommitNode", () => {
       is_remote_branch: true,
       name: "name4",
       upstream: "upstream",
-      top_commit: "2414721hijklmnopqrstuvwxyz01234567890123",
+      top_commit: "2414721b194453f058079d897d13c4e377f92dc6",
       tag: "v1.0.5-0-g2414721"
     }
   ];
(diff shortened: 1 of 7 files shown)


tests/test-components/PastCommitNode.spec.tsx
```diff
@@ -29,13 +31,15 @@ describe("PastCommitNode", () => {
       is_remote_branch: false,
       name: "name3",
       upstream: "upstream",
+      top_commit: "2414721hijklmnopqrstuvwxyz01234567890123",
```
saulshanabrook: This should be changed to `2414721b194453f058079d897d13c4e377f92dc6` so it matches.

tests/test-components/PastCommitNode.spec.tsx
```diff
@@ -29,13 +31,15 @@ describe("PastCommitNode", () => {
...
       tag: "v1.0.4-14-g2414721"
     },
     {
       is_current_branch: false,
       is_remote_branch: true,
       name: "name4",
       upstream: "upstream",
+      top_commit: "2414721hijklmnopqrstuvwxyz01234567890123",
```
saulshanabrook: Same here `2414721b194453f058079d897d13c4e377f92dc6`

tests/test-components/PastCommitNode.spec.tsx
neelamgehlot: Done

tests/test-components/PastCommitNode.spec.tsx
neelamgehlot: Done

tests/test-components/PastCommitNode.spec.tsx
```diff
@@ -29,13 +31,15 @@ describe("PastCommitNode", () => {
...
       upstream: "upstream",
+      top_commit: "2414721b194453f058079d897d13c4e377f92dc6",
       tag: "v1.0.4-14-g2414721"
     },
     {
       is_current_branch: false,
       is_remote_branch: true,
       name: "name4",
       upstream: "upstream",
+      top_commit: "2414721b194453f058079d897d13c4e377f92dc6",
```
jaipreet-s: WDYT about adding some assertions on this new `prop` so that its intended side-effect is clear?

neelamgehlot: Talked offline with @jaipreet-s , all the test cases are already using the new prop, so there is no need to add any more assertions.

PR #295, opened 2018-12-28
Fix installation instructions

orome:
Make it clear that all three steps are required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,11 +18,10 @@ A JupyterLab extension for version control using git
 
 ## Install
 
-```bash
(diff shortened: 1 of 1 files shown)


orome:
@jaipreet-s So it's correct then that all three steps are required? If so then what steps are needed to keep `jupyterlab-git` up to date? Is running `pip install -U jupyterlab-git` sufficient? (#296)

jaipreet-s:
@orome  Yes, I would  use all 3 steps - Since this is a frontend as well as a server extension, this requires keeping the sources in sync across NPM and pip.

PR #326, opened 2019-04-01
Fix dark theme appearance

### LINKED ISSUE #307: Make sidebar colors better for JupyterLab Dark
I am using the built in JupyterLab Dark theme and the git sidebar doesn't match well with the colors. The History box is barely visible. Could there be a setting to adjust the colors easily or it auto-detect if the theme is JupyterLab Dar or JupyterLab Light?

Example of view:

image

saulshanabrook:
This closes #307 by improving the appearance of in the dark theme. 

It also [switches from detecting the with JS to CSS](https://github.com/jupyterlab/jupyterlab-git/commit/0c340ab58376f7688430cf397c2063d45dc1a771).

# Main Page
### Before
Screen Shot 2019-04-01 at 4 45 03 PM

### After
Screen Shot 2019-04-01 at 5 46 27 PM

### Before
Screen Shot 2019-04-01 at 4 45 17 PM

### After
Screen Shot 2019-04-01 at 5 46 11 PM

# History Page
### Before
Screen Shot 2019-04-01 at 4 45 07 PM

### After
Screen Shot 2019-04-01 at 5 46 23 PM

### Before
Screen Shot 2019-04-01 at 4 45 12 PM

### After
Screen Shot 2019-04-01 at 5 46 17 PM

### DIFF
--- a/src/components/BranchHeader.tsx
+++ b/src/components/BranchHeader.tsx
@@ -48,7 +48,6 @@ export interface IBranchHeaderProps {
   disabled: boolean;
   toggleSidebar: Function;
   showList: boolean;
-  currentTheme: string;
   sideBarExpanded: boolean;
 }
 
@@ -205,10 +204,10 @@ export class BranchHeader extends React.Component<
               className={
                 this.props.disabled
                   ? classes(
-                      branchDropdownButtonStyle(this.props.currentTheme),
+                      branchDropdownButtonStyle,
                       headerButtonDisabledStyle
                     )
-                  : branchDropdownButtonStyle(this.props.currentTheme)
+                  : branchDropdownButtonStyle
               }
               title={'Change the current branch'}
               onClick={() => this.toggleSelect()}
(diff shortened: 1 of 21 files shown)


jaipreet-s:
Really awesome that we don't have to pipe the `isLightTheme` variable through all the components.

jaipreet-s:
BTW the `After` screenshots are missing the remote branch information. Is that just a side-effect of your workspace not having a remote set?

saulshanabrook:
> BTW the After screenshots are missing the remote branch information. Is that just a side-effect of your workspace not having a remote set?

Yep! I just checked it and it works when the remote is set:

Screen Shot 2019-04-03 at 3 48 35 PM

jaipreet-s:
I see the `Git` icon on the left sidebar is also not white in Dark theme. If we have an image available, can you fix that as well?

saulshanabrook:
> I see the Git icon on the left sidebar is also not white in Dark theme. If we have an image available, can you fix that as well?

I don't think we have one available.

jaipreet-s:
> I don't think we have one available.

:-\ Ok. This is great as is then!

PR #328, opened 2019-04-03
Add full file path to tooltip over file

### LINKED ISSUE #327: Changed/Staged file path
Currently I only see the name of the changed or staged file. How can I see the full path of the file?

This is an issue when I have two modified files in different folders but with the same name. 

One quick fix idea is to show the full path in a tool tip when the user mouses over the filename.

Displaying the file path was mentioned in an issue with a larger scope ( #227 ), but I did not see a plan for implementing it.

saulshanabrook:
This fixes #327 by showing a the full file path on hover over a file name:

Screen Shot 2019-04-03 at 3 54 52 PM

### DIFF
--- a/src/components/FileItem.tsx
+++ b/src/components/FileItem.tsx
@@ -277,6 +277,7 @@ export class FileItem extends React.Component<IFileItemProps, {}> {
               this.props.app
             )
           }
+          title={this.props.file.to}
         >
           {this.props.extractFilename(this.props.file.to)}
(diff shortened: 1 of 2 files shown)


jaipreet-s:
Quick change!

I do think we should start a unit test/spec for this component and validate at-least the tooltip

saulshanabrook:
@jaipreet-s Sounds good. I just added a test for the tooltip.

PR #332, opened 2019-04-05
Fix some tslint errors

jaipreet-s:
Not all errors are "auto-fixable" but taking some incremental steps towards fixing them in gitClone.ts and index.ts

Ideally, we should be in a state where tslint erors should fail the buil.d


jaipreet-s:
@saulshanabrook : Do you have bandwidth to take a look at this. If not, no worries, I can assign someone else.

PR #336, opened 2019-04-10
Stricter lint

saulshanabrook:
* Make tslint rules compatible with prettier
* run linting in CI
* run linting on commit on touched files
* integrate tslint into typescript compilation pipeline so that errors appear in vscode (after changing typescript interpreter to use workspace one)
* fix all other linting errors manually

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,8 @@
     "clean": "rimraf lib",
     "prepublish": "npm run build",
     "watch": "tsc -w",
-    "test": "jest"
+    "test": "jest",
+    "lint": "tslint --project .",
   },
   "files": [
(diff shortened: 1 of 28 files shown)


package.json
```diff
@@ -15,7 +15,9 @@
     "clean": "rimraf lib",
     "prepublish": "npm run build",
     "watch": "tsc -w",
-    "test": "jest"
+    "test": "jest",
+    "lint": "tslint --project .",
+    "tslint-check": "tslint-config-prettier-check ./tslint.json"
```
jaipreet-s: I don't see this target used anywhere?

saulshanabrook: I used it locally. It tells you what tslint config options conflict with prettier.

PR #342, opened 2019-04-17
Upgrade to jupyterlab 1.0.0a3

gnestor:
Upgrade jupyterlab dependencies to work with 1.0.0a3


gnestor:
This build is failing because it's trying to `jupyter labextension install .` with jupyterlab stable vs. pre. I don't think we want to change the install command to ` pip install pytest jupyterlab==1.0.0a3 psutil` because we will probably forget to change it back when 1.0.0 lands.

saulshanabrook:
@gnestor What about waiting till 1.0 is out before upgrading the dependencies here?

Do you need this merged before then?

jaipreet-s:
I have the same question as Saul.  This'll break consumers using the last non-pre-release version of JupyterLab,  who perhaps only want to use the released versions.

(Also thanks for PR!)

gnestor:
> This'll break consumers using the last non-pre-release version of JupyterLab, who perhaps only want to use the released versions.

We would release with a pre tag, like the @jupyterlab/github extension: https://www.npmjs.com/package/@jupyterlab/github

> What about waiting till 1.0 is out before upgrading the dependencies here?

It's just currently not working with the 1.0 alpha, so if we merged this and did a pre-release, both alpha and stable users will be able to use it.

The only issue is that CI will fail unless we change something in the travis file... 🤔

jaipreet-s:
> The only issue is that CI will fail unless we change something in the travis file... 🤔

Our setup.py has a dependency on `notebook`, and not `jupyterlab`, which is all we need to get access to the relevant base classes in the server extension.

So, I don't think we need to install jupyterlab as a Python dependency in travis. Looking at the file [history](https://github.com/jupyterlab/jupyterlab-git/commits/master/.travis.yml) , we only had the notebook dependency at one point, so it should work just with it.

gnestor:
Ok cool! Should I just delete the `install` entry in the travis file?

jaipreet-s:
> Ok cool! Should I just delete the `install` entry in the travis file?

Replacing `jupyterlab` with `notebook` should work 🤞

gnestor:
Ok I just updated...

I have a feeling that travis is still going to fail on the `jupyter labextension install .` step 🤔

set92:
Since they are going to release jupyterlab 1.0 in 2 days (https://github.com/jupyterlab/jupyterlab/issues/6504) it would be good to have this merged with the master.

gnestor:
@set92 I agree. There might be some new breaking API changes that affect jupyterlab-git so I need to take a second look and fix this merge conflict before we merge and publish a new version.

jaipreet-s:
This was merged as part of #374  and released as part of version 0.7. Thanks all!

PR #344, opened 2019-04-29
Fix/push pull windows

### LINKED ISSUE #291: push and pull fail, GIT_TERMINAL_PROMPT=0 command not found
A fresh install of the jupyter-lab git extension, stage and commit on an existing git repository works fine. However, pressing the button `push` fails with a message just saying `Push failed`. In the terminal protocol, it says 
```
[E 13:29:11.800 LabApp] 500 POST /git/push?1544963351615 (::1) 124.80ms referer=http://localhost:8916/lab
```
Pressing button `pull` fails with a message `The command ""GIT_TERMINAL_PROMPT=0 git pull --no-commit"" is either misspelled or could not be found.`. No terminal message protocolled. 

My Git by default uses saved credentials, so there is no need to prompt for username/password.
I never had problems using git on my system, both with the terminal commands or with third-party software, so it must be an issue related to jupyter lab. The command `GIT_TERMINAL_PROMPT=0 git pull --no-commit` works fine when I execute it in my shell (cygwin's mintty). It does not work (CommandNotFound) in PowerShell, however, which opens when I go to Git>Open terminal in the jupyter lab menu. But I never use PowerShell for anything, so I see no reason why PowerShell is used here.

Does jupyter lab/git use the PowerShell? I this the reason for the commands to fail (it looks to me that PowerShell cannot handle prefixes like `GIT_TERMINAL_PROMPT=0` before the actual command... Can I tell jupyter lab/git not to use PowerShell?

I am on Windows 7, Python 3.6.0 and Git 2.11.1 installations (not under cygwin), and jupyter lab 0.35.4.

ameier38:
- fixes #291 by passing in env dictionary
- separates commands when passing to Popen

### DIFF
--- a/tests/unit/test_pushpull.py
+++ b/tests/unit/test_pushpull.py
@@ -1,3 +1,4 @@
+import os
 from subprocess import PIPE
 
 from mock import patch, call, Mock
@@ -9,6 +10,8 @@ from jupyterlab_git.git import Git
 def test_git_pull_fail(mock_subproc_popen):
     # Given
     process_mock = Mock()
+    env = os.environ.copy()
+    env['GIT_TERMINAL_PROMPT'] = '0'
     attrs = {
         'communicate.return_value': ('output', 'Authentication failed'.encode('utf-8')),
         'returncode': 1
@@ -22,11 +25,12 @@ def test_git_pull_fail(mock_subproc_popen):
     # Then
     mock_subproc_popen.assert_has_calls([
         call(
-            ['GIT_TERMINAL_PROMPT=0 git pull --no-commit'],
+            ['git pull --no-commit'],
             stdout=PIPE,
             stderr=PIPE,
             cwd='/bin/test_curr_path',
-            shell=True
+            shell=True,
+            env=env,
         ),
         call().communicate()
     ])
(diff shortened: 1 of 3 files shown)


ameier38:
@jaipreet-s let me know if you have any comments. Would like to get this merged as I cannot currently use on Windows.

jaipreet-s:
Hi @ameier38 - Thanks for picking this up! I'll get back on this ASAP.

tests/unit/test_pushpull.py
```diff
@@ -93,6 +103,8 @@ def test_git_push_fail(mock_subproc_popen):
 def test_git_push_success(mock_subproc_popen):
     # Given
     process_mock = Mock()
+    env = os.environ.copy()
```
jaipreet-s: Only commend I have is that if we could enhance these tests:

We could mock out `os.environ` to return a dummy value and then assert that subprocess was called with the dummy value + `GIT_TERMINAL_PROMPT` as the env parameter

ameier38: good catch, updated.

jaipreet-s:
Thanks!

Issue #346, opened 2019-05-01
Open Terminal option in Git menu doesn't resolve Jupyter home dir


jaipreet-s:
The Open Terminal option does a "cd $REPONAME" which opens the terminal on the `home_dir`/$REPONAME

This doesn't factor in the base directory of the Jupyter server.

mathematicalmichael:
I don't even see this option because the git menu refuses to line up with the navigation. 
I use the docker-stacks, so `/home/jovyan` is the default server entry point. Is that the problem?

PR #347, opened 2019-05-02
Use the full path when opening the terminal from the Git menu

jaipreet-s:
**Description**

Currently, the menu item opens a terminal based on the path of the current working directory relative to the server root. In some cases, the terminal starts from a directory other than the server root so this "cd $PATH" fails.

To avoid this, just construct the full path and "cd $SERVER_ROOT/$PATH" while executing the command

Other changes
* Moved the addMenuItems code into the gitMenuCommands file
* Removed unused command.

Addresses #346

### DIFF
--- a/src/gitMenuCommands.ts
+++ b/src/gitMenuCommands.ts
@@ -1,13 +1,9 @@
-import { Dialog, showDialog } from '@jupyterlab/apputils';
-
 import { JupyterLab } from '@jupyterlab/application';
-
+import { Dialog, InstanceTracker, showDialog } from '@jupyterlab/apputils';
+import { IMainMenu } from '@jupyterlab/mainmenu';
 import { ServiceManager } from '@jupyterlab/services';
-
-import { InstanceTracker } from '@jupyterlab/apputils';
-
 import { Terminal } from '@jupyterlab/terminal';
-
+import { Menu } from '@phosphor/widgets';
 import { Git } from './git';
 
 /**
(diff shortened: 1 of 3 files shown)


src/gitMenuCommands.ts
```diff
@@ -68,7 +89,11 @@ export function addCommands(app: JupyterLab, services: ServiceManager) {
           terminal.session.send({
             type: 'stdin',
             content: [
-              'cd "' + currentFileBrowserPath.split('"').join('\\"') + '"\n'
+              'cd "' +
+                app.info.directories.serverRoot +
```
jaipreet-s: I'm not 100% sure if this is the best way to get the Jupyter server root. Open to suggestions.

jasongrout: That serverRoot variable is for UI information only, and should not be relied on in code (the server may choose to make it an empty string, for example, or the root "file system" may be provided by a database, not corresponding to a directory on the file system). See https://github.com/jupyterlab/jupyterlab/blob/cf1063590f3f438b771b679fafc2eeed0838421e/packages/application/src/frontend.ts#L278-L303 for more complete documentation.

In general, working with directories on the server outside of the sandbox of the server root should be done by a server extension.

saulshanabrook: Thanks @jasongrout for that clarification. @jaipreet-s So we could make an endpoint in the server extension that returns the absolute path to the git directory.

jaipreet-s: Thanks. This is on my TODO list to get back to.

jaipreet-s: > Thanks @jasongrout for that clarification. @jaipreet-s So we could make an endpoint in the server extension that returns the absolute path to the git directory.

@saulshanabrook In #395, I've added a server endpoint to return the server root which can be used for this as well.

fcollonval:
Superseeded by #438

Issue #349, opened 2019-05-14
Multiple git commands are invoked upon pathChanged


urihoenig:
Every path change invokes multiple git commands just in order to check if the path is a git repo or not.
This involves unnecessary "git status" which goes over the whole sub tree and in some configurations might have a major performance impact.

telamonian:
fixed in #350

PR #350, opened 2019-05-14
Minimize git access upon changed path

urihoenig:
Upon path change, disableIfInGitDirectory unnecessarily invokes all_history which ends up running 4 git commands instead of 1, just to check if the new path is a git repo.

This results in performance impact in some configurations (as "git status" goes over the whole file sub tree).

all_history verifies first show_top_level's result and only then proceeds to the other commands, so it should be enough in this case.

Addresses #349

### DIFF
--- a/src/gitClone.ts
+++ b/src/gitClone.ts
@@ -55,7 +55,7 @@ export class GitClone extends Widget {
    */
   disableIfInGitDirectory(): void {
     this.gitApi
-      .allHistory(this.fileBrowser.model.path)
+      .showTopLevel(this.fileBrowser.model.path)
       .then(response => {
         if (response.code === 0) {
           this.enabledCloneButton.parent = null;


setup.py
```diff
@@ -8,7 +8,7 @@
 
 setuptools.setup(
     name='jupyterlab_git',
-    version='0.5.0',
```
jaipreet-s: Please fix the versioning. I'd stay on 0.5 till we release 0.6

urihoenig: I'm sorry, I was sure I removed this. Will do now.

urihoenig:
> Thanks for the contribution! Could you provide some details on how this has been tested?

So far I only managed to modify it locally on an already installed jupyterlab with the git extension (in the vendors~main).

I'm still struggling with installing it in jupyter after building it locally (I changed the version locally and npm pack fails). Is it possible? (jupyter labextension install without npmjs)

jaipreet-s:
> > Thanks for the contribution! Could you provide some details on how this has been tested?
> 
> So far I only managed to modify it locally on an already installed jupyterlab with the git extension (in the vendors~main).
> 
> I'm still struggling with installing it in jupyter after building it locally (I changed the version locally and npm pack fails). Is it possible? (jupyter labextension install without npmjs)

You can follow the instructions [here](https://github.com/jupyterlab/jupyterlab-git#install-1) to install a local development version of jupyterlab-git in your local JupyterLab setup

PR #366, opened 2019-06-20
Identity feature

### LINKED ISSUE #351: UI to configure name and email settings for local user?
When committing to Git, it is important to have proper attribution. Git has two settings that must be changed before making an initial commit:

- `name`
- `email`

Please consider adding a UI that helps a user configure their `name` and `email` settings for Git, with particular consideration given to how JupyterLab may be used in multi-user environments.

## Quick references

- https://help.github.com/en/articles/setting-your-username-in-git
- https://help.github.com/en/articles/setting-your-commit-email-address-in-git

fcollonval:
Fixes #351 and will imply simplification in #345.

## User changes

If the user has no `user.name` and no `user.email` defined in Git global configuration, the GitPanel will be replaced by a warning and a button prompting the user for a name and an email.

image

If the user provides them, they will be added as global configuration options in git. Then the extension will behave as previously.

Note;

* The email is not verified to be a valid email string
* If the user removes his identity after opening JupyterLab (very unlikely), the extension is never checking again if an identity is set.

## Backward incompatibilities

None

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -18,6 +18,44 @@ class Git:
         super(Git, self).__init__(*args, **kwargs)
         self.root_dir = os.path.realpath(os.path.expanduser(root_dir))
 
+    def config(self, **kwargs):
+        """Get or set Git options.
+        
+        If no kwargs, all options are returned. Otherwise kwargs are set.
+        """
+        response = {"code": 1}
+
+        if len(kwargs):
+            output = []
+            for k, v in kwargs.items():
+                cmd = ["git", "config", "--global", "--add", k, v]
+                p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+                out, err = p.communicate()
+                output.append(out.decode("utf-8").strip())
+                response["code"] = p.returncode
+                if p.returncode != 0:
+                    response["command"] = " ".join(cmd)
+                    response["message"] = err.decode("utf-8").strip()
+                    return response
+
+            response["message"] = "\n".join(output).strip()
+        else:
+            cmd = ["git", "config", "--global", "--list"]
+            p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            output, error = p.communicate()
+
+            response = {"code": p.returncode}
+
+            if p.returncode != 0:
+                response["command"] = " ".join(cmd)
+                response["message"] = error.decode("utf-8").strip()
+            else:
+                raw = output.decode("utf-8").strip()
+                response["options"] = dict(
+                    [l.split("=", maxsplit=1) for l in raw.split("\n")]
+                )
+
+        return response
 
     def changed_files(self, base=None, remote=None, single_commit=None):
         """Gets the list of changed files between two Git refs, or the files changed in a single commit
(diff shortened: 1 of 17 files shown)


jaipreet-s:
Thanks for this. Based on a quick look,  this will disable all of the Git panel functionality and replace it with this error message if theres no identity set. Now, the username/email is not required till the user starts to commit the information, so disabling the "changed/staged" view, and history view, and the upcoming diff functionality seems heavy for not having a username/email set.

My suggestion would be would be to delay this check till the user goes to commit the the information

fcollonval:
Thanks for the feedback @jaipreet-s.

Indeed the current approach is disabling the main panel. I'll update it following your suggestion to check once and only once at the first commit attempt. 
Does that sounds better?

jaipreet-s:
That sounds great! I'll also engage @weihwang  to have a look at the username/email popup wording/UX.

jupyterlab_git/git.py
```diff
@@ -16,6 +16,45 @@ def __init__(self, root_dir, *args, **kwargs):
...
+                response["code"] = p.returncode
+                if p.returncode != 0:
+                    response["command"] = " ".join(cmd)
+                    response["message"] = err.decode("utf-8").strip()
+                    return response
+
+            response["message"] = "\n".join(output).strip()
+        else:
+            cmd = ["git", "config", "--global", "--list"]
+            p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
jaipreet-s: We don't need to check for the global configuration, right. As long as user name/email is configured at the repo or global leve, we're good. `git config user.email/name` will return the repo level user name/email or the global one if no override is set.

jupyterlab_git/git.py
```diff
@@ -16,6 +16,45 @@ def __init__(self, root_dir, *args, **kwargs):
...
+        """Get or set Git options.
+        
+        If no kwargs, all options are returned. Otherwise kwargs are set.
+        """
+        response = {"code": 1}
+
+        if len(kwargs):
+            output = []
+            for k, v in kwargs.items():
+                cmd = ["git", "config", "--global", "--add", k, v]
```
jaipreet-s: In my opinion, setting these are the global level and potentially overwriting global configuration is overkill. As long as we set it at the repo level, it should suddice.

fcollonval:
@jaipreet-s does it look better?

jupyterlab_git/git.py
fcollonval: I set options at repository level

jupyterlab_git/git.py
fcollonval: You are right, I'll do that.

mathematicalmichael:
This is a great idea! 
Am I right in interpreting that it first checks for local user info, then global, if missing? 
Then sets local if so.

Do these changes to .git/config migrate upon push?

src/widgets/gitPushPull.ts
```diff
@@ -1,6 +1,6 @@
 import { Spinner } from '@jupyterlab/apputils';
 import { Widget } from '@phosphor/widgets';
-import { Git, IGitPushPullResult } from './git';
+import { Git, IGitPushPullResult } from '../git';
```
jaipreet-s: I like this new directory structure. Thanks for moving.

jaipreet-s:
@fcollonval This looks great. 

Looks like there is one merge conflict and an unrelated Travis CI failure though

fcollonval:
@jaipreet-s I finally overcome the CI error. Could you please have a look?

PR #374, opened 2019-07-02
Port to JupyterLab v1

fcollonval:
This makes the application compatible with JupyterLab v1.

I have some troubles with:

- the icons (the discard one in particular). It is not scaled to fit the button. Does anyone get an idea to fix this?
- the jest tests - I don't understand the error

```bash
 FAIL  tests/test-components/FileItem.spec.tsx
  ● Test suite failed to run

    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    C:\Users\d567619\CoSApp\jupyterlab-git\node_modules\@jupyterlab\apputils\lib\index.js:3
    export * from './clientsession';
    ^^^^^^

    SyntaxError: Unexpected token export
```


jaipreet-s:
Looks like this is getting the same Jest errors as #342

fcollonval:
@jaipreet-s and @gnestor sorry I was in the mood to port some extension to 1.0 and forget to check the existence of a related PR...

Note in #342 the renaming of conditional style is missing: `data-theme-light` => `data-jp-theme-light`

@blink1073 do you have an idea about the jest error. Could a fix be inspired by the following commit: https://github.com/jupyterlab/jupyterlab/commit/c4bcc0d81b104fe1e07d7f04e574d42649f8f7c4

krassowski:
I have the same jest issue in https://github.com/krassowski/jupyterlab-go-to-definition/issues/8, tried a couple of things that were changed in upstream in testutilis (including jupyterlab/jupyterlab@c4bcc0d) but could not figure way around. @blink1073 help would be greatly appreciated!

blink1073:
Yes, basically you need to tell Babel to preprocess any JS files: https://kulshekhar.github.io/ts-jest/user/config/

krassowski:
Thank you!

@fcollonval I initially followed jupyterlab/jupyterlab@c4bcc0d (though with difficulties) and then finally succeeded (https://github.com/krassowski/jupyterlab-go-to-definition/commit/4bbcaa6664581ac996756170201a6283c11bb9cf) after reading through https://github.com/facebook/jest/issues/6229 - seems that there are many known issues with babel/jest-ts configuration. Hope it might save you a few minutes!

fcollonval:
Thanks a lot to @krassowski and @blink1073 

I fixed the jest configuration (note to @krassowski I figure out what to use from @jupyterlab/testutils to not introduce it as dependency if you are interested).

There is one styling issue remaining; some icons are bigger than their container:
image

fcollonval:
The icons are fixed:
image

@jaipreet-s this is good for me.

jaipreet-s:
Awesome @fcollonval, thanks for fixing the tests.

The icons in Firefox look a bit misaligned now, is that a side effect of this change?

Screen Shot 2019-07-05 at 11 12 14 AM

fcollonval:
Hey @jaipreet-s, I had a quick look. Its seems that Chrome is adding style to the button like `align-items: flex-start` automagically (style comes from user agent stylesheet). But this is not done in Firefox. I try playing with the align-items in Firefox but the vertical alignment didn't change. 
This should come from elsewhere as in this PR I only changed the `background-size` attribute on the problematic buttons.
I propose to merge this to be able to make a release for JupyterLab v1.0. And we open another issue regarding the styling with Firefox. What do you think?

Issue #268, opened 2018-11-17
Diff any two nodes in history view


dhirschfeld:
Moved over from #245.

As it says in the title, it would be very convenient to be able to see the diff between any two points selected in the history view.

jaipreet-s:
Hi @dhirschfeld , I'm consolidating discussions around visual diff into #317. Thanks

fcollonval:
Now that #900 is in this should be easy to implement by defining a context menu in the history tab to select to node and call the `gitShowDiff` commands with the proper `DiffModel`

fcollonval:
Fixed by https://github.com/jupyterlab/jupyterlab-git/pull/1108 and https://github.com/jupyterlab/jupyterlab-git/pull/1117

Issue #252, opened 2018-10-17
View diff on filename click


weihwang:
Proposal to connect filename double click in a commit detail view or working changes list view to the file diff view. V1 will open a terminal, with an improvement to a graphic interface view.

![image](https://user-images.githubusercontent.com/551333/47106674-14755e00-d1fc-11e8-825d-9ab2c14e27df.png)

![image](https://user-images.githubusercontent.com/551333/47106649-04f61500-d1fc-11e8-8675-fae586b80616.png)

dhirschfeld:
+:100:!

I like how in VSCode you only need to select a file for it to show the diff. This however probably works best with VSCode's concept of a temporary view of the file - i.e. as soon as you select a different file the old view is *replaced*  with the view of the new file. This works since you wouldn't want simply selecting files to leave dozens of tabs open. To get a permanent tab you have to actually double-click the file.

weihwang:
@dhirschfeld Thinking about the single click, it seem pretty heavy to render a diff view. Also this requires a change to the behavior of Jupyterlab to support single-persistent window that changes on each single-click. In that use-case, if supported, would a diff view usually be expected to be the default view as opposed to just the current state of the file as an easy browse interaction?

I think it would break down to separate tasks:
1. Should we support single-click persistent window reload for file "preview/browse"
2. If supported, should this experience show "diff" view vs. "current" view?

dhirschfeld:
The single click open is what is known as [Preview Mode](https://code.visualstudio.com/docs/getstarted/userinterface#_preview-mode) - you can click-browse through files without cluttering your workspace with multiple tabs. When you find the file you're interested in you can double-click it to make it *sticky* - i.e. to persist when you click/open a new file

fcollonval:
Although we don't have a _Preview Mode_, the feature is in now. So I close this issue.

PR #384, opened 2019-07-17
Visual diff for Notebook files

jaipreet-s:
This change introduces the feature to be able to diff ipynb from various points inside the Git experience. The currently supported points are the Changed, Staged areas, and the History panel.

The `<Diff />` component maintains a registry of DiffProvider components and delegates to each provider based on the file extension.

Note
* This change is off the pre 1.0 version of the Git extension because NBDime isn't upgraded to 1.0 

Addresses #252 #268 

TODOs

* [x] Port to 1.0 after https://github.com/jupyter/nbdime/pull/483 and use `ReactWidget` interface

### DIFF
--- a/src/components/FileList.tsx
+++ b/src/components/FileList.tsx
@@ -115,14 +115,12 @@ export class FileList extends React.Component<IFileListProps, IFileListState> {
       label: 'Open',
       caption: 'Open selected file',
       execute: async () => {
-        try {
-          await this.openListedFile(
-            this.state.contextMenuTypeX,
-            this.state.contextMenuTypeY,
-            this.state.contextMenuFile,
-            this.props.app
-          );
-        } catch (err) {}
+        await this.openListedFile(
+          this.state.contextMenuTypeX,
+          this.state.contextMenuTypeY,
+          this.state.contextMenuFile,
+          this.props.app
+        );
       }
     });
 
--- a/src/components/FileList.tsx
+++ b/src/components/FileList.tsx
@@ -110,84 +110,90 @@ export class FileList extends React.Component<IFileListProps, IFileListState> {
       disableFiles: false
     };
 
-    commands.addCommand(CommandIDs.gitFileOpen, {
-      label: 'Open',
-      caption: 'Open selected file',
-      execute: async () => {
-        await this.openListedFile(
-          this.state.contextMenuTypeX,
-          this.state.contextMenuTypeY,
-          this.state.contextMenuFile,
-          this.props.app
-        );
-      }
-    });
+    if (!commands.hasCommand(CommandIDs.gitFileOpen)) {
+      commands.addCommand(CommandIDs.gitFileOpen, {
+        label: 'Open',
+        caption: 'Open selected file',
+        execute: async () => {
+          await this.openListedFile(
+            this.state.contextMenuTypeX,
+            this.state.contextMenuTypeY,
+            this.state.contextMenuFile,
+            this.props.app
+          );
+        }
+      });
+    }
 
-    commands.addCommand(CommandIDs.gitFileDiffWorking, {
-      label: 'Diff',
-      caption: 'Diff selected file',
-      execute: () => {
-        openDiffView(
-          this.state.contextMenuFile,
-          this.props.app,
-          {
-            currentRef: { specialRef: 'WORKING' },
-            previousRef: { gitRef: 'HEAD' }
-          },
-          this.props.renderMime
-        );
-      }
-    });
+    if (!commands.hasCommand(CommandIDs.gitFileDiffWorking)) {
+      commands.addCommand(CommandIDs.gitFileDiffWorking, {
+        label: 'Diff',
+        caption: 'Diff selected file',
+        execute: () => {
+          openDiffView(
+            this.state.contextMenuFile,
+            this.props.app,
+            {
+              currentRef: { specialRef: 'WORKING' },
+              previousRef: { gitRef: 'HEAD' }
+            },
+            this.props.renderMime
+          );
+        }
+      });
+    }
 
-    commands.addCommand(CommandIDs.gitFileDiffIndex, {
-      label: 'Diff',
-      caption: 'Diff selected file',
-      execute: () => {
-        openDiffView(
-          this.state.contextMenuFile,
-          this.props.app,
-          {
-            currentRef: { specialRef: 'INDEX' },
-            previousRef: { gitRef: 'HEAD' }
-          },
-          this.props.renderMime
-        );
-      }
-    });
+    if (!commands.hasCommand(CommandIDs.gitFileDiffIndex)) {
+      commands.addCommand(CommandIDs.gitFileDiffIndex, {
+        label: 'Diff',
+        caption: 'Diff selected file',
+        execute: () => {
+          openDiffView(
+            this.state.contextMenuFile,
+            this.props.app,
+            {
+              currentRef: { specialRef: 'INDEX' },
+              previousRef: { gitRef: 'HEAD' }
+            },
+            this.props.renderMime
+          );
+        }
+      });
+    }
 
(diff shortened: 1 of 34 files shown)


src/components/PastCommits.tsx
```diff
@@ -40,6 +42,7 @@ export class PastCommits extends React.Component<IPastCommitsProps, {}> {
           refresh={this.props.refresh}
           sideBarExpanded={this.props.sideBarExpanded}
           display={this.props.showList}
+          renderMime={this.props.renderMime}
```
dleen: Yeah, now that I see how many places this is passed/used, definitely use context: https://reactjs.org/docs/context.html#when-to-use-context

src/components/diff/Diff.tsx
```diff
@@ -0,0 +1,54 @@
...
+  renderMime: IRenderMimeRegistry;
+  path: string;
+  diffContext: IDiffContext;
+}
+
+/**
+ * The parent diff component which maintains a registry of various diff providers.
+ * Based on the extension of the file, it delegates to the relevant diff provider.
+ */
+export class Diff extends React.Component<IDiffProps, {}> {
```
dleen: This class is a bit strange since it doesn't do much else except rendering another component. I can't help but think it's unnecessary, but an alternative solution isn't immediately obvious. It could be stateless: https://itnext.io/react-component-class-vs-stateless-component-e3797c7d23ab. Does this help in this case? https://reactjs.org/docs/composition-vs-inheritance.html

src/components/PastCommits.tsx
saulshanabrook: Yeah I would second this. It would be great to remove some of these props we pass down the whole hierarchy and put them on context.

src/components/FileList.tsx
```diff
@@ -130,6 +127,43 @@ export class FileList extends React.Component<IFileListProps, IFileListState> {
         }
       });
     }
+
+    if (!commands.hasCommand(CommandIDs.gitFileDiffWorking)) {
```
saulshanabrook: Why are we doing this conditionally?

src/components/diff/Diff.tsx
saulshanabrook: Yeah this should just be a functional stateless component. 

```typescript
function Diff(props: IDiffProps) {
   ...
}
```

src/components/diff/DiffWidget.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { IRenderMimeRegistry } from '@jupyterlab/rendermime';
+
+import { getRefValue, IDiffContext } from './model';
+import { Diff, isDiffSupported } from './Diff';
+import { JupyterLab } from '@jupyterlab/application';
+import { showDialog } from '@jupyterlab/apputils';
+import { PathExt } from '@jupyterlab/coreutils';
+import { style } from 'typestyle';
+
+export class DiffWidget extends Widget {
```
saulshanabrook: I believe this should extend `ReactWidget` instead of `Widget`. Then you fill in the `render` method and you don't have to mess with `ReactDOM.render`...

src/components/diff/NBDiffHeader.tsx
```diff
@@ -0,0 +1,56 @@
...
+  path: string;
+  diffContext: IDiffContext;
+}
+
+/**
+ * A React component to render the header which shows metadata around the diff
+ * being rendered. Shows the path to the file and the previous and current ref
+ * being used for the diff.
+ */
+export class NBDiffHeader extends React.Component<INBDiffHeaderProps, {}> {
```
saulshanabrook: This should also be a functional component

src/components/diff/NbDiff.tsx
```diff
@@ -0,0 +1,201 @@
...
+/**
+ * A React component which renders the diff is a single Notebook cell.
+ *
+ * This uses the NBDime PhosporJS CellDiffWidget internally. To get around the
+ * PhosporJS <=> ReactJS barrier, it uses React Refs(https://reactjs.org/docs/refs-and-the-dom.html)
+ *
+ * During component render, a Ref is created for the ReactDOM and after the component
+ * is mounted, the PhosporJS widget is created and attached to the Ref.
+ */
+export class CellDiff extends React.Component<ICellDiffProps, {}> {
```
saulshanabrook: This too.

src/components/diff/DiffWidget.tsx
jaipreet-s: This PR is on pre-1.0 changes and since nbdime isn't on 1.0 yet, and AFAIK `ReactWidget` isn't available on pre-1.0. I plan to refactor to use `ReactWidget` once

src/components/diff/NBDiffHeader.tsx
jaipreet-s: Done!

src/components/diff/Diff.tsx
jaipreet-s: In terms of necessity, I do think we need this top-level component to delegate to a filetype-specific component and maintain the registry of various diff providers. For e.g., the next thing on the plate is to add `<PlainTextDiff />` and it'll invoke that for non-ipynb files.

I'm good to make it a stateless component for less verbosity.

src/components/PastCommits.tsx
jaipreet-s: TIL ReactContext :-) 

My latest commit addresses this.

src/components/FileList.tsx
jaipreet-s: I kept it consistent with what we have in this file. I tested out by removing all the conditionals in this file and didn't see any issues on local testing, so went ahead and removed them all.

src/components/FileList.tsx
```diff
@@ -110,93 +107,119 @@ export class FileList extends React.Component<IFileListProps, IFileListState> {
...
+      caption: 'Open selected file',
+      execute: async () => {
+        try {
+          await this.openListedFile(
+            this.state.contextMenuTypeX,
+            this.state.contextMenuTypeY,
+            this.state.contextMenuFile,
+            this.props.app
+          );
+        } catch (err) {}
```
dleen: Would you at least want to log an error here?

src/components/FileList.tsx
```diff
@@ -110,93 +107,119 @@ export class FileList extends React.Component<IFileListProps, IFileListState> {
       disableFiles: false
     };
 
-    /** Add right-click menu options for files in repo
-     *
-     */
-
-    if (!commands.hasCommand(CommandIDs.gitFileOpen)) {
```
dleen: Why are the `if` statements no longer required?

src/components/FileList.tsx
jaipreet-s: This try catch should'nt been there in the first place. Removed.

src/components/FileList.tsx
jaipreet-s: It was left-over from the existing implementation. Commands have unique IDs so this isn't required.  Validated by checking examples in the `jupyterlab` repo

https://github.com/search?utf8=%E2%9C%93&q=%22commands.addCommand%22+user%3Ajupyterlab&type=Code&ref=advsearch&l=&l=

jaipreet-s:
@saulshanabrook @dleen Added a couple of commits https://github.com/jupyterlab/jupyterlab-git/pull/384/commits/1f1ebd1f996040de202c4a8937fe977a5ed7e64a https://github.com/jupyterlab/jupyterlab-git/pull/384/commits/3343c7f455432985b7a34b4b58f628fab4f6f89c that

1. Rebase on top of the JL1.0 changes
2. Use the latest version of nbdime compatible w/ JL1.0
3. Use the "ReactWidget" interface

jaipreet-s:
Hi @saulshanabrook , I'm going to merge this in. LMK if you have any feedback and I'll address that in a follow-up PR.

PR #385, opened 2019-07-23
Fix remote checkout

nibheis:
Currently, checking out a remote branch results in a detached branch.

The fix included in this branch does the following:

* it provides a new method to get the branch reference:
_get_branch_reference(self, branchname, current_path)

* it changes the behavior of the checkout_branch(...) method so that the '--track' parameter is passed during remote branches checkout.

Current checkout behavior is unchanged for local branch checkouts and also in case of error.

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -652,12 +652,7 @@ class Git:
         if p.returncode == 0:
             return { "code": 0, "message": my_output.decode("utf-8") }
         else:
-            response = {
-                "code": p.returncode,
-                "message": my_error.decode("utf-8"),
-            }
-            response["command"] = " ".join(cmd)
-            return response
+            return { "code": p.returncode, "message": my_error.decode("utf-8"), "command":  " ".join(cmd) }
 
     def checkout(self, filename, top_repo_path):
         """
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -646,7 +646,6 @@ class Git:
         )
 
         my_output, my_error = p.communicate()
-        p.returncode = 99
         if p.returncode == 0:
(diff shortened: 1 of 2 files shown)


fcollonval:
Thx for the contribution. Could you please rebase and add tests for this?

nibheis:
Tests added:

```
jupyterlab_git/tests/test_branch.py::test_checkout_branch_noref_success PASSED
jupyterlab_git/tests/test_branch.py::test_checkout_branch_noref_failure PASSED
jupyterlab_git/tests/test_branch.py::test_checkout_branch_remoteref_success PASSED
jupyterlab_git/tests/test_branch.py::test_checkout_branch_headsref_failure PASSED
jupyterlab_git/tests/test_branch.py::test_checkout_branch_headsref_success PASSED
jupyterlab_git/tests/test_branch.py::test_checkout_branch_remoteref_failure PASSED
jupyterlab_git/tests/test_branch.py::test_get_branch_reference_success PASSED
jupyterlab_git/tests/test_branch.py::test_get_branch_reference_faillure PASSED
```

nibheis:
... and just rebased.

Leletir:
@nibheis thank you very much for you work, I've been too slow ...

Leletir:
@fcollonval why don't we use fast string instead of str.format ?

fcollonval:
> why don't we use fast string instead of str.format ?

The jupyter server is officially supporting python 3.5 and above. So as this extension is quite popular and does not require stricter python version, it is better to support the same python version range. And unfortunately, this means we cannot use f-strings.

Leletir:
@fcollonval of course, I didn't think about the support of others python version. 
Thank you both for your work !

PR #395, opened 2019-08-09
Use the path relative to server root for diffs

### LINKED ISSUE #391: git diff doesn't work for files inside folders
Seems like path is not set correctly on /gitdiff POST request

in my case project structure is like
project->folder->notebook.ipynb

POST request to /gitdiff includes
`"file_path": "folder/notebook.ipynb" `
and it fails with 422 reply:

`HTTP 422: Unprocessable Entity (Invalid notebook: /home/jovyan/folder/notebook.ipynb)`
("project" is missing in the path)

instead of 
`"file_path": "project/folder/notebook.ipynb"`

jaipreet-s:
Closes #391 

This change adds a endpoint in the server extension to get the `server_root` and then resolve the path of the file relative to `server_root` before calling the nbdime API

### DIFF
--- a/jupyterlab_git/handlers.py
+++ b/jupyterlab_git/handlers.py
@@ -443,6 +443,14 @@ class GitConfigHandler(GitHandler):
         self.finish(json.dumps(response))
 
 
+class GitServerRootHandler(GitHandler):
+
+    def post(self):
+        # Similar to https://github.com/jupyter/nbdime/blob/master/nbdime/webapp/nb_server_extension.py#L90-L91
+        self.finish(json.dumps({
+            "server_root": getattr(self.contents_manager, 'root_dir', None)
+        }))
+
 def setup_handlers(web_app):
     """
     Setups all of the git command handlers.
@@ -471,7 +479,8 @@ def setup_handlers(web_app):
         ("/git/clone", GitCloneHandler),
         ("/git/upstream", GitUpstreamHandler),
         ("/git/config", GitConfigHandler),
-        ("/git/changed_files", GitChangedFilesHandler)
+        ("/git/changed_files", GitChangedFilesHandler),
+        ("/git/server_root", GitServerRootHandler)
     ]
 
     # add the baseurl to our paths
--- a/jupyterlab_git/handlers.py
+++ b/jupyterlab_git/handlers.py
@@ -445,7 +445,7 @@ class GitConfigHandler(GitHandler):
 
 class GitServerRootHandler(GitHandler):
 
-    def post(self):
+    def get(self):
         # Similar to https://github.com/jupyter/nbdime/blob/master/nbdime/webapp/nb_server_extension.py#L90-L91
         self.finish(json.dumps({
             "server_root": getattr(self.contents_manager, 'root_dir', None)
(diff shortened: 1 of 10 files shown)


jaipreet-s:
Ping @saulshanabrook

src/components/diff/DiffWidget.tsx
```diff
@@ -55,8 +62,27 @@ export function openDiffView(
...
 }
+
+/**
+ * Gets the path of the file relative to the Jupyter server root. This resolves the server root path after calling
+ * the `/git/server_root` REST API, and uses the Git repo root and the file path relative to the repo root to resolve
+ * the rest.
+ * @param path the file path relative to Git repo root
+ * @param topRepoPath the Git repo root path
+ */
+export async function getRelativeFilePath(
```
dleen: How often is this function called? It seems like you should only ever need to call `/git/server_root` once per session

jupyterlab_git/handlers.py
```diff
@@ -443,6 +443,14 @@ def post(self):
         self.finish(json.dumps(response))
 
 
+class GitServerRootHandler(GitHandler):
+
+    def post(self):
+        # Similar to https://github.com/jupyter/nbdime/blob/master/nbdime/webapp/nb_server_extension.py#L90-L91
+        self.finish(json.dumps({
+            "server_root": getattr(self.contents_manager, 'root_dir', None)
```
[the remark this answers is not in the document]
jaipreet-s: "null"
```
In [5]: data = { 
   ...: "server_root": None 
   ...: }                                                                                                                                                 

In [6]: json.dumps(data)                                                                                                                                  
Out[6]: '{"server_root": null}'
```

src/components/SinglePastCommitInfo.tsx
```diff
@@ -217,15 +217,16 @@ export class SinglePastCommitInfo extends React.Component<
                     <button
                       className={`${diffIconStyle}`}
                       title={'View file changes'}
-                      onClick={() => {
-                        openDiffView(
+                      onClick={async () => {
+                        await openDiffView(
```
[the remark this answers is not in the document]
jaipreet-s: This was needed because `openDiffView` is now async
https://github.com/jupyterlab/jupyterlab-git/pull/395/files#diff-114241b8de041aad5f23f605270e246bR22

src/components/diff/DiffWidget.tsx
jaipreet-s: This is called `once` when the user clicks the "Diff" button. This isn't invoked periodically on refresh or anything so its not a high call-rate use-case. 
I did think about calling it once during initialization but plumbing the value down through all the `Widgets` didn't seem worth the complexity in the current state. We can always refactor to make this more top-level if the need arises.

jupyterlab_git/handlers.py
```diff
@@ -443,6 +443,14 @@ def post(self):
         self.finish(json.dumps(response))
 
 
+class GitServerRootHandler(GitHandler):
+
+    def post(self):
```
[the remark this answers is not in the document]
jaipreet-s: I went with POST since most Git handlers were using POST. But you're right, this can totally be a GET, so will change it.

src/components/diff/DiffWidget.tsx
fcollonval: Caching the response in a `static` attribute would be a nice solution.

jaipreet-s: @fcollonval : Thats a good suggestion. I switched to just saving the result in a variable as a cache

PR #415, opened 2019-09-20
Standalone pip package

fcollonval:
This PR integrate the jupyterlab extension within the pip package as well as the activation of the server extension. When merged, installing through pip will be as simple as:

```bash
pip install jupyterlab-git
jupyter lab build
```

And if the user forgot the build, jupyter lab will prompt a dialog for it.


jupyter-config/jupyter_notebook_config.d/jupyterlab_git.json
```diff
@@ -0,0 +1,7 @@
+{
```
blink1073: I think this also needs to be added to the manifest file

PR #423, opened 2019-10-15
instead of `yarn` use `jlpm`, which is guaranteed to be installed alongside Jupyterlab

telamonian:
I ran into a spot of trouble trying to install the latest version of master. I (and presumably a lot of other jlab users) do not have a system-wide install of a `yarn` command. The package should instead use the `jlpm` command in place of `yarn` for all of the build commands. `jlpm` (which is really just a copy of `yarn` that is bundled with jlab) is guaranteed to be installed if a user has JupyterLab installed, and since JupyterLab is most certainly a required dependency of `jupyterlab-git`, it just makes more sense.

This PR accomplishes the `yarn` => `jlpm` substitution.


telamonian:
I also had to add an install of `jupyterlab` to the top of the CI script:

```
pip install "jupyterlab<2.0.0"
```

Given the way that the dependencies in `package.json` are specified, this extension "should" be compatible (or at least it claims that it is) with jlab up to `v2.0`, so that's what I install.

It seems to be intended that `pip install jupyterlab-git[test]` in `v0.8.2` will also pull in `jupyterlab`, but since currently the most recent release on Pypi is `v0.8.1`, that doesn't yet happen when the CI runs.

.travis.yml
```diff
@@ -6,6 +6,7 @@ cache:
   directories:
     - /home/travis/.yarn-cache/
 script:
+  - pip install "jupyterlab<2.0.0"
```
fcollonval: Could you please change it to `jupyterlab~=1.0` and update the `extras_require[test]` in the `setup.py` accordingly?

setup.py
```diff
@@ -5,8 +5,8 @@
 
 from setupbase import (
     create_cmdclass, install_npm, ensure_targets,
-    find_packages, combine_commands, ensure_python,
-    get_version, HERE
+    combine_commands, ensure_python, get_version,
+    HERE, which
```
fcollonval: Could you remove `which` as it's unused?

fcollonval:
Thx for the PR.

> It seems to be intended that pip install jupyterlab-git[test] in v0.8.2 will also pull in jupyterlab, but since currently the most recent release on Pypi is v0.8.1, that doesn't yet happen when the CI runs.

Actually, `pip install --find-links=dist jupyterlab_git[test]` is installing the latest commit. But unfortunately the requirements are installed after the `npm` commands is executed. And so `jlpm` ends up not being found if jupyterlab is not first installed 😞

.travis.yml
telamonian: `jupyterlab~=1.0` means the same thing as `jupyterlab<2.0.0`, right? I'm fine with making this change

setup.py
telamonian: Absolutely

.travis.yml
telamonian: Done!

setup.py
telamonian: Done!

telamonian:
I made all of the requested changes. Do the PR look good to you?

fcollonval:
Thx for the corrections.

Note: I'm correct. If you look at your last build, it reports:
```bash
$ pip install --find-links=dist jupyterlab_git[test]
Looking in links: dist
Collecting jupyterlab_git[test]
```

In the log you point to, `jlpm` being unfound, `pythin setup.py sdist` fails and so `dist` folder doesn't exist. Consequently pip pulls the version from pypi.

Issue #427, opened 2019-10-17
Notebooks with spaces in the name don't get recognized as .ipynb


thoralf-gutierrez:
## Description

We have notebook files with spaces, like `DataLake - Spark.ipynb`. Those file names get cut at the first space character in the history section, which means they don't get recognized as .ipynb files and we can't click on the diff icon to look at the diff for the notebook.

<img width="414" alt="JupyterLab" src="https://user-images.githubusercontent.com/5640848/67050464-71e10b00-f0ed-11e9-93f9-bd819ade5da0.png">

## Reproduce

1. Create a notebook with a space in its name, e.g. `DataLake - Spark.ipynb`
2. Create commit with the new notebook
3. Go to git pane, then the History tab 
4. Click on your commit
5. Notice that your notebook file name is cut and not recognized as .ipynb

## Expected behavior

It should show the full filename and be recognized as .ipynb. 

## Context

- Python package version: Version: 0.8.1
- Extension version: v0.8.2

thoralf-gutierrez:
I've identified the issue to https://github.com/jupyterlab/jupyterlab-git/blob/master/jupyterlab_git/git.py#L308 , we need to specify `split(maxplits=2)`. There are also no tests for `detailed_log`.

I will push a PR ready to fix both soon.

PR #428, opened 2019-10-18
Fix filename split in detailed log (#427)

thoralf-gutierrez:
Files with spaces weren't being recognized because of a split in `Git.detailed_logs` that was splitting too much, see #427

The `Git.detailed_logs` function was also missing tests so I added some to validate my fix. I used the same test structure as what I found in `test_branch.py`.

I validated this locally too and it works as expected :+1:

JupyterLab


fcollonval:
Fixes #427

PR #437, opened 2019-10-20
Simplified staging UI for git novices

telamonian:
For people who haven't used any VCS before, `git` can prove to be somewhat overwhelming. In particular, the untracked/unstaged/staged concept can seem very foreign to someone who up until now has never done any kind of "commit" more complicated than saving a file to disk. For that reason I think it would be idea to have an option to switch the staging UI to a "simplified" mode. As a model, I'm thinking along the lines of the UI in the Github desktop app:

image

It would simply show which files have changed, rather than splitting them into categories. When the commit button is pushed, all changed files will be automatically staged and commited (although there should also be something like the Github app's checkboxes to opt files out of a commit).

Additionally, some color coding of file names would be nice, a la VSCode:

image

Currently this PR is at proof of concept stage. It adds a `simpleStaging` flag to the user settings. When enabled, the staging UI is simplified, and only shows a single "changed" category. When disabled, the staging UI reverts back to the standard version.

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -497,6 +497,14 @@ class Git:
         my_output = subprocess.check_output(["git", "add", "-A"], cwd=top_repo_path)
         return my_output
 
+    def add_all_changed(self, top_repo_path):
+        """
+        Execute git add_all_changed command & return the result.
+        """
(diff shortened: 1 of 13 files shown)


telamonian:
closed in favor of #448

PR #440, opened 2019-10-30
Fix usage of old node in CI

fcollonval:
This fixes the bug appearing in the scheduled master check: https://travis-ci.org/jupyterlab/jupyterlab-git/builds/604902811?utm_medium=notification&utm_source=github_status

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ cache:
   pip: true
   directories:
     - /home/travis/.yarn-cache/
+before_install:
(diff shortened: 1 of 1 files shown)


fcollonval:
@telamonian thanks for the quick reaction. The root cause is the new version of puppeteer that does not support the default node used in travis:

```
error puppeteer@2.0.0: The engine "node" is incompatible with this module. Expected version ">=8.16.0". Got "8.12.0"
```

PR #441, opened 2019-11-03
Forbid pip to look at the index

fcollonval:


.travis.yml
```diff
@@ -10,8 +10,8 @@ before_install:
 install: pip install pytest "jupyterlab~=1.0"
 script:
   - python setup.py sdist
-  - pip install jupyterlab_git[test] --find-links=dist --no-deps --no-cache-dir -v
-  - pip install jupyterlab_git[test] --find-links=dist -v
+  - pip install jupyterlab_git[test] --pre --no-index --find-links=dist --no-deps --no-cache-dir -v
```
telamonian: Why the `--pre` flag?

.travis.yml
```diff
@@ -10,8 +10,8 @@ before_install:
...
+  - pip install jupyterlab_git[test]
```
telamonian: I suppose it does make sense to remove the `--find-links` flag here. Hopefully we won't ever try to bundle our Python package's Python deps :)

fcollonval:
Yes the `--no-index` forbids pip to request the package from pypi.org. So if the sdist fails, this will fail too.

The `--pre` is needed because if we set the version to be a prerelease (alpha, beta, rc) without it pip will ignore the sdist archive.

PR #447, opened 2019-11-07
enabled full typescript sourcemaps (with breakpoints and everything!)

telamonian:
This is not a user-facing change, but rather a bump in quality of life for the devs. To see the full benefit, this will also require some tweaks to the webpack build system in core (I think I have that already figured out, though). 

The upshot is that when you debug the code (eg, using Chrome devtools), instead of the compiled Javascript, you'll see the original `.ts` sources. You can also set breakpoints wherever you like in the sources, and they'll work correctly/consistently


telamonian:
You can actually see this in action now if you install Jupyterlab from the latest HEAD (with jupyterlab/jupyterlab#7541)

PR #449, opened 2019-11-07
fixes #442: addAllUnstagedFiles now has the correct behavior

### LINKED ISSUE #442: Stage all changes -misleading / unintuitive
There is a "Stage all changes" button to the left of the list of the "Changed" files in my repo. Clicking on it a `git add -A` is triggered, which tries to add (in my case many) more files then just those listed under changed. I find this behavior unintuitive, since I would suspect that only those files listed there would be added.

Would it be possible to add only those files which are listed there?

Thanks for your consideration.

telamonian:
Basically, we doing an inappropriate `git add -A` where we should have been doing `git add -u`. This has been fixed.

In various places in the code we refer to the 3 categories of files that git pays attention to as untracked, unstaged, and staged. We should try to be as consistent about these names throughout the program as we can be (eg, we should probably(?) change the text header over the middle file list from `Changed` -> `Unstaged`). Being consistent will help us to avoid this kind of unexpected behavior in the future.

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -464,6 +464,38 @@ export class GitExtension implements IGitExtension, IDisposable {
     return Promise.resolve(response);
   }
 
+  /** Make request to add all unstaged files into
+   * the staging area in repository 'path'
+   */
+  async addAllUnstaged(): Promise<Response> {
+    await this.ready;
+    const path = this.pathRepository;
+
+    if (path === null) {
+      return Promise.resolve(
+        new Response(
+          JSON.stringify({
+            code: -1,
+            message: 'Not in a git repository.'
+          })
+        )
+      );
+    }
+
+    try {
+      let response = await httpGitRequest('/git/add_all_unstaged', 'POST', {
+        top_repo_path: path
+      });
+      if (response.status !== 200) {
+        const data = await response.json();
+        throw new ServerConnection.ResponseError(response, data.message);
+      }
+      return response.json();
+    } catch (err) {
+      throw ServerConnection.NetworkError;
+    }
+  }
+
   /** Make request to add all untracked files into
    * the staging area in the repository
    */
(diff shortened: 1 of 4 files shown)


src/model.ts
```diff
@@ -464,6 +464,38 @@ export class GitExtension implements IGitExtension, IDisposable {
...
+      let response = await httpGitRequest('/git/add_all_unstaged', 'POST', {
+        top_repo_path: path
+      });
+      if (response.status !== 200) {
+        const data = await response.json();
+        throw new ServerConnection.ResponseError(response, data.message);
+      }
+      return response.json();
+    } catch (err) {
+      throw ServerConnection.NetworkError;
```
fcollonval: This should throw an instantiated error.

```suggestion
      throw new ServerConnection.NetworkError(err);
```

telamonian: all done

telamonian:
I haven't come across this issue, so I looked it up in the [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) they say that for the generic `Error` constructor, it does not matter at all if you use `new`, you get the same result.

On the other, the docs for [`TypeError`]() (which `ServerConnection.NetworkError` extends) don't explicitly say one way or the other, athe example they show is prefixed with `new`. So I guess for `TypeError` you do need to use `new`?

Given that, what exactly was:

```
throw ServerConnection.NetworkError;
```

doing in the first place, and why was it (at least compile-time) valid?

fcollonval:
When reworking the frontend, I got some warnings due to that expression `throw ServerConnection.NetworkError` when running the jest tests. They disappear when instantiating an object instead. But I didn't have time to go to the bottom of it to figure out the reason. It has maybe something to do with jest using node. And so having different rules of valid syntax? My knowledge of the JS universe is unfortunately not large enough 😢

PR #450, opened 2019-11-13
fixes #386: turn off by default overly agressive disabling of create branch/switch branch

### LINKED ISSUE #386: Overly aggressive disabling of branching
Thanks for all the work on this extension, it's coming along nicely!

I often will make uncommited changes to a repository, then later decide I want to commit them in a new branch. Git allows one to checkout a new branch when there are uncommited changes (or even staged changes), and then proceed as normal. However, when I try that in the UI, I get an error message saying 
```
Creating new branch disabled

You have staged changes in current branch. Please commit / discard them before creating a new branch.
```
Is this disabling of branching really necessary? It's not a hard git requirement (also, note that the error message is incorrect if there are unstaged changes). Apologies if there is already an issue discussing this, I didn't find one after a quick search.

telamonian:
Fixes #386

`juptyerlab-git` has recently received a large number of upgrades and fixes. It's almost to the point where I can actually use it in place of the `git` cmd line in my own workflow. There's just one last blocking issue: currently, we make it so that you need to have no changes whatsoever present in your repo in order to create or switch a branch.

Normally, `git` allows you to create a new branch regardless of changes, and it allows you to switch branches so long as the changed files are the same in your head and the branch you're trying to switch to. Personally, I think that breaking the functionality of the underlying `git` tool by default is an anti-pattern, and something we should avoid as much as we can.

On the other hand, I can also see the value in having settings that make things easier/less confusing for a git novice. Thus, I've added a flag to user settings, `"disableBranchWithChanges"`, that enables the old, more restrictive branch ops behavior, and which is off by default.

In a future PR, I think it would be good to pull together all of the settings aimed to making things easier into a single `"EZGit"` option (in the git menu, say) that can be toggled on or off with a single input.


damianavila:
@telamonian I have installed 0.9.0 from conda-forge and tried to change branches with but I am still getting the branch error, even when the setting is False (and is False by default as far as I understand). 

Screen Shot 2019-12-16 at 3 14 54 PM

Any ideas?
AFAIK, this one was included in the latest release, right?

telamonian:
@damianavila That's a real error from git itself, not something that jupyterlab-git is adding. Here's the old error message that would pop up when you had any changes:

```
Creating new branch disabled

You have staged changes in current branch. Please commit / discard them before creating a new branch.
```

On the other hand, the error message from your screenshot means that the branch that you're trying to switch to already has a different version of `untitled.txt` with different changes.

In other words, even if you tried to do the branch switch directly with the `git` cmd line tool, you'd still see the same error message.

damianavila:
Thanks, @telamonian, not sure how that file was modified on the other branch but starting from scratch I am not able to reproduce the issue anymore. Thanks!

PR #451, opened 2019-11-13
Update README.md

elgravel:
Modified installation steps, as the ones available previously were not enough to install this extension.


fcollonval:
@elgravel Thx for your contribution. The instructions are correct for the master.

For a particular version, I advice to set the repository to the tag commit (via the three combobox on the top left above the files list). See for example for v0.8.2 https://github.com/jupyterlab/jupyterlab-git/tree/v0.8.2

Issue #454, opened 2019-11-22
The git sidepanel does not refresh on file changes/autorefresh interval too long


telamonian:
ref: https://github.com/jupyterlab/jupyterlab-git/pull/448#issuecomment-554193273

Currently, when you make changes to a file on disk in a git repo, it takes a significant lag period before said changes show up in the git sidepanel UI. One quick fix would be to shorten the autorefresh interval. Even better though would be to listen for some flavor of "file changed" signal and trigger a refresh then.

Not sure exactly which signal to listen to, but I assume there's something relevant somewhere in one of:

- `@jupyterlab/docmanager`
- `@jupyterlab/docregistry`
- `@jupyterlab/filebrowser`

kgryte:
@telamonian Unfortunately, e.g., in `@jupyterlab/filebrowser`, the same polling logic and interval is as followed here (see <https://github.com/jupyterlab/jupyterlab/blob/78e7bf2d6848fe22f2b59200c832e60f83b6ad22/packages/filebrowser/src/model.ts#L90>).

If locking up the UI is a concern, it would be nice to have a server push-based model, rather than a pull-based model (e.g., via server sent events (SSEs)), but I sense that would entail a bit more plumbing than desired and also would seem to require that the backend of this plugin be refactored to be async.

IMO I think reducing the interval and maybe making it configurable should be the interim solution. I'll submit a PR to this end...

telamonian:
@kgryte Before you do a PR that just reduces the interval, take another look at the snippet you linked to above. In particular the `services.contents.fileChanged.connect(...)` line

telamonian:
Tracing it out, in the same constructor that sets the autorefresh interval for the filebrowser:

```typescript
    const refreshInterval = options.refreshInterval || DEFAULT_REFRESH_INTERVAL;

    const { services } = options.manager;
    services.contents.fileChanged.connect(this._onFileChanged, this);
    services.sessions.runningChanged.connect(this._onRunningChanged, this);

    ...

    this._poll = new Poll({
      factory: () => this.cd('.'),
      frequency: {
        interval: refreshInterval,
        backoff: true,
        max: 300 * 1000
      },
      standby: 'when-hidden'
    });
```

also connects the `_onFileChanged` method (of the filebrowser model that's being constructed) to the `fileChanged` signal of `services.contents` (which is a `Contents.IManager`). In turn, `_onFileChanged` will directly ring the `refresh` method of the model's `_poll`:

```
[...]

PR #459, opened 2019-11-26
Update refresh interval

kgryte:
This PR resolves https://github.com/jupyterlab/jupyterlab-git/issues/454 by

-   reducing the refresh interval to create a more responsive user experience. Previously, if a file change was made outside of the lab environment, it could take up to 10 seconds for the Git extension to register that a change had been made. After experimenting with different durations, I settled on a (maximum) duration of 3 seconds. In my experience, once I had to wait longer, I immediately began wondering if something was wrong. Maybe others are more patient.
-   exposing a `refreshInterval` configurable setting. This should allow users (and organizations) to adjust the polling behavior according to taste and desired performance profiles.
-   adding logic for listening to `fileChanged` events emitted by the default file browser extension and syncing the Git extension accordingly.

A future PR may want to consider refactoring the Git extension backend to be async, in order to ensure that the backend remains responsive when shelling out to execute Git commands. The current polling rate of 3 seconds should not be problematic, in and of itself, from a perf perspective within the frontend/UI. However, I cannot readily vouch for the performance implications of repeatedly performing synchronous commands in very large Git repositories on the backend.

By allowing the refresh interval to be configurable and documented accordingly, we should allow users (and organizations) to tailor as need be, regardless of the backend implementation.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -58,23 +58,36 @@ export default plugin;
 /**
  * Activate the running plugin.
  */
-function activate(
+async function activate(
   app: JupyterFrontEnd,
   mainMenu: IMainMenu,
   restorer: ILayoutRestorer,
   factory: IFileBrowserFactory,
   renderMime: IRenderMimeRegistry,
   settingRegistry: ISettingRegistry
-): IGitExtension {
-  const key = plugin.id;
+): Promise<IGitExtension> {
+  let settings: ISettingRegistry.ISettings;
 
+  // Register Git icons with the icon registry
   registerGitIcons(defaultIconRegistry);
 
-  // Create the Git model
-  let gitExtension = new GitExtension(app);
-
-  // Connect file browser with git model
+  // Get a reference to the default file browser extension
   const filebrowser = factory.defaultBrowser;
+
+  // Wait for the application and file browser extension to be restored:
+  await Promise.all([app.restored, filebrowser.model.restored]);
+
+  // Attempt to load application settings
+  try {
+    settings = await settingRegistry.load(plugin.id);
+  } catch (error) {
+    console.error(`Failed to load settings for the Git Exetnsion.\n${error}`);
+  }
+  // Create the Git model
+  const gitExtension = new GitExtension(app, settings);
+
+  // Sync the Git extension path with the file browser extension's current path
+  gitExtension.pathRepository = filebrowser.model.path;
 
   // Whenever the file browser path changes, sync the Git extension path
   filebrowser.model.pathChanged.connect(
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -106,6 +114,12 @@ function activate(
         createGitMenu(app, gitExtension, factory.defaultBrowser, settings),
         { rank: 60 }
       );
+
+      // Listen for changes to extension settings:
+      settings.changed.connect((settings: ISettingRegistry.ISettings) => {
+        gitExtension.refreshInterval = settings.composite
+          .refreshInterval as number;
+      });
```
telamonian: It looks like you'll need more than this to ensure that the `refreshInterval`'s in the model instance and the settings are in sync when jlab starts up/is refreshed.

src/model.ts
```diff
@@ -914,6 +924,7 @@ export class GitExtension implements IGitExtension, IDisposable {
   private _readyPromise: Promise<void> = Promise.resolve();
   private _pendingReadyPromise = 0;
   private _poll: Poll;
+  private _refreshInterval = 3000; // ms
```
telamonian: Related to above comment about model/settings sync, have you inspected `_refreshInterval` in devtools and made sure that it's in sync with the value in user settings under various circumstances (start up, page refresh, whatever other weird situation you can think of, etc)?

src/index.ts
kgryte: @telamonian Thanks! I added logic for syncing the refresh interval as soon as the settings are resolved which should cover the start-up and refresh cases mentioned below.

src/model.ts
kgryte: With my recent change, the refresh interval, when configured in the user settings, persisted during both start-up and page refresh. In which case, I believe these cases are now covered.

kgryte:
@telamonian Thanks for reviewing! I updated the PR and responded to your comments.

src/model.ts
```diff
@@ -27,20 +22,35 @@ export class GitExtension implements IGitExtension, IDisposable {
...
   }
 
+  /**
+   * Number of milliseconds between polling the file system for changes.
+   */
+  public get refreshInterval(): number {
+    return this._refreshInterval;
+  }
+
+  public set refreshInterval(v: number) {
```
telamonian: Is there a reasonable way that we could make the `set` private or protected (ie called only in constructor and on settings change)?

Essentially either: 

a) `set` has to be private and carefully handled, or 
b) designed in such a way as to ensure that it can never cause a desyncing with `settings`

Two possible designs for the latter are to:
- pass `settings` in somehow and use its `set` method (see this [bit of ThemeManager](https://github.com/jupyterlab/jupyterlab/blob/78e7bf2d6848fe22f2b59200c832e60f83b6ad22/packages/apputils/src/thememanager.ts#L195-L263) for various examples)
- leave the refresh logic in place put hoist the polling logic out of `GitExtension` and into `index.ts` where the `settings` is first materialized

Both make sense, but it also makes sense for the `GitExtension` model to not be tightly coupled to the concept of jlab's global application settings, so I'm leaning towards hoist.

src/model.ts
telamonian: If we remove `get refreshInterval` then it makes sense to also remove this private var, since it won't be used anywhere.

In general, in cases like this we should probably always try to work with values taken directly from the `settings` object, since that's effectively supposed to be the single source of truth.

src/model.ts
```diff
@@ -27,20 +22,35 @@ export class GitExtension implements IGitExtension, IDisposable {
...
         max: 300 * 1000
       },
       standby: 'when-hidden'
     });
   }
 
+  /**
+   * Number of milliseconds between polling the file system for changes.
+   */
+  public get refreshInterval(): number {
```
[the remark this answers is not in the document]
kgryte: Removed.

src/model.ts
kgryte: Removed.

src/model.ts
kgryte: Hoist ended being a bit messy, as the Git model currently has logic for managing the poll. So I opted for the former: linking the model with the settings.

kgryte:
@telamonian I updated my PR.  I opted for adding a public `settings` setter on the underlying Git model. As of now, in this PR

-   the model manages updates to the extension settings (in particular, the `refreshInterval`)
-   the model only begins watching a repository once the settings have been resolved (!). This is a departure from previous behavior in which polling began as soon as the model was instantiated.
-   as the model now monitors settings for changes, treating the registry as the sole source of truth, the `refreshInterval` should not drift, as could have happened according to logic in previous commits.

In my local testing, the `refreshInterval` was applied upon initial load, upon updating via the settings UI, upon refresh, and upon restarting and reloading JupyterLab.

Does the current logic satisfy review requirements?

kgryte:
@telamonian Fixed failing CI tests by adding appropriate guards for when settings have not yet been resolved. Should be ready for review.

kgryte:
@fcollonval I've updated the PR, removing the `settings` setter and refactoring the activation function as an `async` function.

kgryte:
Looks like CI failed due to an issue with `nvm`. If either @telamonian or @fcollonval could trigger a new build, that would be appreciated.

telamonian:
@kgryte LGTM! Your current pattern for consuming the setting minimizes both new public API and opportunities for the refresh interval value to become different from the setting, which makes me happy. Thank you for being willing to experiment with the PR, and for putting in the extra work required.

kgryte:
@telamonian Thanks for reviewing!

src/index.ts
```diff
@@ -58,62 +58,71 @@ export default plugin;
...
-  let gitExtension = new GitExtension(app);
-  // Connect file browser with git model
+  // Get a reference to the default file browser extension
   const filebrowser = factory.defaultBrowser;
+
+  // Attempt to load application settings
+  try {
+    settings = await settingRegistry.load(plugin.id);
+  } catch (error) {
+    console.error(`Failed to load settings for the Git Exetnsion.\n${error}`);
```
fcollonval:
```suggestion
    console.error(`Failed to load settings for the Git Extension.\n${error}`);
```

src/index.ts
```diff
@@ -58,62 +58,71 @@ export default plugin;
...
-      console.error(
-        `Failed to load settings for the Git Exetnsion.\n${reason}`
-      );
-    });
-
+  // Whenever a user adds/renames/saves/deletes/modifies a file within the lab environment, refresh the Git status
+  filebrowser.model.fileChanged.connect(() => gitExtension.refreshStatus());
+
+  // Provided we were able to load application settings, create the extension widgets
+  if (typeof settings !== void 0) {
```
fcollonval: Is there a particular reason to use that test? 
Could the following be enough?
```suggestion
  if (settings) {
```

kgryte: Updated.

src/index.ts
kgryte: Fixed.

kgryte:
@fcollonval Updated the PR and resolved lingering issues.

PR #461, opened 2019-11-27
Don't build JLab extension for sdist

### LINKED ISSUE #430: Consider unbundling labextension from Python extension
Currently, the Typescript for the labextension is bundled in with the Python package. This smells like an anti-pattern. Beyond the issues I ran into in #424, there's a bunch of other obvious problems. For example, every time there's a new release of the labextension, we'll also have to do a new release of the Python package, even if there's no changes to the Python code. A better solution would be to just call out to the `labextension` app in `setup.py` (`jupyter labextension install jupyterlab_git` for release, `jupyter labextension install .` for develop).

fcollonval:
Alternative solution for #430 compared to #458 

The purpose of the `sdist` package is to contain all source files needed for building a fully functional package. This implies that it needs building tools to work and this is the usage for python package shipping cython, C/C++ or Fortran code. So it makes sense to ship the Typescript files and associated configuration in `sdist`.

But as mentioned in #430 and #458, it does not make sense to build the JS files twice (once for `sdist` and one for `bdist`). So to respect the above comment, the production of the JLab extension should only be carried out when executing the installation or the `bdist` command.

The difference of the source and wheel archive can be seen below:

image

Fix #430

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ script:
   - pip install jupyterlab_git[test] --pre --no-index --find-links=dist --no-deps --no-cache-dir -v
   # Install the extension dependencies
   - pip install jupyterlab_git[test]
+  - jupyter serverextension list
   - jupyter labextension list
   - jupyter lab build
   - pytest jupyterlab_git -r ap
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ script:
   - jupyter labextension list
   - jupyter lab build
   - pytest jupyterlab_git -r ap
+  - jlpm install
   - jlpm run test
   - python -m jupyterlab.browser_check
   - jlpm run lint
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,7 +5,6 @@ include setupbase.py
 
 include package.json
 include *.config.js
-include setupJest.js
 include ts*.json
 include jupyterlab_git/labextension/*.tgz
 recursive-include jupyter-config *.json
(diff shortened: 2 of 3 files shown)


telamonian:
It does not make sense for the sdist to contain the typescript code, since it doesn't need to (or benefit from) being rebuilt on different platforms, unlike

> cython, C/C++ or Fortran code.

The `.tag.gz` produced by `npm pack` contains the non-minified, non-bundled `.js` and `.js.map` source files, which are more than sufficient for the sdist. For example, with #458, full typescript source maps are still available for debugging.

There's truly no benefit to shipping the `.ts` sources. It's redundant, it bloats the sdist, and it makes an already convoluted build system just slightly harder to maintain.

telamonian:
For reference, here's the definition of sdist from the docs:

> Source Distribution (or “sdist”)

>    A distribution format (usually generated using python setup.py sdist) that provides metadata and the essential source files needed for installing by a tool like pip, or for generating a Built Distribution.

The `.ts` sources are definitely not "essential" in the sense mentioned above, as they are not needed for installing with pip or for generating a bdist/wheel

fcollonval:
> The .ts sources are definitely not "essential" in the sense mentioned above, as they are not needed for installing with pip or for generating a bdist/wheel

That is one interpretation of the definition. But I can argue that the *source files* (i.e. the source of truth) are the `.ts` files and not the transpiled code. But as mentioned in #461, there are no point of endlessly arguing on which type of files and which commands "feels" right to ship and use in the setup script. The main point, as you said, is the use case - i.e. to be able to install the backend and the frontend extensions from a simple `pip install *something*`; where *something* can be the package name (to request pypi.org), the git url, a local folder or the source package.

So the latest version of #458 looks good to me and I'll close this one.

PR #462, opened 2019-11-30
Adjust commit textarea styling and element alignment

kgryte:
This PR

-   adds support for resizing the commit message `textarea`. This is important for multiline commit messages containing a subject and body. Currently, one is unable to resize the `textarea` to see the entire message, and, instead, one must scroll.
-   because the `textarea` is now resizable, aligns the commit box items to the starting alignment. Otherwise, when one resizes the `textarea`, the commit button moves, which does not seem either desirable or necessary.
-   tweaks the initial `textarea` height so that the button height and `textarea` height are approximately the same and appear aligned.

### DIFF
--- a/src/style/BranchHeaderStyle.ts
+++ b/src/style/BranchHeaderStyle.ts
@@ -113,18 +113,18 @@ export const textInputStyle = style({
 export const stagedCommitStyle = style({
   resize: 'none',
   display: 'flex',
-  alignItems: 'center',
+  alignItems: 'flex-start',
   margin: '8px'
 });
 
 export const stagedCommitMessageStyle = style({
   width: '75%',
   fontWeight: 300,
-  height: '32px',
+  height: '35px',
   overflowX: 'auto',
   border: 'var(--jp-border-width) solid var(--jp-border-color2)',
   flex: '20 1 auto',
-  resize: 'none',
+  resize: 'vertical',
   padding: '4px 8px',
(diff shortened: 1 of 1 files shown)


fcollonval:
@kgryte Thanks for this PR. 

Just a screenshot, so it is easier to see what we are talking about:
image

I have mixed feeling about the resizable text area. I would rather have something resizing with the content.

Do you think using _contentEditable_ could do the job?

Here are some references:
https://stackoverflow.com/questions/15865982/it-is-possible-to-expand-a-textarea-only-with-css
https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Editable_content

kgryte:
Agreed, automatic resizing would be preferable; however, I am not convinced that `contenteditable` is preferred.

Notably, one has to roll her own placeholder text logic for a `contenteditable` element.

As documented in the SO post, the way to make `textarea` elements dynamically resize is via JavaScript. This is doable, but I did not see this as being worth the effort atm.

Making the message area resizable already provides a UX win. Subsequent improvements, if desired, could be addressed in later PRs and based on user feedback.

The lack of resizing has already been flagged as problematic during internal reviews.

fcollonval:
@kgryte thank you for the detailed answer.

I'm fine with this PR. But as it concerns the UI, I would like to wait for feedback on this.

telamonian:
superceded by #480

PR #463, opened 2019-12-01
Add demo notebook

kgryte:
This PR

-   adds a demo notebook.

Having a demo notebook in the plugin repository is useful for two reasons:

1.  when opening the associated Binder, a user can immediately play around with a sample notebook having varied outputs, without needing to create one herself. This is particularly useful when testing the NBDime diffs.
2.  when performing local development, having a demo notebook is useful when testing out new and/or modified features.

The included demo notebook is based on a similar notebook from JupyterLab core.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@ A JupyterLab extension for version control using git
 
 ![](http://g.recordit.co/N9Ikzbyk8P.gif)
 
+To see the extension in action, open the example notebook included in the Binder [demo](https://mybinder.org/v2/gh/jupyterlab/jupyterlab-git/master?urlpath=lab).
+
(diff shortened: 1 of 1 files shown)


fcollonval:
Thanks for the PR @kgryte 

Let's follow JupyterLab structure and put the notebook in an `example` folder.

Could you also add a note in the README that there is a notebook to play with when using binder?

kgryte:
@fcollonval Renamed folder to [`examples`](https://github.com/jupyterlab/jupyterlab/tree/1d787dbe2c487b470c062c5ab1ac5961a1f58d24/examples) and added a note in the README.

PR #464, opened 2019-12-01
Adjust icon spacing in path header component

kgryte:
Currently, icon margins simply expand and contract as the Git side panel expands and contracts, leading to (arguably) unnatural icon alignment. This PR changes to fixed icon margins and ensures that the icons remain right-aligned as the side panel is resized.

Before:

image

After:

Screen Shot 2019-11-30 at 5 05 44 PM

### DIFF
--- a/src/style/PathHeaderStyle.ts
+++ b/src/style/PathHeaderStyle.ts
@@ -30,7 +30,7 @@ export const repoRefreshStyle = style({
   boxSizing: 'border-box',
   outline: 'none',
   padding: '0px 6px',
-  margin: 'auto 5px auto auto',
+  margin: 'auto 5px auto 5px',
   height: '24px',
 
   $nest: {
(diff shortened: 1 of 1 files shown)


fcollonval:
The current behaviour is inherited from the file browser in JupyterLab to have an homogeneous user experience.
But to be even more homogeneous, I would even like to get the repository path below the icons.

I would rather keep it that way. @telamonian @ellisonbg what is your opinion?

kgryte:
@fcollonval I'd be willing to submit a PR to core to bring the file browser extension in line with the behavior introduced by this PR. :)

fcollonval:
@kgryte that's a great idea.

kgryte:
**Update**: issue filed on [core](https://github.com/jupyterlab/jupyterlab/issues/7593).

PR #465, opened 2019-12-01
Adjust branch dropdown alignment

kgryte:
This PR

-   adjusts branch dropdown item alignment.

Currently, branch dropdown items are center aligned which is poor for readability, as one must constantly reset to a new visual alignment position before reading the next list item.

Before:

Screen Shot 2019-11-30 at 5 25 05 PM

After:

Screen Shot 2019-11-30 at 5 23 38 PM


fcollonval:
@kgryte thank you for this enhancement.

I'm fine with this PR. But as it concerns the UI, I would like to wait for feedback on this from others.

PR #469, opened 2019-12-01
Fixes #460: better handling of branch edge cases

### LINKED ISSUE #460: Simple staging is broken for newly init-ed repos
Currently, when you init a new repo and try to turn on simple staging, you get this error

```
TypeError: this._currentMarker is null model.js:764

The above error occurred in the <FileList> component:
    in FileList (created by GitPanel)
    in div (created by GitPanel)
    in GitPanel

...
```

This needs fixing.

telamonian:
Fixes #460

Simple staging was broken for newly init-ed repos. This turns out to have been due to broken lookup of branch names for the case of repos with no commits. This PR fixes the general issue by changing around the internals of the `branch_heads` and `get_current_branch` functions in `git.py` in the serverextension to be more robust in these situations.

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -412,15 +412,14 @@ class Git:
                     if is_current_branch:
                         current_branch = branch
 
-                # Remote branch is seleted use 'git branch -a' as fallback machanism
-                # to get add detached head on remote branch to preserve older functionality
-                # TODO : Revisit this to checkout new local branch with same name as remote
-                # when the remote branch is seleted, VS Code git does the same thing.
-                if not current_branch and self.get_current_branch(current_path) == "HEAD":
+                # Above can fail in certain cases, such as an empty repo with
+                # no commits. In that case, just fall back to determining
+                # current branch
+                if not current_branch:
                     branch = {
                         "is_current_branch": True,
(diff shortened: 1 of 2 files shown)


telamonian:
The bug was previously hidden by questionable error handling on the labextension side here:

https://github.com/jupyterlab/jupyterlab-git/blob/a78cac8482ff5d4c7258515b75d3dfd0606cc4bf/src/components/GitPanel.tsx#L87-L91

Basically, in the case of lookup failure of the current branch/current branch name, the default behavior up to this point has been to assume that the branch name is `master`. We may want to change that

jupyterlab_git/git.py
```diff
@@ -812,15 +801,17 @@ def get_current_branch(self, current_path):
         )
         output, error = p.communicate()
         if p.returncode == 0:
-            return output.decode("utf-8").strip()
+            return output.decode("utf-8").split('/')[-1].strip()
+        elif "not a symbolic ref" in error.decode("utf-8"):
```
fcollonval: Should we compare lower case error message to prevent error seen elsewhere #409:

```suggestion
        elif "not a symbolic ref" in error.decode("utf-8").lower():
```

PR #477, opened 2019-12-04
followup #450: fixes behavior when `"disableBranchWithChanges": true`

telamonian:
followup #450

Currently, if `"disableBranchWithChanges": true` is in settings, branch ops are disabled completely in all cases. This PR restores the old "disable when there are changes" behavior when the setting is `true`

### DIFF
--- a/src/components/GitPanel.tsx
+++ b/src/components/GitPanel.tsx
@@ -201,7 +201,10 @@ export class GitPanel extends React.Component<
               (this.props.settings.composite[
                 'disableBranchWithChanges'
               ] as boolean) &&
-              (!!this.state.unstagedFiles || !!this.state.stagedFiles)
+(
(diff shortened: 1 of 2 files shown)


src/components/GitPanel.tsx
```diff
@@ -198,10 +198,10 @@ export class GitPanel extends React.Component<
             stagedFiles={this.state.stagedFiles}
             data={this.state.branches}
             disabled={
-              this.state.pastCommits.length === 0 ||
               (this.props.settings.composite[
                 'disableBranchWithChanges'
-              ] as boolean)
+              ] as boolean) &&
+              (!!this.state.unstagedFiles || !!this.state.stagedFiles)
```
kgryte:
```suggestion
(
   (this.state.unstagedFiles && this.state.unstagedFiles.length) ||
   (this.state.stagedFiles && this.state.stagedFiles.length)
)
```

PR #480, opened 2019-12-04
resizable commit message input box

### LINKED ISSUE #225: Commit message box should have code font and resize based on message size
The commit message box should use the code font family, since commit messages are 'code-like' text. Additionally, it should start as a single line in height, and resize dynamically up to a max of around 10 lines. 

A non-styled visual prototype of this might look something like this:

telamonian:
fixes #225
supercedes #462

This replaces the current `textarea` that we use for the commit message with a `TextareaAutosize`, which is a drop-in replacement that autoresizes with the number of lines in the current input text. The `react-textarea-autosize` that provides this is minimal in size, and adds no extra deps beyond itself.

This PR will also carry over some the styling from #462 in order to make the resizable text input look nice next to the giant commit check mark button.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "nbdime": "~5.0.1",
     "react": "~16.8.4",
     "react-dom": "~16.8.4",
+    "react-textarea-autosize": "^7.1.2",
     "typestyle": "^2.0.1"
   },
   "devDependencies": {
@@ -71,6 +72,7 @@
     "@types/jest": "^24",
     "@types/react": "~16.8.13",
     "@types/react-dom": "~16.0.5",
+    "@types/react-textarea-autosize": "^4.3.5",
     "enzyme": "3.7.0",
     "enzyme-adapter-react-16": "1.7.0",
     "husky": "1.3.1",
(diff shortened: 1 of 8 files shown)


kgryte:
@telamonian Submitted a PR against your fork which brings over the styles from #462: https://github.com/telamonian/jupyterlab-git/pull/1.

telamonian:
@fcollonval Care to sign off on this?

telamonian:
@kgryte I just tested out the styling you added. Some of it clashes with the behavior of the resizable text box. I'll fix it

telamonian:
I fixed up the styling. Out of curiosity, I did a bit of a deep dive into the inner workings of the TextareaAutosize component. It's a bit opaque, but it seems to always have a height of

```
rows*(1em + floor(1em/4px))
```

which can't be represented exactly in CSS because CSS has no `floor` func. I've set the height of the commit button to a rough equivalent.

@fcollonval Care to sign off on this?

fcollonval:
I'll give a try and let you know. Thanks you both for yet another great improvement.

PR #482, opened 2019-12-05
fixes #476: updated animated gif in readme to show off all the progress made up to `v0.9.0`

### LINKED ISSUE #476: Update README screencast
The extension deserves a more recent screencast.

telamonian:
fixes #476

Here's the new `.gif`:

ui_glow_up_090

I thinks it's good enough (and much better than leaving in the obsolete one), but I'm open to suggestions for improving it. Most helpful would be suggestions in the form of a script of actions to perform while the `.gif` is recording


fcollonval:
Thanks @telamonian for the new shiny animation.

I have a done another one with the simple staging UI:

git_simple_staging

Unfortunately I came across a bug in the history view...

kgryte:
@lresende I think we need to update both, as the above gifs are now fairly out-of-date.

telamonian:
closing in favor of #576

PR #487, opened 2019-12-05
Fixes #481: Adjusts bottom panel margin

### LINKED ISSUE #481: styling defect: gap between list of logs and bottom of git UI panel in history tab
Currently, the bottom of the log list in the history tab looks a bit weird, especially if the git sidepanel is set narrow enough for there to be a horizontal scrollbar:

image

There's a gap between the bottom of the log list and the bottom of the actual `GitPanel` widget (outlined in red below):

image

We should fix this

kgryte:
This PR resolves #481 by

-   adjusting the bottom panel margin thus ensuring that (a) the panel content extends to the bottom of the Git panel and (b) any side scrollbars are flush against the bottom of the Git panel.

New behavior:

Screen Shot 2019-12-05 at 3 15 29 PM

### DIFF
--- a/src/style/HistorySideBarStyle.ts
+++ b/src/style/HistorySideBarStyle.ts
@@ -4,5 +4,6 @@ export const historySideBarStyle = style({
   display: 'flex',
   flexDirection: 'column',
   paddingLeft: 0,
-  overflowY: 'scroll'
+  overflowY: 'scroll',
+  marginBlockEnd: 0
 });


telamonian:
As a note (mostly for myself), the inappropriate CSS value of `margin-block-end: 1em` is getting set by the top level user agent stylesheet, so overriding it like this PR does is probably the only possible fix.

The user agent stylesheet also sets `margin-block-start: 1em`, which we may also want to override

PR #488, opened 2019-12-06
Remove unused svg files

fcollonval:
Remove unused svg files and CSS variables.


fcollonval:
I was hesitant also. But if we need to restore some svg files, we should find this PR easily :wink:

PR #490, opened 2019-12-06
Use ServerTest for handler testing

fcollonval:


telamonian:
@fcollonval Can you please explain the intention of this a little?

fcollonval:
@telamonian the idea is to prefer testing handlers against a notebook test server instead of over patching the handler machinery. I did this to encourage good test practise as patching can hide bugs.

Issue #492, opened 2019-12-07
`pexpect.spawn` does not support Windows


fcollonval:
Per the documentation, the currently used `pexpect.spawn` to handle git authentication does not support Windows:
https://pexpect.readthedocs.io/en/stable/overview.html#pexpect-on-windows

For Windows,  they recommand:
> Pexpect can be used on Windows to wait for a pattern to be produced by a child process, using pexpect.popen_spawn.PopenSpawn, or a file descriptor, using pexpect.fdpexpect.fdspawn.

ktaletsk:
Documentation also says "PopenSpawn is not a direct replacement for spawn". That suggests to me that we would need to add separate implementation of GitAuthInputWrapper for Windows. Any thoughts on that? I am willing to help if needed.

fcollonval:
@ktaletsk thank you for kicking in. Any help would be appreciated 😄 

Your proposal to handle Windows specifically is fine for me.

fcollonval:
Entering passwords with `pexpect` is always going to be a little shaky. I did some research, and it looks like it should be possible to pass auth to git via the `env` arg in `subprocess.Popen`. The key is the `GIT_ASKPASS` env var. I'll play around with it and see if I can get a working prototype.

Refs:
https://stackoverflow.com/q/1456269/425458
https://serverfault.com/a/976927/58565
https://github.com/libgit2/pygit2
https://coolaj86.com/articles/vanilla-devops-git-credentials-cheatsheet/
https://git-scm.com/docs/git-credential-store
https://git-scm.com/docs/gitcredentials
https://git-scm.com/book/en/v2/Git-Tools-Credential-Storage
https://dev.to/pratham/custom-git-credential-helper-pfa
https://stackoverflow.com/a/10054470/425458
https://stackoverflow.com/q/5343068/425458
https://git-scm.com/docs/api-credentials
https://alanedwardes.com/blog/posts/git-username-password-environment-variables/

_Originally posted by @telamonian in https://github.com/jupyterlab/jupyterlab-git/pull/493#issuecomment-568840306_

jaipreet-s:
This would also be a good opportunity to start adding integration testing + testing on windows in the CI setup

1. **Integration Testing**:  Use the server REST API to perform Git push/pull actions against a test Git repository
[...]

PR #493, opened 2019-12-07
Asynchronous git cmds

### LINKED ISSUE #439: Make subprocess call asynchronous
## Description

Make the git subprocess call asynchronously.

This is a follow-up of #435 to improve user experience in case such as #432 (summarize here below).

The /git/all_history api fetches info for all of the branches of a repo. This is implemented via a call to Git.branch. Part of the info that Git.branch fetches is the branch's tag, which is the result of this call:
```python
 command = ["git", "describe", "--tags", commit_sha] 
 p = subprocess.Popen( 
     command, 
     stdout=PIPE, 
     stderr=PIPE, 
     cwd=os.path.join(self.root_dir, current_path), 
 ) 
```

The tag is supposed to be the most recent tag prior to the head commit of a branch. Having to call git describe over and over is very slow for repos with lots of tags. For example, the core jlab repo has 7000+ tags. If you try to use jupyterlab-git on the core jlab repo, the call to /git/all_history can **take up to 15 seconds**, mostly because of fetching tag, and **freezes the entirety of jlab** in the meantime. The worst part is, we don't even use tag anywhere, so this both slow and pointless.

_originally posted by @telamonian in #432_

## Possible fix

Use `tornado.ioloop.IOLoop.run_in_executor` for wrapping subprocess execution - see:
- example: https://github.com/fcollonval/jupyter_conda/blob/master/jupyter_conda/envmanager.py#L86-L89
- documentation: https://www.tornadoweb.org/en/stable/guide/coroutines.html#calling-blocking-functions

fcollonval:
Fix #439 

Main changes:
- Create two utils functions `execute` and `execute_with_authentication` to handle asynchronously git command execution.
- Rework all unit tests to handle the asynchronous code - in particular patching is now done at the `execute` functions level.

Note: I took the liberty to run black formatter on the Python code. Sorry for the format noise.


fcollonval:
The failing test in `jupyterlab_git/tests/test_branch.py:724` seems to come from a erroneous expected behavior.

```
AssertionError: Calls not found.
Expected: [call(['git', 'for-each-ref', '--format=%(refname:short)%09%(objectname)%09%(upstream:short)%09%(HEAD)', 'refs/heads/'], cwd='/bin/test_curr_path'),
    call(['git', 'for-each-ref', '--format=%(refname:short)%09%(objectname)', 'refs/remotes/'], cwd='/bin/test_curr_path')]
Actual: [call(['git', 'for-each-ref', '--format=%(refname:short)%09%(objectname)%09%(upstream:short)%09%(HEAD)', 'refs/heads/'], cwd='/bin/test_curr_path')]
```

The actual value seems legit as the code implies early bail out:

https://github.com/jupyterlab/jupyterlab-git/blob/f838270be6e34370a63a24dc04530730cca9369a/jupyterlab_git/git.py#L370-L375

@telamonian do you have any idea why the test was not failing before??

Note to myself: the browser app seems to dislike the asynchronous call - running JLab with this branch does not prompt any error.

jupyterlab_git/git.py
```diff
@@ -3,61 +3,111 @@
...
+            returncode = p.exitstatus
+            p.close()  # close process
+            return returncode, response
+
+    current_loop = tornado.ioloop.IOLoop.current()
+    returncode, output = await current_loop.run_in_executor(
+        None, call_subprocess, cmdline, username, password, cwd, env
+    )
+
+    return returncode, output
```
telamonian: I don't love that the return types of `execute` and `execute_with_authentication` are different. Do they have to be?

jupyterlab_git/tests/test_branch.py
```diff
@@ -1,874 +1,831 @@
...
+            "feature-bar\t01234567899999abcdefghijklmnopqrstuvwxyz\t\t ",
+        ]
+        process_output_remotes = [
+            "origin/feature-foo\tabcdefghijklmnopqrstuvwxyz01234567890123",
+            "origin/master\tabcdefghijklmnopqrstuvwxyz01234567890123",
+        ]
+
+        mock_execute.side_effect = [
+            # Response for get all refs/heads
+            tornado.gen.maybe_future((128, "\n".join(process_output_heads), "")),
```
telamonian: The test is failing because the return code here needs to be 0, since this is meant to represent a successful response

jupyterlab_git/tests/test_branch.py
```diff
@@ -1,874 +1,831 @@
...
+            # Response for get all refs/remotes
+            tornado.gen.maybe_future((128, "\n".join(process_output_remotes), "")),
```
telamonian: `128` => `0` here too

jupyterlab_git/tests/test_branch.py
fcollonval: Thanks a lot @telamonian and sorry I should have seen that one...

jupyterlab_git/git.py
fcollonval: I fully agree (I started to work on that actually 😉 ). But `pexpect.spawn` is by definition unable to split stdout and stderr.

Looking around, [`pexpect.fdpexpect.fdspawn`](https://pexpect.readthedocs.io/en/stable/api/fdpexpect.html) could work (see [that discussion](https://stackoverflow.com/questions/27179383/extracting-stderr-from-pexpect)). It has the advantage that it should address #492 and be consistent with `subprocess.Popen`.

@telamonian @ktaletsk should we move to `fdspawn` to have a stream separation (and to support Windows) ?

telamonian:
Entering passwords with `pexpect` is always going to be a little shaky. I did some research, and it looks like it should be possible to pass auth to git via the `env` arg in `subprocess.Popen`. The key is the `GIT_ASKPASS` env var. I'll play around with it and see if I can get a working prototype.

Refs:
https://stackoverflow.com/q/1456269/425458
https://serverfault.com/a/976927/58565
https://github.com/libgit2/pygit2
https://coolaj86.com/articles/vanilla-devops-git-credentials-cheatsheet/
https://git-scm.com/docs/git-credential-store
https://git-scm.com/docs/gitcredentials
https://git-scm.com/book/en/v2/Git-Tools-Credential-Storage
https://dev.to/pratham/custom-git-credential-helper-pfa
https://stackoverflow.com/a/10054470/425458
https://stackoverflow.com/q/5343068/425458
https://git-scm.com/docs/api-credentials
https://alanedwardes.com/blog/posts/git-username-password-environment-variables/

fcollonval:
I modified the code to have only one `execute` function and I homogenize the API reply (+ ensure it was JSONable as expected by the frontend - especially when an error arises).

I propose to let the authentication as is for this PR. And to improve it in another one.

chrisjohn2306:
Is it possible for someone to approve this , so that the changes can be merged

chrisjohn2306:
When we are giving a wrong git url or wrong credentials the window is freezing again. Even with the async changes. Is something being missed out ?

fcollonval:
Thanks @telamonian and the others :wink: for JLab 2.0
And thanks for the reviews

Issue #495, opened 2019-12-08
Enhance REST API


fcollonval:
The current REST API could be improved by at least passing the git repository folder in the url like for example the [content notebook API](https://github.com/jupyter/jupyter/wiki/Jupyter-Notebook-Server-API#Notebook-and-file-contents-API). In particular this will allow to use the full spectra of HTTP verbs in a more RESTful spirit.

For example here are some interfaces from the notebook content API:

HTTP verb | URL | Action
-- | -- | --
GET | /api/contents/_path_/_file_ | Return a model for a file or directory. 
PUT | /api/contents/_path_/_file_ | Saves the file in the location specified by name and  path. 
PATCH | /api/contents/_path_/_file_ | PATCH renames a notebook without re-uploading content.
POST | /api/contents/_path_/_file_ | Create a new file or directory in the specified path. 
DELETE | /api/contents/_path_/_file_ | delete a file in the given path

Idea of evolution

Before | After | Note
-- | -- | --
POST /git/add | POST /git/_path_/add |
POST /git/add_all_untracked | | Should be merged with `add` using arguments to distinguish
POST /git/add_all_untracked | | Should be merged with `add` using arguments
POST /git/all_history | GET /git/_path_/all_history | Should we keep it?
POST /git/branch | GET /git/_path_/branches |
POST /git/changed_files | GET /git/_path_/changed_files |
POST /git/checkout | ? | They are 3 behaviors within this endpoint. Probably need to be distinguished
_ | GET /git/_path_/branches/_branch_ | Branch switch
_ | POST /git/_path_/branches/_branch_ | Branch creation
_ | ? | Discard all changes
POST /git/clone | POST /git/_path_ |
POST /git/commit | POST /git/_path_/commits |
POST /git/config | GET /git/_path_/config |  Split get and set config actions  
_  | POST /git/_path_/config | 
POST /git/delete_commit | DELETE /git/_path_/commits/_hash_ |
POST /git/detailed_log | GET /git/_path_/detailed_log |
POST /git/diff | GET /git/_path_/diff |
POST /git/diffcontent | ? | Can it be merged with the previous one? 
POST /git/init | POST /git/_path_ | To be merged with `clone`
POST /git/log | GET /git/_path_/log | History count as query argument 
POST /git/pull | POST /git/_path_/pull | Need to stay POST for authentication
POST /git/push | POST /git/_path_/push |
POST /git/reset | ? |
POST /git/reset_to_commit | ? | 
[...]

PR #505, opened 2019-12-24
Add remote command

fcollonval:
To start a respository with a remote from scratch, the extension is missing a command to add a remote repository. This PR fix that usage.

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -25,7 +25,7 @@ async def execute(
     password: "Optional[str]" = None,
 ) -> "Tuple[int, str, str]":
     """Asynchronously execute a command.
-    
+
     Args:
         cmdline (List[str]): Command line to be executed
         cwd (Optional[str]): Current working directory
@@ -102,6 +102,35 @@ async def execute(
     return code, output, error
 
 
+class GitRemote:
+    """Container class for remote git commands"""
+
+    def add(self, top_repo_path, url, name="origin"):
+        """Handle call to `git remote add` command.
+
+        top_repo_path: str
+            Top Git repository path
+        url: str
+            Git remote url
+        name: str
+            Remote name; default "origin"
+        """
+        cmd = ["git", "remote", "add", name, url]
+        p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, cwd=top_repo_path)
+        my_output, my_error = p.communicate()
+        if p.returncode == 0:
+            return {
+                "code": p.returncode,
+                "command": " ".join(cmd)
+            }
+        else:
+            return {
+                "code": p.returncode,
+                "command": " ".join(cmd),
+                "message": my_error.decode("utf-8").strip()
+            }
+
+
 class Git:
     """
     A single parent class containing all of the individual git methods in it.
@@ -110,6 +139,8 @@ class Git:
     def __init__(self, contents_manager):
         self.contents_manager = contents_manager
         self.root_dir = os.path.expanduser(contents_manager.root_dir)
+
+        self.remote = GitRemote()
 
     async def config(self, top_repo_path, **kwargs):
         """Get or set Git options.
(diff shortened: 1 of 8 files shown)


jupyterlab_git/git.py
```diff
@@ -60,6 +60,35 @@ def communicate(self):
             return response
 
 
+class GitRemote:
```
telamonian: Why is this a separate class?

fcollonval: The idea is to structure the backend with a syntax close to git commands; in this case, `git.remote.`_action_. But I could also integrate it directly as a method of `Git` class if you think it is cleaner.

jupyterlab_git/git.py
```diff
@@ -69,6 +98,8 @@ def __init__(self, contents_manager):
         self.contents_manager = contents_manager
         self.root_dir = os.path.expanduser(contents_manager.root_dir)
 
+        self.remote = GitRemote()
```
jaipreet-s: Git repos can (and usually do) have multiple remotes so this seems like a 1:N relationship. WDYT think about making this a list?

jupyterlab_git/handlers.py
```diff
@@ -521,6 +538,7 @@ def setup_handlers(web_app):
         ("/git/log", GitLogHandler),
         ("/git/pull", GitPullHandler),
         ("/git/push", GitPushHandler),
+        ("/git/remote/add", GitRemoteAddHandler),
```
jaipreet-s: I think this could be made a bit more RESTful. I agree with using `/git/remote` as the entity but perhaps we could use HTTP verbs to define our actions, which would also make it extensible for further remote-related use-cases.

- POST `/git/remote`: Add a new remote
- GET `/git/remote`: List/get remotes
- DELETE `/git/remote`: Remove a remote

src/gitMenuCommands.ts
```diff
@@ -103,4 +110,39 @@ export function addCommands(
...
+        console.warn('Not in a Git repository. Unable to add a remote.');
+        return;
+      }
+      let url = args['url'] as string;
+      let name = (args['name'] as string) || 'origin';
+
+      if (!url) {
+        const result = await InputDialog.getText({
+          title: 'Add a remote repository',
+          placeholder: 'Remote Git URL'
```
jaipreet-s: nitpick: _Remote Git repository URL_ is more accurate

src/gitMenuCommands.ts
```diff
@@ -103,4 +110,39 @@ export function addCommands(
...
+    label: 'Add remote repository',
+    caption: 'Add a Git remote repository',
+    isEnabled: () => model.pathRepository !== null,
+    execute: async args => {
+      if (model.pathRepository === null) {
+        console.warn('Not in a Git repository. Unable to add a remote.');
+        return;
+      }
+      let url = args['url'] as string;
+      let name = (args['name'] as string) || 'origin';
```
jaipreet-s: `Origin` is a sensible default to have, but since the server is already setting it, can we skip it at the frontend to avoid duplication?

jupyterlab_git/handlers.py
```diff
@@ -262,6 +262,23 @@ def post(self):
...
 
 
+class GitRemoteAddHandler(GitHandler):
+    """Handler for 'git remote add <name> <url>'."""
+
+    def post(self):
+        """POST request handler to add a remote."""
+        data = self.get_json_body()
+        top_repo_path = data["top_repo_path"]
+        name = data.get("name", "origin")
```
jaipreet-s: The "origin" default is coming up more than once. How about making it a constant and share it?

jupyterlab_git/git.py
fcollonval: I will refractor the code. The idea was to get a API close to the git command syntax. But this obviously give the impression some information are store in a `remote` attribute.

jupyterlab_git/handlers.py
fcollonval: I would definitely like to have a more RESTful API (see #495 for discussion). The trouble is that this requires the repository path to be part of the endpoint - especially for action without body (like GET and usually DELETE). So to stay coherent with the current API, I would rather keep the current logic. And let the change be done in another PR.

src/gitMenuCommands.ts
fcollonval: I agree

jupyterlab_git/git.py
fcollonval: I went for a direct method of the `Git` class.

PR #508, opened 2019-12-24
Use React.RefObject instead of id for plain text diff

fcollonval:
The plain text diff viewer is injected in the DOM by providing the function with a HTML node. Before this PR, the node was found by using a (hopefully) unique HTML id. But this is not a robust solution.

This PR uses [React reference](https://reactjs.org/docs/refs-and-the-dom.html) instead; mechanism recommended for such case.


telamonian:
Can you please give a description of what this PR is for? I have a notion, but it would be nice to have it spelled out clearly

PR #509, opened 2019-12-24
Improve history panel

### LINKED ISSUE #468: Supporting non-public and non-GitHub repositories
## Description

Upon clicking the file icon in the Git history panel for a single file, the extension attempts to [open](https://github.com/jupyterlab/jupyterlab-git/blob/a78cac8482ff5d4c7258515b75d3dfd0606cc4bf/src/components/SinglePastCommitInfo.tsx#L213) a public GitHub URL.

Such behavior is likely not desirable when a repository is not hosted or, if hosted, not hosted on GitHub.

## Reproduce

1. Open JupyterLab with the Git extension installed.
2. Click on the Git extension icon.
3. Click on the history tab.
4. Scroll down to a past commit.
5. Click on the commit.
6. Double click the file icon next to a changed file.

## Expected behavior

I'm not sure the optimal behavior here. For internal usage, opening a GitHub URL is not applicable. Maybe this could be configured?

Or potentially down the road, one may want to open the file directly within JupyterLab to see the diff for that file in that commit. In fact, maybe this should be the default behavior?

This would, I presume, require diffing of plain text files (see #406).

## Context

- Python package version: 0.9.0

- Extension version: 0.9.0

### LINKED ISSUE #485: History details are not properly refresh when reverting
git_simple_staging

When reverting to a commit, the previously opened file lists are not refreshed. This leads to an erroneous history view.

_Originally posted by @fcollonval in https://github.com/jupyterlab/jupyterlab-git/pull/482#issuecomment-562154553_

fcollonval:
Improve history panel and fix #485 

image

- Use dialog for deleting or resetting a commit
- Past commit node expand status switches when clicking anywhere - don't need to click explicitly on `Collapse`
- Move the diff icon on the far right (similar layout to the status panel)
- Clicking anywhere on the file line open the diff panel if available
- Use correctly React `key` to have subpanel updated correctly when resetting or deleting a commit
- Remove double click action querying github - to be more generic (Fix #468)


fcollonval:
Thanks for reviewing this @telamonian

PR #510, opened 2019-12-25
Refactor commit box

kgryte:
This PR

-   refactors the visual display of the commit box to be more in line with the GitHub desktop app (see OP in https://github.com/jupyterlab/jupyterlab-git/pull/448).
-   moves the rendering of the commit box from the file list component in order to more cleanly separate concerns (i.e., not clear that a file list component should also be responsible for committing changes).
-   moves the rendering of the commit box to the Git panel component, along with the logic for actually committing staged changes.
-   extracts commit box styling from the branch header styles into a separate file to more cleanly separate style concerns.
-   refactors unit tests (note: some of the previous tests would be better implemented as UI tests, e.g., using `puppeteer`).

## Notes

-   Currently, some of the tests fail locally, but these tests concern NBDiff and PlainTextDiff, neither of which should have been affected by this PR.

## Screenshots

Here are the proposed changes:

-   Light theme:

Screen Shot 2019-12-24 at 5 18 27 PM
Screen Shot 2019-12-24 at 5 19 32 PM

-   Dark theme:

Screen Shot 2019-12-24 at 5 18 55 PM
Screen Shot 2019-12-24 at 5 19 15 PM


telamonian:
Very pretty

kgryte:
@telamonian I did test with simple staging on. Here are the screenshots:

Screen Shot 2019-12-24 at 9 59 57 PM
Screen Shot 2019-12-24 at 10 00 11 PM
Screen Shot 2019-12-24 at 10 00 25 PM

src/components/CommitBox.tsx
```diff
@@ -1,114 +1,166 @@
...
+      <form className={commitFormClass}>
+        <input
+          className={commitSummaryClass}
+          type="text"
+          placeholder="Summary (required)"
+          title="Enter a commit message summary"
+          value={this.state.summary}
+          onChange={this._onSummaryChange}
+          onKeyPress={this._onSummaryKeyPress}
+        />
```
fcollonval: The advice is a summay of max. 50 characters. This could still be longer that a single line. It may be better to use a `TextareaAutosize` here also (with maybe a `maxRows` attribute).

kgryte: I updated the title attribute to _encourage_ users to enter a single line summary not exceeding 50 characters.

As for using a `textarea`, I ultimately opted against this, following the precedent of GitHub's desktop app which also uses a single line `input` box which horizontally scrolls upon overflow. Personally, I think this may be better in order to "nudge" users to limit their summaries to one line with few(er) characters, while encouraging them to place more extended discussions in the `description` box.

kgryte:
@fcollonval Thanks for reviewing! I updated the commit box styles so that the font color is set to a JupyterLab theme color when the commit button is disabled. Here is a screenshot:

Screen Shot 2019-12-27 at 11 19 51 AM

Do you find this preferable?

src/components/CommitBox.tsx
fcollonval: Thanks for the detailed answer @kgryte. I think we can merge as is; user feedback will tell if it does not fit.

kgryte:
@fcollonval Thanks for merging!

PR #511, opened 2019-12-27
Switch branch header tab order

kgryte:
This PR

-   switches the tab order in the Git panel.
-   aligns the UI more closely with GitHub desktop.
-   works for both default and simplified staging.

Proposed change:

Screen Shot 2019-12-26 at 10 44 13 PM

GitHub desktop:

Screen Shot 2019-12-26 at 10 44 41 PM


kgryte:
@fcollonval Thanks for reviewing!

PR #515, opened 2020-01-03
Pin repository path

fcollonval:
Add the ability to pin the Git repository. When the repository is pinned, the repository is not synchronized with the current file browser path.

pinRepo


fcollonval:
There are other file browser extension out there in which the filebrowser base path is not a good source of truth to determine the user working directory. In particular tree file view like [jupyterlab-unfold](https://github.com/martinRenou/jupyterlab-unfold) may bring unexpected behavior for this extension. It make sense to reboot this PR to cover that case.

penguinmenac3:
From the PR this feature looks awesome. Unfortunately it seems very outdated.

I tried to update it to the latest master, but failed miserably as I lack the detailed understanding of how this worked and how the overall plugin works.

Any chance for an update of this feature and maybe a merge into master?

sntgluca:
Dear all,

I'd love if this PR could move forward. I just came across the [jupyterlab-unfold](https://github.com/jupyterlab-contrib/jupyterlab-unfold/) extension too, and the fact that jupyterlab-git doesn't work well with it is a bummer.

May I ask about your plans and timelines?

Issue #514, opened 2019-12-27
Use a dialog to create a new branch


fcollonval:
The current UI to create a branch is not pretty:

![image](https://user-images.githubusercontent.com/8435071/71514289-3fe2e680-289e-11ea-8a3f-0dcb5b2a7825.png)

It would be better to use a dialog.

fcollonval:
@kgryte will you cover this when working on #513?

kgryte:
@fcollonval Yes, was planning on addressing this, as well.

kgryte:
This was resolved in #518.

PR #518, opened 2020-01-08
Refactor branch menu

### LINKED ISSUE #513: Use dialog to switch branch
I'll like to replace the current UI to switch branches by a dialog with a combobox. What do you think @kgryte @telamonian ?

image

kgryte:
This PR resolves #513 and #514 by

1.  completely refactoring how a user views, switches, and creates branches (the GitHub desktop app served as the reference design).
2.  providing a means to filter branches when choosing a desired branch.
3.  fixing a polling bug in the underlying extension model in which the model only refreshed the status, but not the list of branches, thus causing the dropdown to be out-of-sync if branches were created/deleted outside of the extension.
4.  adding title attributes (and corresponding unit tests) to all buttons and UI elements in order to provide reasonably helpful hints to users not acquainted with the iconography or the UI.
5.  leveraging Material UI React components for rendering lists and displaying dialogs in order to make components more consistent from a development perspective.
6.  refactoring styles to better align with CSS best practice and to follow `typestyle` naming conventions.
7.  splitting rendering logic into separate (smaller) methods to aid development and to assist developers in being able to "grok" rendering logic.
8.  relying on the underlying extension model as the sole source of rendering truth, rather than a mixture of properties and state, thus allowing components to be more easily decoupled.
9.  refactoring unit tests.
10. adding source code documentation to refactored components.

## Demos

1.  Basic usage (light theme)

Normal staging:

git_basic_usage

Simplified staging:

git_basic_usage_simplified_staging

2.  Basic usage (dark theme)

git_basic_usage_dark

3.  Branch menu filter (light theme)

git_menu_filter

4.  Branch menu filter (dark theme)

git_menu_filter_dark

5.  Switching branches (light theme)

git_switch_branches

6.  Switching branches (dark theme)

git_switch_branches_dark

7.  Creating branch (light theme)

git_create_branch

8.  Creating branch (dark theme)

git_create_branch_dark

9.  Attempting to switch branches when branching with changes is disabled

git_switch_branches_disabled

10. Attempting to create a branch when branching with changes is disabled

git_create_branches_disabled

## Notes

-   This PR has implications for https://github.com/jupyterlab/jupyterlab-git/pull/515, which, if this PR is accepted and the changes in that PR are desired, should be refactored to be based on the work in this PR.
-   This PR has implications for https://github.com/jupyterlab/jupyterlab-git/pull/509, which, if this PR is accepted and the changes in that PR are desired, would need some minor modifications if based on the work in this PR.

### DIFF
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -153,7 +153,7 @@ export class BranchMenu extends React.Component<
   /**
    * Renders menu items.
    *
-   * @returns fragment array
+   * @returns array of React elements
    */
   private _renderItems(): React.ReactElement[] {
     return this.state.branches.map(this._renderItem, this);
@@ -164,7 +164,7 @@ export class BranchMenu extends React.Component<
    *
    * @param branch - branch
    * @param idx - item index
-   * @returns fragment
+   * @returns React element
    */
   private _renderItem(
     branch: Git.IBranch,
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -54,7 +54,12 @@ export interface IBranchMenuState {
   /**
    * Current branch name.
    */
-  branch: string;
+  current: string;
+
+  /**
+   * Current list of branches.
+   */
+  branches: Git.IBranch[];
 }
 
 /**
@@ -73,6 +78,8 @@ export class BranchMenu extends React.Component<
   constructor(props: IBranchMenuProps) {
     super(props);
 
+    const repo = this.props.model.pathRepository;
+
     // When the repository changes, we're likely to have a new set of branches:
     this.props.model.repositoryChanged.connect(this._syncState, this);
 
@@ -85,9 +92,8 @@ export class BranchMenu extends React.Component<
     this.state = {
       filter: '',
       branchDialog: false,
-      branch: this.props.model.pathRepository
-        ? this.props.model.currentBranch.name
-        : ''
+      current: repo ? this.props.model.currentBranch.name : '',
+      branches: repo ? this.props.model.branches : []
     };
   }
 
@@ -145,7 +151,7 @@ export class BranchMenu extends React.Component<
    * @returns fragment array
    */
   private _renderItems = () => {
-    return this.props.model.branches.map(this._renderItem);
+    return this.state.branches.map(this._renderItem);
   };
 
   /**
@@ -166,7 +172,7 @@ export class BranchMenu extends React.Component<
         title={`Switch to branch: ${branch.name}`}
         className={classes(
           listItemClass,
-          branch.name === this.state.branch ? activeListItemClass : null
+          branch.name === this.state.current ? activeListItemClass : null
         )}
         key={idx}
         onClick={this._onBranchClickFactory(branch.name)}
@@ -181,11 +187,10 @@ export class BranchMenu extends React.Component<
    * Syncs the component state with the underlying model.
    */
   private _syncState = () => {
-    console.log('SYNCING STATE');
+    const repo = this.props.model.pathRepository;
     this.setState({
-      branch: this.props.model.pathRepository
-        ? this.props.model.currentBranch.name
-        : ''
+      current: repo ? this.props.model.currentBranch.name : '',
+      branches: repo ? this.props.model.branches : []
     });
   };
 
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -98,7 +98,7 @@ export class BranchMenu extends React.Component<
   /**
    * Callback invoked when a component will no longer be mounted.
    */
-  componentWillUnmount() {
+  componentWillUnmount(): void {
     this._removeListeners();
   }
 
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -149,9 +149,9 @@ export class BranchMenu extends React.Component<
    *
    * @returns fragment array
    */
-  private _renderItems = () => {
+  private _renderItems() {
     return this.state.branches.map(this._renderItem);
-  };
+  }
 
   /**
    * Renders a menu item.
@@ -160,7 +160,7 @@ export class BranchMenu extends React.Component<
    * @param idx - item index
    * @returns fragment
    */
-  private _renderItem = (branch: Git.IBranch, idx: number) => {
+  private _renderItem(branch: Git.IBranch, idx: number) {
     // Perform a "simple" filter... (TODO: consider implementing fuzzy filtering)
     if (this.state.filter && !branch.name.includes(this.state.filter)) {
       return null;
@@ -180,12 +180,12 @@ export class BranchMenu extends React.Component<
         {branch.name}
       </ListItem>
     );
-  };
+  }
 
   /**
    * Adds model listeners.
    */
-  private _addListeners = () => {
+  private _addListeners() {
     // When the repository changes, we're likely to have a new set of branches:
     this.props.model.repositoryChanged.connect(this._syncState, this);
 
@@ -194,54 +194,54 @@ export class BranchMenu extends React.Component<
 
     // When the status changes, we may have checked out a new branch (e.g., via the command-line and not via the extension):
     this.props.model.statusChanged.connect(this._syncState, this);
-  };
+  }
 
   /**
    * Removes model listeners.
    */
-  private _removeListeners = () => {
+  private _removeListeners() {
     this.props.model.repositoryChanged.disconnect(this._syncState, this);
     this.props.model.headChanged.disconnect(this._syncState, this);
     this.props.model.statusChanged.disconnect(this._syncState, this);
-  };
+  }
 
   /**
    * Syncs the component state with the underlying model.
    */
-  private _syncState = () => {
+  private _syncState() {
     const repo = this.props.model.pathRepository;
     this.setState({
       current: repo ? this.props.model.currentBranch.name : '',
       branches: repo ? this.props.model.branches : []
     });
-  };
+  }
 
   /**
    * Callback invoked upon a change to the menu filter.
    *
    * @param event - event object
    */
-  private _onFilterChange = (event: any) => {
+  private _onFilterChange(event: any) {
     this.setState({
       filter: event.target.value
     });
-  };
+  }
 
   /**
    * Callback invoked to reset the menu filter.
    */
-  private _resetFilter = () => {
+  private _resetFilter() {
     this.setState({
       filter: ''
     });
-  };
+  }
 
   /**
    * Callback invoked upon clicking a button to create a new branch.
    *
    * @param event - event object
    */
-  private _onNewBranchClick = () => {
+  private _onNewBranchClick() {
     if (!this.props.branching) {
       showErrorMessage('Creating a new branch is disabled', CHANGES_ERR_MSG);
       return;
@@ -249,16 +249,16 @@ export class BranchMenu extends React.Component<
     this.setState({
       branchDialog: true
     });
-  };
+  }
 
   /**
    * Callback invoked upon closing a dialog to create a new branch.
    */
-  private _onNewBranchDialogClose = () => {
+  private _onNewBranchDialogClose() {
     this.setState({
       branchDialog: false
     });
-  };
+  }
 
   /**
    * Returns a callback which is invoked upon clicking a branch name.
@@ -266,7 +266,7 @@ export class BranchMenu extends React.Component<
    * @param branch - branch name
    * @returns callback
    */
-  private _onBranchClickFactory = (branch: string) => {
+  private _onBranchClickFactory(branch: string) {
     const self = this;
     return onClick;
 
@@ -311,5 +311,5 @@ export class BranchMenu extends React.Component<
     function onError(err: any) {
       showErrorMessage('Error switching branch', err.message);
     }
-  };
+  }
 }
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -79,7 +79,6 @@ export class BranchMenu extends React.Component<
     super(props);
 
     const repo = this.props.model.pathRepository;
-    this._addListeners();
 
     this.state = {
       filter: '',
@@ -87,6 +86,13 @@ export class BranchMenu extends React.Component<
       current: repo ? this.props.model.currentBranch.name : '',
       branches: repo ? this.props.model.branches : []
     };
+  }
+
+  /**
+   * Callback invoked immediately after mounting a component (i.e., inserting into a tree).
+   */
+  componentDidMount(): void {
+    this._addListeners();
   }
 
   /**
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -185,7 +185,7 @@ export class BranchMenu extends React.Component<
   /**
    * Adds model listeners.
    */
-  private _addListeners() {
+  private _addListeners(): void {
     // When the repository changes, we're likely to have a new set of branches:
     this.props.model.repositoryChanged.connect(this._syncState, this);
 
@@ -199,7 +199,7 @@ export class BranchMenu extends React.Component<
   /**
    * Removes model listeners.
    */
-  private _removeListeners() {
+  private _removeListeners(): void {
     this.props.model.repositoryChanged.disconnect(this._syncState, this);
     this.props.model.headChanged.disconnect(this._syncState, this);
     this.props.model.statusChanged.disconnect(this._syncState, this);
@@ -208,7 +208,7 @@ export class BranchMenu extends React.Component<
   /**
    * Syncs the component state with the underlying model.
    */
-  private _syncState() {
+  private _syncState(): void {
     const repo = this.props.model.pathRepository;
     this.setState({
       current: repo ? this.props.model.currentBranch.name : '',
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -186,9 +186,6 @@ export class BranchMenu extends React.Component<
    * Adds model listeners.
    */
   private _addListeners(): void {
-    // When the repository changes, we're likely to have a new set of branches:
-    this.props.model.repositoryChanged.connect(this._syncState, this);
-
     // When the HEAD changes, decent probability that we've switched branches:
     this.props.model.headChanged.connect(this._syncState, this);
 
@@ -200,7 +197,6 @@ export class BranchMenu extends React.Component<
    * Removes model listeners.
    */
   private _removeListeners(): void {
-    this.props.model.repositoryChanged.disconnect(this._syncState, this);
     this.props.model.headChanged.disconnect(this._syncState, this);
     this.props.model.statusChanged.disconnect(this._syncState, this);
   }
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -200,9 +200,9 @@ export class BranchMenu extends React.Component<
    * Removes model listeners.
    */
   private _removeListeners = () => {
-    this.props.model.repositoryChanged.disconnect(this._syncState);
-    this.props.model.headChanged.disconnect(this._syncState);
-    this.props.model.statusChanged.disconnect(this._syncState);
+    this.props.model.repositoryChanged.disconnect(this._syncState, this);
+    this.props.model.headChanged.disconnect(this._syncState, this);
+    this.props.model.statusChanged.disconnect(this._syncState, this);
   };
 
   /**
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -150,7 +150,7 @@ export class BranchMenu extends React.Component<
    * @returns fragment array
    */
   private _renderItems() {
-    return this.state.branches.map(this._renderItem);
+    return this.state.branches.map(this._renderItem, this);
   }
 
   /**
@@ -221,27 +221,27 @@ export class BranchMenu extends React.Component<
    *
    * @param event - event object
    */
-  private _onFilterChange(event: any) {
+  private _onFilterChange = (event: any): void => {
     this.setState({
       filter: event.target.value
     });
-  }
+  };
 
   /**
    * Callback invoked to reset the menu filter.
    */
-  private _resetFilter() {
+  private _resetFilter = (): void => {
     this.setState({
       filter: ''
     });
-  }
+  };
 
   /**
    * Callback invoked upon clicking a button to create a new branch.
    *
    * @param event - event object
    */
-  private _onNewBranchClick() {
+  private _onNewBranchClick = (): void => {
     if (!this.props.branching) {
       showErrorMessage('Creating a new branch is disabled', CHANGES_ERR_MSG);
       return;
@@ -249,16 +249,16 @@ export class BranchMenu extends React.Component<
     this.setState({
       branchDialog: true
     });
-  }
+  };
 
   /**
    * Callback invoked upon closing a dialog to create a new branch.
    */
-  private _onNewBranchDialogClose() {
+  private _onNewBranchDialogClose = (): void => {
     this.setState({
       branchDialog: false
     });
-  }
+  };
 
   /**
    * Returns a callback which is invoked upon clicking a branch name.
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -50,6 +50,11 @@ export interface IBranchMenuState {
    * Boolean indicating whether to show a dialog to create a new branch.
    */
   branchDialog: boolean;
+
+  /**
+   * Current branch name.
+   */
+  branch: string;
 }
 
 /**
@@ -67,9 +72,14 @@ export class BranchMenu extends React.Component<
    */
   constructor(props: IBranchMenuProps) {
     super(props);
+    this.props.model.repositoryChanged.connect(this._onRepositoryChange, this);
+    this.props.model.headChanged.connect(this._onHeadChange, this);
     this.state = {
       filter: '',
-      branchDialog: false
+      branchDialog: false,
+      branch: this.props.model.pathRepository
+        ? this.props.model.currentBranch.name
+        : ''
     };
   }
 
@@ -148,9 +158,7 @@ export class BranchMenu extends React.Component<
         title={`Switch to branch: ${branch.name}`}
         className={classes(
           listItemClass,
-          branch.name === this.props.model.currentBranch.name
-            ? activeListItemClass
-            : null
+          branch.name === this.state.branch ? activeListItemClass : null
         )}
         key={idx}
         onClick={this._onBranchClickFactory(branch.name)}
@@ -159,6 +167,31 @@ export class BranchMenu extends React.Component<
         {branch.name}
       </ListItem>
     );
+  };
+
+  /**
+   * Callback invoked upon a change to the repository path.
+   */
+  private _onRepositoryChange = () => {
+    this._syncState();
+  };
+
+  /**
+   * Callback invoked upon a change to the current HEAD.
+   */
+  private _onHeadChange = () => {
+    this._syncState();
+  };
+
+  /**
+   * Syncs the component state with the underlying model.
+   */
+  private _syncState = () => {
+    this.setState({
+      branch: this.props.model.pathRepository
+        ? this.props.model.currentBranch.name
+        : ''
+    });
   };
 
   /**
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -187,9 +187,7 @@ export class BranchMenu extends React.Component<
    * Syncs the component state with the underlying model.
    */
   private _syncState = () => {
-    console.log('SYNCING STATE');
     const repo = this.props.model.pathRepository;
-    console.log(this.props.model.branches);
     this.setState({
       current: repo ? this.props.model.currentBranch.name : '',
       branches: repo ? this.props.model.branches : []
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -187,7 +187,9 @@ export class BranchMenu extends React.Component<
    * Syncs the component state with the underlying model.
    */
   private _syncState = () => {
+    console.log('SYNCING STATE');
     const repo = this.props.model.pathRepository;
+    console.log(this.props.model.branches);
     this.setState({
       current: repo ? this.props.model.currentBranch.name : '',
       branches: repo ? this.props.model.branches : []
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -173,7 +173,7 @@ export class BranchMenu extends React.Component<
           listItemClass,
           branch.name === this.state.current ? activeListItemClass : null
         )}
-        key={idx}
+        key={branch.name}
         onClick={this._onBranchClickFactory(branch.name)}
       >
         <span className={listItemIconClass} />
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -181,6 +181,7 @@ export class BranchMenu extends React.Component<
    * Syncs the component state with the underlying model.
    */
   private _syncState = () => {
+    console.log('SYNCING STATE');
     this.setState({
       branch: this.props.model.pathRepository
         ? this.props.model.currentBranch.name
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -105,9 +105,9 @@ export class BranchMenu extends React.Component<
   /**
    * Renders the component.
    *
-   * @returns fragment
+   * @returns React element
    */
-  render() {
+  render(): React.ReactElement {
     return (
       <div className={wrapperClass}>
         <div className={filterWrapperClass}>
@@ -155,7 +155,7 @@ export class BranchMenu extends React.Component<
    *
    * @returns fragment array
    */
-  private _renderItems() {
+  private _renderItems(): React.ReactElement[] {
     return this.state.branches.map(this._renderItem, this);
   }
 
@@ -166,7 +166,10 @@ export class BranchMenu extends React.Component<
    * @param idx - item index
    * @returns fragment
    */
-  private _renderItem(branch: Git.IBranch, idx: number) {
+  private _renderItem(
+    branch: Git.IBranch,
+    idx: number
+  ): React.ReactElement | null {
     // Perform a "simple" filter... (TODO: consider implementing fuzzy filtering)
     if (this.state.filter && !branch.name.includes(this.state.filter)) {
       return null;
@@ -278,7 +281,7 @@ export class BranchMenu extends React.Component<
      * @private
      * @param event - event object
      */
-    function onClick() {
+    function onClick(): void {
       if (!self.props.branching) {
         showErrorMessage('Switching branches is disabled', CHANGES_ERR_MSG);
         return;
@@ -298,7 +301,7 @@ export class BranchMenu extends React.Component<
      * @private
      * @param result - result
      */
-    function onResolve(result: any) {
+    function onResolve(result: any): void {
       if (result.code !== 0) {
         showErrorMessage('Error switching branch', result.message);
       }
@@ -310,7 +313,7 @@ export class BranchMenu extends React.Component<
      * @private
      * @param err - error
      */
-    function onError(err: any) {
+    function onError(err: any): void {
       showErrorMessage('Error switching branch', err.message);
     }
   }
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -79,15 +79,7 @@ export class BranchMenu extends React.Component<
     super(props);
 
     const repo = this.props.model.pathRepository;
-
-    // When the repository changes, we're likely to have a new set of branches:
-    this.props.model.repositoryChanged.connect(this._syncState, this);
-
-    // When the HEAD changes, decent probability that we've switched branches:
-    this.props.model.headChanged.connect(this._syncState, this);
-
-    // When the status changes, we may have checked out a new branch (e.g., via the command-line and not via the extension):
-    this.props.model.statusChanged.connect(this._syncState, this);
+    this._addListeners();
 
     this.state = {
       filter: '',
@@ -95,6 +87,13 @@ export class BranchMenu extends React.Component<
       current: repo ? this.props.model.currentBranch.name : '',
       branches: repo ? this.props.model.branches : []
     };
+  }
+
+  /**
+   * Callback invoked when a component will no longer be mounted.
+   */
+  componentWillUnmount() {
+    this._removeListeners();
   }
 
   /**
@@ -181,6 +180,29 @@ export class BranchMenu extends React.Component<
         {branch.name}
       </ListItem>
     );
+  };
+
+  /**
+   * Adds model listeners.
+   */
+  private _addListeners = () => {
+    // When the repository changes, we're likely to have a new set of branches:
+    this.props.model.repositoryChanged.connect(this._syncState, this);
+
+    // When the HEAD changes, decent probability that we've switched branches:
+    this.props.model.headChanged.connect(this._syncState, this);
+
+    // When the status changes, we may have checked out a new branch (e.g., via the command-line and not via the extension):
+    this.props.model.statusChanged.connect(this._syncState, this);
+  };
+
+  /**
+   * Removes model listeners.
+   */
+  private _removeListeners = () => {
+    this.props.model.repositoryChanged.disconnect(this._syncState);
+    this.props.model.headChanged.disconnect(this._syncState);
+    this.props.model.statusChanged.disconnect(this._syncState);
   };
 
   /**
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -195,7 +195,7 @@ export class BranchMenu extends React.Component<
     // When the HEAD changes, decent probability that we've switched branches:
     this.props.model.headChanged.connect(this._syncState, this);
 
-    // When the status changes, we may have checked out a new branch (e.g., via the command-line and not via the extension):
+    // When the status changes, we may have checked out a new branch (e.g., via the command-line and not via the extension) or changed repositories:
     this.props.model.statusChanged.connect(this._syncState, this);
   }
 
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -72,8 +72,16 @@ export class BranchMenu extends React.Component<
    */
   constructor(props: IBranchMenuProps) {
     super(props);
-    this.props.model.repositoryChanged.connect(this._onRepositoryChange, this);
-    this.props.model.headChanged.connect(this._onHeadChange, this);
+
+    // When the repository changes, we're likely to have a new set of branches:
+    this.props.model.repositoryChanged.connect(this._syncState, this);
+
+    // When the HEAD changes, decent probability that we've switched branches:
+    this.props.model.headChanged.connect(this._syncState, this);
+
+    // When the status changes, we may have checked out a new branch (e.g., via the command-line and not via the extension):
+    this.props.model.statusChanged.connect(this._syncState, this);
+
     this.state = {
       filter: '',
       branchDialog: false,
@@ -167,20 +175,6 @@ export class BranchMenu extends React.Component<
         {branch.name}
       </ListItem>
     );
-  };
-
-  /**
-   * Callback invoked upon a change to the repository path.
-   */
-  private _onRepositoryChange = () => {
-    this._syncState();
-  };
-
-  /**
-   * Callback invoked upon a change to the current HEAD.
-   */
-  private _onHeadChange = () => {
-    this._syncState();
   };
 
   /**
--- a/src/components/GitPanel.tsx
+++ b/src/components/GitPanel.tsx
@@ -192,7 +192,12 @@ export class GitPanel extends React.Component<
     }
   };
 
-  render() {
+  /**
+   * Renders the component.
+   *
+   * @returns React element
+   */
+  render(): React.ReactElement {
     return (
       <div className={panelWrapperClass}>
         {this._renderToolbar()}
@@ -204,9 +209,9 @@ export class GitPanel extends React.Component<
   /**
    * Renders a toolbar.
    *
-   * @returns fragment
+   * @returns React element
    */
-  private _renderToolbar() {
+  private _renderToolbar(): React.ReactElement {
     const disableBranching = Boolean(
       this.props.settings.composite['disableBranchWithChanges'] &&
         ((this.state.unstagedFiles && this.state.unstagedFiles.length) ||
@@ -224,9 +229,9 @@ export class GitPanel extends React.Component<
   /**
    * Renders the main panel.
    *
-   * @returns fragment
+   * @returns React element
    */
-  private _renderMain() {
+  private _renderMain(): React.ReactElement {
     if (this.state.inGitRepository) {
       return (
         <React.Fragment>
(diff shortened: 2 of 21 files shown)


telamonian:
Whelp, that improved my mood. Athan, it's great. Love the branch filter box.

I'll test this out

src/components/BranchMenu.tsx
```diff
@@ -0,0 +1,315 @@
...
+        .catch(onError);
+    }
+
+    /**
+     * Callback invoked upon promise resolution.
+     *
+     * @private
+     * @param result - result
+     */
+    function onResolve(result: any) {
```
saulshanabrook: Ha you really don't like anonymous functions do you :)

src/components/GitPanel.tsx
```diff
@@ -198,23 +195,101 @@ export class GitPanel extends React.Component<
...
-          pastCommits={this.state.pastCommits}
-          model={this.props.model}
-          renderMime={this.props.renderMime}
-        />
+  /**
+   * Renders a toolbar.
+   *
+   * @returns fragment
+   */
+  private _renderToolbar = () => {
```
saulshanabrook: Just curious, why do you prefer these as anonymous functions instead of methods?

fcollonval:
Thx @kgryte for this great work. I have two questions looking at the screenshots:

- Should the branch list be displayed on top of the change/history/commit components? When the list got expended, all other panels are compacted done and are not really nice. Do you know what happen in the Github desktop app?
- The panel for the repository path is large. I would rather stick to the proposal of #515 - maybe by displaying the path instead of the directory name. It will be closer to the filebrowser look and will take less vertical space. What do you think?

src/components/BranchMenu.tsx
kgryte: @saulshanabrook The standard reason for avoiding anonymous functions (including anonymous function expressions) is for debugging. For example, for callbacks provided to `Promise` methods, stack traces will include the function/method name when the function is named, thus allowing easier mapping of the problematic function to its source implementation (e.g., rather than searching for a line number, one can use `grep`). While benefits may be limited for one-off functions, the benefits increase for functions/methods/design patterns reused across a codebase. 

While named function expression support has improved in recent ECMAScript editions, historically, using them was not "best" practice, due to debugging and browser quirks (e.g., IE).

Lastly, moving the function body to a named function more clearly separates the usage of the function from its implementation and avoids common bracket matching issues (although this is obviated by modern IDE features such as automatic bracket pairing and highlighting, etc). Far too often, developers rely on nested inline anonymous functions which is generally bad practice because often the enclosing scopes include unnecessary identifiers and, without compiler tricks, results in unnecessary stack allocation. Better practice is using named functions at the highest possible scope to both limit a function's scope to what is absolutely necessary (e.g., only those parent scope identifiers which are needed by the function; commonly, this is the top-most module scope) and ensure that JS engines can minimize function creation and GC.

saulshanabrook: Thanks, looking more into this I didn't realize how many ways there were to declare functions in JS...

This is rather surprising to me, coming from a Python world, but it looks like if you set a variable to an arrow function it will get its name from the variable name (https://stackoverflow.com/a/37488652). 

> Far too often, developers rely on nested inline anonymous functions which is generally bad practice because often the enclosing scopes include unnecessary identifiers and, without compiler tricks, results in unnecessary stack allocation. Better practice is using named functions at the highest possible scope to both limit a function's scope to what is absolutely necessary (e.g., only those parent scope identifiers which are needed by the function; commonly, this is the top-most module scope)

Is there a difference between arrow function scoping and regular function scoping? I know that arrow functions don't have a new `this` so you don't have to have the `self = this` line with them.

src/components/GitPanel.tsx
saulshanabrook: Sorry I guess these are called "arrow functions" not anonymous functions.

kgryte: @saulshanabrook I don't actually prefer using anonymous arrow function expression assignments. I think I was initially following precedent elsewhere in the codebase (when in Rome...), but, upon reflection, I realized that their usage obviates the need for explicit `this` binding for event handlers. I went ahead and changed all methods which automatically have the correct `this` context, but kept the arrow function expressions for those situations where we want the `this` context to refer to the class instance, which happens to exclusively be for DOM event handlers.

I suppose an additional concern for not originally doing this was ensuring that all methods "looked" consistent. One subtlety by making this distinction is that, for those new to the code base, not necessarily obvious why one "style" is chosen over the other, unless readers know how `this` binding works. We could use `function#bind`, but this has been problematic historically due to perf and browser quirks.

saulshanabrook: Thanks again. It might be a good idea to make a JupyterLab wide style for these types of choices, since the distinctions seem rather subtle (hint: I don't get it).

src/components/BranchMenu.tsx
kgryte: @saulshanabrook Re: anonymous function expression assignments. Correct, but this is a recent feature stemming from ES6/ES2015. Historically, this was not the case. In stack traces, the function would be marked as "anonymous" despite assignment to a variable.

Re: scoping. None beside lexical `this`. Scoping still works the same. A nested arrow function will still be able to access variables from the parent scope.

src/components/GitPanel.tsx
kgryte: @saulshanabrook Re: arrow/anonymous functions. Actually, they were both. :) 

Re: styles. Well, if it were up to me... :) ...I'd prohibit all anonymous functions, all arrow functions, and all function expressions, and only allow function declarations. In other JS projects, we've configured ESLint to enforce such prohibitions. However, in today's JS and especially for Python devs who love lambdas, instituting this sort of change in a project such as JupyterLab is likely a nonstarter.

kgryte:
@fcollonval 

> Should the branch list be displayed on top of the change/history/commit components? When the list got expanded, all other panels are compacted down and are not really nice.

Thanks for catching this. I tweaked the styles to ensure that the panels have a minimum height. I updated the Gifs "basic usage" in (1) and (2) to show the new behavior. Now when the branch dropdown is open, the file list and history are not squished to `0` height.

The main design inspiration here for expanding menus within a single panel is the design pattern commonly used in common photo editing software such as Adobe Photoshop, Gimp, Inkscape, and Graphic. All the aforementioned use some form of collapsible menus and scrolling within a single panel.

One of the main advantages of this design pattern is the ability to have multiple menus expanded/open while still being able to interact with the main content (in photo editing software, that would be the canvas; in JLab, that would be a notebook or other document).

If we opted for overlays and/or dialogs, users would not be able to achieve a similar UX. 

> The component for the repository path is large.

I reduced the height of these menu buttons, along with the tab height (per @telamonian's suggestions).

The reason why I opted for a similar menu button design as the branch menu is design consistency. While #515 opts for a dialog, we could achieve something similar via an in-panel menu akin to the branch menu (including filtering and "pinning"). I've already stubbed this in-panel menu in this PR.

Once this PR is accepted and merged, I'd be willing to help work on this, as this will facilitate, which #515 already proposes, working with multiple Git repositories within JLab.

kgryte:
@telamonian I went ahead and implemented branching from any branch. Here is a demo:

git_create_branch_from_any_base

fcollonval:
Thanks for the detailed answer @kgryte

The ability to see the branch and the history is really nice.

Following your inspiration for the layout, should we open an issue to
discuss having something similar for the status/commit panel and the
history one instead of using tabs?

Regarding the path component, I don't think having something similar to the
branches is a good idea for the two following reasons:
* No other UI elements display information that will be unchanged when
switching between repository (eg there is no interest of displaying a file
browser and the git history at the same time as the latter depend of the
choice done in the former).
* We will end up reimplementing a feature available in the core with other
usage and/or style. As jupyter is an open source community, it sounds more
appropriate to reuse and improve the existing code. In this particular
case, this could be bringing a filter feature to the file browser (for
reference https://github.com/jupyterlab/jupyterlab/issues/818).

So this brings me to a request, could you restrict this PR to the branches
features and open a separate one to address/discuss the path component?

src/components/NewBranchDialog.tsx
```diff
@@ -0,0 +1,440 @@
+import * as React from 'react';
+import { classes } from 'typestyle';
+import List from '@material-ui/core/List';
+import ListItem from '@material-ui/core/ListItem';
+import Dialog from '@material-ui/core/Dialog';
+import DialogActions from '@material-ui/core/DialogActions';
```
fcollonval: Using dialog component from third-party library breaks the homogeneity of the code. But in the other hand this eases code development. @telamonian what is your opinion on this?

Note: personally, if the dialog look is similar to the default JupyterLab dialog, I'm fine with it.

src/components/NewBranchDialog.tsx
```diff
@@ -0,0 +1,440 @@
...
+     *
+     * @private
+     * @param event - event object
+     */
+    function onClick() {
+      self.setState({
+        base: branch
+      });
+    }
+  }
```
fcollonval: Can be simplified as in the branch menu.

src/components/NewBranchDialog.tsx
```diff
@@ -0,0 +1,440 @@
...
+    // Create the branch:
+    this._createBranch(branch);
+  };
+
+  /**
+   * Creates a new branch.
+   *
+   * @param branch - branch name
+   */
+  private _createBranch(branch: string): void {
```
fcollonval: Can be simplified by using a asynchronous method as for the method in the branch menu.

src/components/NewBranchDialog.tsx
```diff
@@ -0,0 +1,440 @@
...
+
+  /**
+   * Adds model listeners.
+   */
+  private _addListeners(): void {
+    // When the repository changes, we're likely to have a new set of branches:
+    this.props.model.repositoryChanged.connect(this._syncState, this);
+
+    // When the HEAD changes, decent probability that we've switched branches:
+    this.props.model.headChanged.connect(this._syncState, this);
```
fcollonval: Only this one should appear

src/components/NewBranchDialog.tsx
```diff
@@ -0,0 +1,440 @@
...
+    // When the status changes, we may have checked out a new branch (e.g., via the command-line and not via the extension):
+    this.props.model.statusChanged.connect(this._syncState, this);
+  }
+
+  /**
+   * Removes model listeners.
+   */
+  private _removeListeners(): void {
+    this.props.model.repositoryChanged.disconnect(this._syncState, this);
+    this.props.model.headChanged.disconnect(this._syncState, this);
```
fcollonval: Only this one should appear

src/components/NewBranchDialog.tsx
```diff
@@ -0,0 +1,440 @@
...
+   * Returns a React component for rendering a branch menu.
+   *
+   * @param props - component properties
+   * @returns React component
+   */
+  constructor(props: INewBranchDialogProps) {
+    super(props);
+
+    const repo = this.props.model.pathRepository;
+    this._addListeners();
```
fcollonval: Should be moved in `ComponentWillMount` to avoid state update before rendering

src/components/Toolbar.tsx
```diff
@@ -0,0 +1,371 @@
...
+   * Returns a React component for rendering a panel toolbar.
+   *
+   * @param props - component properties
+   * @returns React component
+   */
+  constructor(props: IToolbarProps) {
+    super(props);
+
+    const repo = this.props.model.pathRepository;
+    this._addListeners();
```
fcollonval: Should be moved in `ComponentWillMount`

src/components/Toolbar.tsx
```diff
@@ -0,0 +1,371 @@
...
+              'jp-Icon-16'
+            )}
+          />
+        </button>
+        {this.state.branchMenu ? (
+          <BranchMenu
+            model={this.props.model}
+            branching={this.props.branching}
+          />
+        ) : null}
```
fcollonval: Could be simplified
```ts
        {this.state.branchMenu && (
          <BranchMenu
            model={this.props.model}
            branching={this.props.branching}
          />
        )}
```

src/components/Toolbar.tsx
```diff
@@ -0,0 +1,371 @@
...
+   */
+  private _addListeners(): void {
+    // When the repository changes, we're likely to have a new set of branches:
+    this.props.model.repositoryChanged.connect(this._syncState, this);
+
+    // When the HEAD changes, decent probability that we've switched branches:
+    this.props.model.headChanged.connect(this._syncState, this);
+
+    // When the status changes, we may have checked out a new branch (e.g., via the command-line and not via the extension):
+    this.props.model.statusChanged.connect(this._syncState, this);
```
fcollonval: `statusChanged` should not be listened by this component

src/components/Toolbar.tsx
```diff
@@ -0,0 +1,371 @@
...
+  }
+
+  /**
+   * Removes model listeners.
+   */
+  private _removeListeners(): void {
+    this.props.model.repositoryChanged.disconnect(this._syncState, this);
+    this.props.model.headChanged.disconnect(this._syncState, this);
+    this.props.model.statusChanged.disconnect(this._syncState, this);
```
fcollonval: `statusChanged` should not be listened by this component

src/components/BranchMenu.tsx
```diff
@@ -0,0 +1,315 @@
+import * as React from 'react';
+import { classes } from 'typestyle';
+import List from '@material-ui/core/List';
+import ListItem from '@material-ui/core/ListItem';
+import ClearIcon from '@material-ui/icons/Clear';
```
fcollonval: Is it ok to import icons from third-party library or is it better to ship all icons except those shared with JupyterLab? @telamonian what do you think?

src/style/NewBranchDialog.ts
```diff
@@ -0,0 +1,246 @@
...
+});
+
+export const listItemIconClass = style({
+  width: '16px',
+  height: '16px',
+
+  /* top | right | bottom | left */
+  margin: 'auto 8px auto 0',
+
+  backgroundImage: 'var(--jp-icon-git-branch-light-theme)',
```
fcollonval: We should prefer the svg injection method for new icon.

See #515 for an example on how to add and use a new one.

src/style/Toolbar.ts
```diff
@@ -0,0 +1,165 @@
...
+
+export const repoIconClass = style({
+  backgroundImage: 'var(--jp-icon-git-repo)',
+  backgroundSize: '16px',
+  backgroundRepeat: 'no-repeat',
+  backgroundPosition: 'center'
+});
+
+export const branchIconClass = style({
+  backgroundImage: 'var(--jp-icon-git-branch)',
```
fcollonval: We should prefer the svg injection method for new icon.

See #515 for an example on how to add and use a new one.

src/style/Toolbar.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  background: 'var(--jp-layout-color1)',
+  backgroundImage: 'var(--jp-icon-refresh)',
+  backgroundSize: '16px',
+  backgroundRepeat: 'no-repeat',
+  backgroundPosition: 'center'
+});
+
+export const repoIconClass = style({
+  backgroundImage: 'var(--jp-icon-git-repo)',
```
fcollonval: We should prefer the svg injection method for new icon.

See #515 for an example on how to add and use a new one.

src/components/BranchMenu.tsx
```diff
@@ -0,0 +1,315 @@
...
+    }
+    return (
+      <ListItem
+        button
+        title={`Switch to branch: ${branch.name}`}
+        className={classes(
+          listItemClass,
+          branch.name === this.state.current ? activeListItemClass : null
+        )}
+        key={idx}
```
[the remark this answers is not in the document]
kgryte: Updated. Thanks for catching!

src/components/GitPanel.tsx
```diff
@@ -33,6 +37,7 @@ export interface IGitSessionNodeState {
   untrackedFiles: Git.IStatusFileResult[];
 
   isHistoryVisible: boolean;
+  tab: number;
```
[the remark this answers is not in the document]
kgryte: Done. Thanks for noticing this!

src/components/BranchMenu.tsx
```diff
@@ -0,0 +1,315 @@
...
+        <span className={listItemIconClass} />
+        {branch.name}
+      </ListItem>
+    );
+  }
+
+  /**
+   * Adds model listeners.
+   */
+  private _addListeners(): void {
```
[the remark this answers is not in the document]
kgryte: @fcollonval Sorry, but I don't follow. I included comments explaining why the branch menu listens for each of the three events.

Listening for a `head` change is not enough because, if I create a new branch from the current branch and do so outside of JupyterLab, the `HEAD` does not change, and, thus, the branch change does not get reflected in the UI. The signal which does get emitted when the branch changes, but the `HEAD` does not change, is the `status` signal.

More of an edge case, but a repository change does not necessarily entail a `HEAD` change either. If I had two clones of the same repository sitting side-by-side, I could switch repositories, but still point to the same hash.

Yes, this does mean that multiple event listeners could be triggered, potentially even all 3; however, per the implementation proposed in this PR, this should not trigger a re-render unless model state changes in the interim. In which case, the perf cost is likely to be negligible, and I think acceptable in order to account for edge cases, particularly when mixing Git management inside and outside of JupyterLab (which is something I do regularly).

IMO, I don't think we should have a separate poll for refreshing a list of branches, as multiple polls at different intervals leads to synchronicity issues. And further, even if we polled `refreshBranch`, I don't think that that method should emit a `headChanged` signal, particularly if the `HEAD` has not changed.

Lastly, this PR intentionally eschews the mixed data flow model, as mentioned in the OP and as was used previously, where child components rely on both unidirectional data flow (via properties) **and** querying the underlying extension model (e.g., using getters). Such a mixed model leads to synchronicity bugs, as encountered during development.

Instead, the implementation proposed in this PR relies exclusively on the model (and not parent components) as the source of truth. Moving toward a poll method would most likely entail reverting back to a mixed data flow (e.g., see the `HistorySideBar` component) where, presuming `refreshBranch` does not emit a `HEAD` change event, we'd have to set a `branches` property on the branch menu component to ensure that the UI does not drift from the underlying Git repository.

Furthermore, again presuming that we'd not move toward relying on polling and emitting `HEAD` change events, the fact that changing a repository currently invokes `refresh` (which invokes `refreshBranch`) is only relevant to the implementation currently proposed by this PR insofar as the model's `_branches` property is updated, which the branch menu subsequently accesses after receiving one (or more) signal(s) that the list of branches has potentially changed.

In short, if potentially redundant signals is an overarching concern, I'd be open to throttling, but, in general, I am not particularly inclined to move toward polling or removing listeners which would potentially leave the UI in a state which deviates from the underlying Git repository.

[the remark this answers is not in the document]
kgryte: One further addendum: looking back through the `model` implementation, maybe what you meant is that `refresh` and thus a `status` change signal is emitted upon changing the repository. In which case, you're right. I will remove the `repositoryChanged` listeners.

src/components/BranchMenu.tsx
```diff
@@ -0,0 +1,315 @@
...
+    this.props.model.headChanged.connect(this._syncState, this);
+
+    // When the status changes, we may have checked out a new branch (e.g., via the command-line and not via the extension):
+    this.props.model.statusChanged.connect(this._syncState, this);
+  }
+
+  /**
+   * Removes model listeners.
+   */
+  private _removeListeners(): void {
```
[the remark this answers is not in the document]
kgryte: Done.

src/components/BranchMenu.tsx
```diff
@@ -0,0 +1,315 @@
...
+   * Returns a React component for rendering a branch menu.
+   *
+   * @param props - component properties
+   * @returns React component
+   */
+  constructor(props: IBranchMenuProps) {
+    super(props);
+
+    const repo = this.props.model.pathRepository;
+    this._addListeners();
```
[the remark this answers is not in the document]
kgryte: Changed. I am guessing that the same could be said of [`GitPanel`](https://github.com/jupyterlab/jupyterlab-git/blob/576c4448f2ce19c63604a949e343da498cf81858/src/components/GitPanel.tsx#L64) from which I based placing listener binding in the constructor?

src/components/NewBranchDialog.tsx
kgryte: In general, I recommend moving away from JupyterLab dialogs. If anything, the use of JupyterLab [dialogs](https://github.com/jupyterlab/jupyterlab-git/blob/576c4448f2ce19c63604a949e343da498cf81858/src/widgets/gitPushPull.ts#L14) breaks homogeneity of the extension code (i.e., React components), not the other way around.

Upon initially submitting the PR, I ran the use of Material UI dialogs past @telamonian and he thought this to be fine.

src/components/NewBranchDialog.tsx
kgryte: Removed the `repositoryChanged` listener, but kept the `status` listener, per discussion above.

src/components/NewBranchDialog.tsx
kgryte: Changed. Lifecycle method: `componentDidMount`.

src/components/Toolbar.tsx
kgryte: Changed. Lifecycle method: `componentDidMount`.

src/components/Toolbar.tsx
kgryte: Removed `repositoryChanged` listener, but kept `statusChanged` listener. In short, a `headChanged` signal is not guaranteed when switching branches (e.g., via the command-line outside of the extension) and `statusChanged` catches repository changes and status changes occurring outside the extension. Yes, this might mean some "false positive" signals (e.g., checking out a previous commit triggering a `statusChanged` signal), but these should not trigger a re-render as component state will remain the same.

src/components/Toolbar.tsx
kgryte: Updated in light of the above.

src/style/NewBranchDialog.ts
kgryte: I would argue against this. SVG injection just results in more DOM nodes and worse application performance. Better to use CSS image assets and allow the browser to leverage caching.

src/style/Toolbar.ts
kgryte: Same comment as [here](https://github.com/jupyterlab/jupyterlab-git/pull/518#discussion_r370909318).

src/style/Toolbar.ts
kgryte: Same comment as [here](https://github.com/jupyterlab/jupyterlab-git/pull/518#discussion_r370909318).

src/components/BranchMenu.tsx
kgryte: We could eventually move to using the [`jp-CloseIcon`](https://github.com/jupyterlab/jupyterlab/blob/3cb64ca6885358e7f9623f518518bb3f9de017bc/packages/ui-components/style/icons.css#L318) once it is ready for use. For now, I don't think importing an icon should be a blocker. 

In line with [elsewhere](https://github.com/jupyterlab/jupyterlab-git/pull/518#discussion_r370909318), I could, however, just get the SVG and leverage CSS. Here, I opted for convenience.

src/components/NewBranchDialog.tsx
kgryte: Removed `repositoryChanged`, but kept the `statusChanged` listener.

src/components/BranchMenu.tsx
```diff
@@ -0,0 +1,315 @@
...
+            />
+            {this.state.filter ? (
+              <button className={filterClearClass}>
+                <ClearIcon
+                  titleAccess="Clear the current filter"
+                  fontSize="small"
+                  onClick={this._resetFilter}
+                />
+              </button>
+            ) : null}
```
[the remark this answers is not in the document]
kgryte: Personally, I don't see this as a simplification. While using `||` is relatively common in JavaScript-land when setting default values, using `&&` in this way is not. While `||` lends itself to saying `this OR that`, permitting a mental model other than truth tables, `&&` translates to `this AND that` which people commonly (although [mistakenly](https://www.ecma-international.org/publications/files/ECMA-ST/Ecma-262.pdf)) assume will be a `boolean`.

Anecdotally, that this erroneous mental model is common is supported by the fact that I've both seen and experienced interviews in which one is, as something of a trick question to distinguish language "aficionados", asked the result type of `'' && null`.

And while Python behaves similarly to JavaScript,

```python
>>> '' and false
''
```

Julia, e.g., goes so far as to prohibit non-boolean values in boolean contexts.

```julia
julia> "" && false
ERROR: TypeError: non-boolean (String) used in boolean context
```

In short, I personally tend to favor being explicit rather than relying on language quirks that are neither widely ubiquitous (despite being shown in the React documentation) as convention nor straightforward to immediately reason about.

src/components/Toolbar.tsx
kgryte: Same as [here](https://github.com/jupyterlab/jupyterlab-git/pull/518#discussion_r370911399).

src/components/NewBranchDialog.tsx
telamonian: It's not that I wouldn't prefer a home grown, tastefully themed solution, but there's some problems:

- The Dialog box from @jupyterlab/apputils just isn't that flexible. It hard to significantly customize or use in complex scenarios.

- The apputils Dialog also isn't React based. The way it implements events (like what happens when you click a button) aren't naturally compatible with standard React patterns.

Because of the above, I once wasted an entire evening trying to implement a custom form using the Dialog from jlab core. I wanted to open the dialog box using one React component that was nested within a tree, then update the top-level component in the tree based on the user's input. It was not working out. On the other hand, once I decided to switch to the Dialog from material-ui, I had it solved within about 15 min.

I definitely see your point Frederic about homogeneity, and about the value of keeping things simple and consistent. It is true that material-ui brings along its own complex config system (with a rather foreign set of conventions). However, I don't think it makes sense for us to favor any particular approach based on a primary argument of homogeneity, or "because it's provided by jlab core". Instead, we should ask ourselves which component will allow us to build the most kick-ass and easily maintained `jupyterlab-git` that we can manage. The Dialog component from material-ui is simply much better than the Dialog from jlab core, even more so since we're primarily working within React here.

So based on my own experience (and I suppose I am still a bit salty about the apputils Dialog wasting my time), I think that Athan absolutely made the right call here to use material-ui. My only caveat is the same as Frederic's: the theming overall needs to mesh with that of the rest of jlab (though it certainly doesn't have to perfectly match).

On the horizon of a month or two from now, a React-based dialog box is definitely on my list of things to add to @jupyterlab/ui-components. Once that exists, maybe it will be appropriate to revisit this discussion.

src/components/BranchMenu.tsx
```diff
@@ -0,0 +1,315 @@
...
+    /**
+     * Callback invoked upon encountering an error.
+     *
+     * @private
+     * @param err - error
+     */
+    function onError(err: any) {
+      showErrorMessage('Error switching branch', err.message);
+    }
+  }
```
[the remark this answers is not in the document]
kgryte: Yes, we could use `async`/`await` here, although, in application and library code, I typically do not.

The primary reason between that wrapping an `async` function in a `try`/`catch` block further conflates operational and programmer errors.

However, if using `async`/`await` is the desired style, I am willing to update accordingly.

src/components/NewBranchDialog.tsx
kgryte: To your point @telamonian, I tried to style the dialog such that it does mesh, using theme classes, etc. So we should be good on that front.

src/components/NewBranchDialog.tsx
kgryte: Similar reply as [here](https://github.com/jupyterlab/jupyterlab-git/pull/518#discussion_r370911838).

src/components/NewBranchDialog.tsx
kgryte: Similar reply as [here](https://github.com/jupyterlab/jupyterlab-git/pull/518#discussion_r370911838).

src/components/BranchMenu.tsx
telamonian: Hmm, good question. For the purposes of this PR, and probably in general, I think extension devs using a handful of icons from a third party package is fine (I know that some people are *very* attached to font-awesome, for example).

I do think you're right that the best practice would be to ship the icons ourselves. There's a boatload of icon functionality (theming, replacement, reuse, etc) that in general just won't work with 3rd party icons. I'm actually hard at work right now putting the finishing touches on `LabIcon` (formerly `JLIcon`), which will be the new icon system for jlab 2.0. It should be finished in the next day or two (or rather, it better well be, or else I'll end up blocking the 2.0rc release :sweat_smile:). Once that's done with I'll start working on adding to the [jupyterlab-git-for-jlab-2.0](https://github.com/jupyterlab/jupyterlab-git/pull/520) PR that @blink1073 started, with the goal of moving all of our icons over to `LabIcon` (including finding/making equivalents of the material-ui icons).

src/components/NewBranchDialog.tsx
```diff
@@ -0,0 +1,440 @@
...
+  branches: Git.IBranch[];
+}
+
+/**
+ * React component for rendering a dialog to create a new branch.
+ */
+export class NewBranchDialog extends React.Component<
+  INewBranchDialogProps,
+  INewBranchDialogState
+> {
```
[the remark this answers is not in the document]
kgryte: I agree with the sentiment to keep things DRY; however, I don't believe composition is warranted here, given differing rendering logic which would need to be abstracted over.

Were we to need yet another branch list and accompanying filter box somewhere else, perhaps then we could consider creating an abstraction. At this point, I consider that premature, particularly as doing so at this time does not do much in the way of reducing LOC or complexity.

src/components/NewBranchDialog.tsx
kgryte: My motivation here was to use named functions rather than inline arrow functions for better debugging (e.g., via stack traces) and to elevate callback creation outside of rendering logic. 

Additionally, I find that, personally, when callbacks are [inlined](https://github.com/jupyterlab/jupyterlab-git/blob/576c4448f2ce19c63604a949e343da498cf81858/src/components/SinglePastCommitInfo.tsx#L230) within JSX, it convolves declarative rendering (this is what it should look like) with imperative programming logic (this is what it should do). Accordingly, my intuition is to elevate callback factories to the component prototype and use closures, with the added benefit of being able to provide JSDoc annotations for all functions.

Perhaps, as a matter of taste, this approach is a bit verbose, but IMO it strikes the right balance wrt separation of concerns and, in general, "debuggability".

src/components/BranchMenu.tsx
kgryte: Awesome! Thanks for the update @telamonian!

src/style/NewBranchDialog.ts
telamonian: > SVG injection just results in more DOM nodes and worse application performance

@kgryte Well, that opinion is worrisome. Since 99% of the icons in core are now rendered as svg nodes, and since I am 100% responsible for that change, I could use a bit more guidance as to your strong stance on the issue. When I first switched the icons over to the icons-as-svg-nodes pattern I did do a bunch of profiling. I did not see any change in performance that was significant relative to the noise. 

Since I know that you know more than me about the deep mysteries of JS, I am a bit concerned. How grounded in your own experience is your opinion on the matter? Is the icons-as-svg-nodes pattern an inherently Bad Idea, or are you just operating on the general principle that more nodes => more overhead => more bad?

From my viewpoint, the big advantage that the icons-as-svg-nodes pattern has over the older icons-as-css-background pattern is that svg nodes are reachable by css. Given the right setup, this allows icons to be dynamically repainted when the theme changes with the actual colors provided by the particular theme. As far as I've been able to dig up, there doesn't seem to be another good way of dynamically and flexibly theming icons

kgryte: @telamonian You are right. When it comes to theming as is wanted in JupyterLab, you don't have much of a choice apart from loading different theme assets via CSS (as is currently down now in this PR for dark and light modes).

However, one needs to be careful in ensuring that the use of SVG icons does not cause the DOM tree to explode. For example, were we to use SVG injection here (for branch icons) and we had `1000` branches, assuming, for a moment, no virtual scrolling/windowing, then we'd inject 1000 SVG icons into the DOM. Upon doing so, performance, in particular DOM tree traversal, will be adversely affected based on the principle that, yes, more node => more overhead => more bad.

That this principle affects JupyterLab more generally is supported by the fact that JupyterLab exhibits poor performance for large notebooks (e.g., due to the number of DOM nodes associated with each input and output cell).

In this case, particularly when an image is going to be repeated many times within a document, leveraging CSS is better (perf-wise), despite being more limited/restrictive for theming.

For what you are doing with JLab where you can expect, e.g., toolbar icons to render once, or at most a handful times, in the application, using SVG icons is fine. Issues arise, however, when SVG nodes proliferate within an HTML document--a point to which the web data visualization community can readily attest.

telamonian: Hmm, interesting, and something I definitely need to keep an eye out for. Thanks for the info. From what you say, I think the only likely pain point in core is then going to be the filebrowser. I know it's fine up to at least ~2000 items, but I can see how it might spiral out past that. Some basic pagination is probably in order.

There's also some more general possible optimizations that I've been putting off until I had reason to no longer think they'd be premature. There's an interesting one using `xmlns:xlink` that can supposedly save cycles when rendering an icon in multiple places.

src/components/BranchMenu.tsx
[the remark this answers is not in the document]
telamonian: @kgryte I'm not as sure about the wider JS world, but I do know that taking advantage of `&&` short-circuiting is part of the idiom of Python (it's discussed extensively even in introductory tutorials), it's part of the idiom of React, and it's also used frequently in jlab core. Whether you change to using it here or not is a minor stylistic (and maybe performance?) niggle, but in general I don't think we have any reason to shy away from using the short-circuiting syntax/semantics.

I think you can reason about the behavior if you remember two things:

- Most (all?) values in Python/JS can participate in logical operations, but they won't be automagically converted to booleans

- The point of short-circuiting is to not evaluate the second clause of a logical operator if the answer is known after evaluating the first

> asked the result type of `'' && null`

From the above, you can work out that since `''` is falsey, the second clause will be ignored and the return value will just be `''`. 

Another tricky question would be "what's the return value of `'' || null`?". I'll try to prove my point by working it out (since I don't actually know the answer): `''` is falsey, so no short circuiting happens here, the second clause will be eval-ed, and the return will be `null`.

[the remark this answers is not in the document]
telamonian: and I was right! yaaay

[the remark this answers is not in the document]
kgryte: @telamonian I think you not knowing the answer *a priori* for `'' || null` is part of my point. ;)

But recognized that logical short-circuiting is a common Python idiom and found in React. Just not convinced that because one can, and does, that one should.

Using logical short-circuiting does **not** provide a performance benefit when compared to using a ternary operator in this situation. For reference, see sections [12.13.3](https://www.ecma-international.org/publications/files/ECMA-ST/Ecma-262.pdf) and [12.14.3](https://www.ecma-international.org/publications/files/ECMA-ST/Ecma-262.pdf).

src/style/NewBranchDialog.ts
kgryte: @telamonian Yeah, I think that windowing/virtual scrolling could be an effective solution in the filebrowser to minimize the number of DOM nodes.

telamonian:
@kgryte, from an unwieldy long comment thread on a review comment above:

> @telamonian I think you not knowing the answer *a priori* for `'' || null` is part of my point. ;)
>
> But recognized that logical short-circuiting is a common Python idiom and found in React. Just not convinced that because one can, and does, that one should.
>
>Using logical short-circuiting does **not** provide a performance benefit when compared to using a ternary operator in this situation. For reference, see sections [12.13.3](https://www.ecma-international.org/publications/files/ECMA-ST/Ecma-262.pdf) and [12.14.3](https://www.ecma-international.org/publications/files/ECMA-ST/Ecma-262.pdf).

Gotta love a man who can quote the standard.

I did read through those sections, though, and it still kind of seems like short circuiting with `&&` **does** provide a performance benefit when compared to using a ternary operator. Comparing the steps [12.13.3.4](https://www.ecma-international.org/publications/files/ECMA-ST/Ecma-262.pdf) and [12.14.3.4](https://www.ecma-international.org/publications/files/ECMA-ST/Ecma-262.pdf), the standard makes it sound like there's one less round of evaluate-then-`GetValue` for the `false` branch in the short circuiting case. Or am I misinterpreting it?

telamonian:
More seriously, I'm reviewing this PR right now (jlab 2.0rc was finally released!), with the intention of getting it pulled in tomorrow or Friday. Does that sound good to you @fcollonval, or were there more changes you wanted to see?

kgryte:
@telamonian Re: performance. I went ahead and ran benchmarks against Node v12 which bundles a recent version of V8.

## Benchmarks

```javascript
var isnan = require( '@stdlib/math/base/assert/is-nan' );
var bench = require( '@stdlib/bench' );

var opts = {
	'iterations': 1e9,
	'repeats': 3
};

bench( 'ternary', opts, function benchmark( b ) {
	var r;
	var i;

	b.tic();
	for ( i = 0; i < b.iterations; i++ ) {
		r = (i&1) ? i+2 : 0;
		if ( r !== r ) {
			b.fail( 'something went wrong!' );
		}
	}
	b.toc();
	if ( isnan( r ) ) {
		b.fail( 'something went wrong!' );
	}
	b.pass( 'benchmark finished' );
	b.end();
});

bench( 'short-circuit', opts, function benchmark( b ) {
	var r;
	var i;

	b.tic();
	for ( i = 0; i < b.iterations; i++ ) {
		r = (i&1) && i+2;
		if ( r !== r ) {
			b.fail( 'something went wrong!' );
		}
	}
	b.toc();
	if ( isnan( r ) ) {
		b.fail( 'something went wrong!' );
	}
	b.pass( 'benchmark finished' );
	b.end();
});
```

The code being tested is logically equivalent:

-  if the last bit is "on", return `i+2`; otherwise, return `0`.

In the second benchmark, we follow the spec, where

-   if `i&1` is falsy, then we return the evaluated result (i.e., `0`).
-   if `i&1` is truthy, then we return the evaluated result of `i+2`.

The same logic applies to benchmark which uses a ternary operator.

## Results

System:

-   Macbook Pro
-   Node Version: `12.9.1`
-   V8 Version: `7.6.303.29`

```text
TAP version 13
# ternary
  ---
  iterations: 1000000000
  elapsed: 1.65652842
  rate: 603672106.0300312
  ...
ok 1 benchmark finished
# ternary
  ---
  iterations: 1000000000
  elapsed: 1.1751802040000001
  rate: 850933326.3071201
  ...
ok 2 benchmark finished
# ternary
  ---
  iterations: 1000000000
  elapsed: 1.190982218
  rate: 839643098.6847866
  ...
ok 3 benchmark finished
# short-circuit
  ---
  iterations: 1000000000
  elapsed: 4.296960323
  rate: 232722651.55611956
  ...
ok 4 benchmark finished
# short-circuit
  ---
  iterations: 1000000000
  elapsed: 3.491717165
  rate: 286392039.43083405
  ...
ok 5 benchmark finished
# short-circuit
  ---
  iterations: 1000000000
  elapsed: 3.5987020579999998
  rate: 277877963.74444956
  ...
ok 6 benchmark finished
#
1..6
# total 6
# pass  6
#
# ok
```

## Analysis

If we take the results at face value, then ternary operations are more performant (at least in V8). Whether these results hold across browser engines is not considered.

The bigger picture takeaway is that whether one uses ternary operators or short-circuiting is **very unlikely** to be a performance bottleneck. At this level of granularity, claiming performance benefits based on the spec is a fool's errand due to the nature of modern browser compiler technology.

Most of the time, we cannot really say with certainty which equivalent operations will be more performant across browsers or from one browser version to the next. While there are tried and true principles (e.g., avoid functional class APIs, such as `Array#forEach`, and write JS like you would C, such as vanilla `for` loops), anything at the level of ternary versus logical is unlikely to surface any universal truths, unfortunately.

For the curious, if we want to speculate why ternary operations are "faster" in V8, my guess is that ternary operations are simply more ubiquitous, and, thus, have been given greater performance optimization consideration by compiler engineers, which would be in line with their stated goal of optimizing for real-world use cases. But, admittedly, all this speculation and discussion leads us astray.

fcollonval:
@telamonian thanks for posting on this.
There is one change that should be done for me is the suppression of the button element for the repo menu and the right arrow. This gives the impression that some interaction is possible when it is not.

See L204 of components/Toolbar.tsx

Otherwise this is good for me.

style/variables.css
```diff
@@ -4,38 +4,43 @@
...
-  --jp-icon-git-clone: url('./images/git-clone-icon.svg');
-  --jp-git-diff-deleted-color: rgba(255, 0, 0, 0.2);
   --jp-git-diff-added-color: rgba(155, 185, 85, 0.2);
-  --jp-git-diff-output-color: rgba(0, 141, 255, 0.3);
+  --jp-git-diff-deleted-color: rgba(255, 0, 0, 0.2);
   --jp-git-diff-output-border-color: rgba(0, 141, 255, 0.7);
+  --jp-git-diff-output-color: rgba(0, 141, 255, 0.3);
+  --jp-icon-clear-white: url('./images/clear-white.svg');
+  --jp-icon-discard-file-selected: url('./images/discard-selected.svg');
+  --jp-icon-git-branch-light-theme: url('./images/git-branch.svg');
```
telamonian: get rid of this var once you've fixed up the dark theme stuff for the new branch dialog box

telamonian:
> The bigger picture takeaway is that whether one uses ternary operators or short-circuiting is **very unlikely** to be a performance bottleneck. At this level of granularity, claiming performance benefits based on the spec is a fool's errand due to the nature of modern browser compiler technology.

Well said, and the benchmarks provide a convincing object lesson. Thanks for doing those. Overall this has been a surprisingly instructive PR. It's been nice to see some typescript written with a different perspective/idiom from what I'm used to.

In the benchmarks, are the trivial conditionals:

```
if ( r !== r ) {
			b.fail( 'something went wrong!' );
		}
```

just there to ensure that the loop contents don't get optimized away?

src/components/BranchMenu.tsx
```diff
@@ -0,0 +1,317 @@
...
+   * Callback invoked when a component will no longer be mounted.
+   */
+  componentWillUnmount(): void {
+    this._removeListeners();
+  }
+
+  /**
+   * Renders the component.
+   *
+   * @returns fragment
```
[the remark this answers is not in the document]
kgryte: @telamonian Updated. Sorry for the confusion. I was using "fragment" loosely along the lines of how, in HTML templating, one commonly refers to small chunks of HTML as "fragments".

I've updated the various descriptions to be more precise.

src/components/GitPanel.tsx
```diff
@@ -198,23 +193,99 @@ export class GitPanel extends React.Component<
...
+        {this._renderToolbar()}
+        {this._renderMain()}
+      </div>
+    );
+  }
+
+  /**
+   * Renders a toolbar.
+   *
+   * @returns fragment
```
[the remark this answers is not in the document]
kgryte: Updated.

src/components/GitPanel.tsx
```diff
@@ -198,23 +193,99 @@ export class GitPanel extends React.Component<
...
+   */
+  private _renderTabs() {
+    return (
+      <Tabs
+        classes={{
+          root: tabsClass,
+          indicator: tabIndicatorClass
+        }}
+        value={this.state.tab}
+        aria-label="Panel tabs"
```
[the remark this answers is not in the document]
kgryte: Removed.

src/components/NewBranchDialog.tsx
```diff
@@ -0,0 +1,443 @@
...
+   */
+  render() {
+    return (
+      <Dialog
+        classes={{
+          paper: branchDialogClass
+        }}
+        open={this.props.open}
+        onClose={this._onClose}
+        aria-labelledby="new-branch-dialog"
```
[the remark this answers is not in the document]
kgryte: Removed.

src/components/NewBranchDialog.tsx
```diff
@@ -0,0 +1,443 @@
...
+  }
+
+  /**
+   * Callback invoked upon a change to the branch name input element.
+   *
+   * @param event - event object
+   */
+  private _onNameChange = (event: any): void => {
+    this.setState({
+      name: event.target.value
```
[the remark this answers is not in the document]
kgryte: @telamonian On my MacBook, I am unable to hold down a key to generate repeats. Even within this comment box on GitHub, I have to press each time I want to generate a key.

I do see a slight lag if I attempt to quickly type a bunch of random characters. However, for normal typing speeds, I do not see any noticeable difference compared to entering this comment on GitHub.

If I had to hazard a guess as to why the `name` input has a slightly longer lag than the filter box, it would be due to repaints and layouts. Namely, the `name` input node is slightly higher up the DOM tree and, thus, updates to it may result in larger tree updates and layout update surface area. But this is only a guess.

From reviewing the code, I don't see anything obvious that can be done to mitigate any lag, so going to mark this as resolved.

kgryte:
@telamonian Re: checks. That is correct. In this case, the goal would be to thwart "intelligent" compilers from skipping ahead to the last `i` and calling it a day. This sort of optimization is unlikely, as requires rather advanced speculation.

In general, however, when timing within loops, conditional checks are commonly necessary to prevent loop-invariant code motion, dead code elimination, and constant folding/propagation optimizations. It is also why an additional check outside of the loop is necessary in order to prevent the entire loop from being optimized away.

In this particular example, I used `NaN` checks based on the assumption that an optimizing compiler would be unlikely to reason about value ranges *a priori* and conclude that `NaN` is actually impossible. Based on the return value `r` being a `double`, a `NaN` check would be a valid test.

src/components/Toolbar.tsx
```diff
@@ -0,0 +1,371 @@
...
+      </div>
+    );
+  }
+
+  /**
+   * Renders a repository menu.
+   *
+   * @returns fragment
+   */
+  private _renderRepoMenu() {
```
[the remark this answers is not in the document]
kgryte: Disabled button and removed right arrow icon.

style/variables.css
kgryte: Removed.

kgryte:
@telamonian I have

-   updated the docstrings
-   removed the aria attributes
-   fixed the dark theme styling for the branch dialog
-   disabled the repo menu button and removed the arrow icon indicating that a button is functional

Screen Shot 2020-02-11 at 8 23 24 PM

I believe this resolves the remaining issues.

telamonian:
Yah, that's basically what I did (though sans import of `classes`)

kgryte:
@telamonian It is already imported: https://github.com/jupyterlab/jupyterlab-git/commit/cac88a50bdfded2c8305ad1e6f7cae04bf3bd838#diff-f3df9dfa36cc17388ea613f8e3916891L2

telamonian:
Other two commits? I have no idea what you're talking about

kgryte:
@telamonian Must have been dreaming. ;)

src/components/NewBranchDialog.tsx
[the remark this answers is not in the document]
telamonian: > On my MacBook, I am unable to hold down a key to generate repeats. Even within this comment box on GitHub, I have to press each time I want to generate a key.

@kgryte Ohhh yeah, that's right. 5 or 6 os versions ago when they added that "show accented/alt chars on key hold" feature I had to turn it off:

```
defaults write -g ApplePressAndHoldEnabled -bool false
```

Much nicer that way, and there are also some special options for increasing the repeat speed beyond what the system preferences UI allows.

Anyhow, this is something to be aware of, but isn't a showstopper, so I think it's fine for now

kgryte:
@telamonian Hurray! Thanks for reviewing!

PR #519, opened 2020-01-08
List all untracked files

fcollonval:
Explicitly list all untracked files instead of the containing folder if none of it's files are tracked.


kgryte:
LGTM.

Issue #526, opened 2020-01-17
Installation trouble


goofts:
jupyter serverextension list

goofts:
VM819:1 GET https://nbs.goofts.online/jupyter/git/server_root?1579253166187 404 (Not Found)

fcollonval:
Please have a look at the [Troubleshooting section](https://github.com/jupyterlab/jupyterlab-git#troubleshooting) and report here if you still have an error.

Note: I was able to reach your endpoint without error

```
{"server_root": "/***"}
```

@blink1073 @telamonian I think we have a security problem with this extension handlers... 😨 🚨 I should not have been able to get that answer. Trying https://nbs.goofts.online/jupyter/ redirect me correctly to the password page.

I am not too familiar with the security. Will using the `@tornado.web.authenticated` decorator on the handler solve this?

telamonian:
Yeah, my understanding is that you would at least need the token as part of the url in order to access anything from any endpoint in the jupyter notebook server, but I guess that's not the case...

PR #527, opened 2020-01-19
Add `authenticated` decorator on handler

fcollonval:
This fixes the security hole discovered thanks to #526 

When adding the decorator `@tornado.web.authenticated` to the handlers, they returned the following message instead of the valid answer:
```
{"message": "Forbidden", "reason": null}
```


telamonian:
This turned out to be a pain to rebase (due to the changes in #493 and others), so I just started a new nearly identical PR: #558

PR #539, opened 2020-02-25
Add support for diffing all files

### LINKED ISSUE #483: followup #406: diff all the plaintext things
ref https://github.com/jupyterlab/jupyterlab-git/pull/406#discussion_r354111456

#406 has been reviewed/approved, and once it's pulled in we'll be able to diff plaintext files. However, the UI/frontend only allows diffing a file if its extension is recognized. Ideally, we should be able to diff any plaintext-like file that `git diff` can handle.

One approach would be to allow the user to ask for a diff of any file on the frontend (via the right click menu or otherwise) and then sort out diff-able files from binary files on the backend. On the backend, we can detect binary (and therefore not diff-able) files either via git (its a bit tricky to get git to cough up that particular piece of info, but apparently it's doable), or by using the [same byte checking trick](https://stackoverflow.com/a/6134127/425458) that git itself uses

kgryte:
This PR

-   adds support for diffing all files by assuming `text/plain` by [default](https://github.com/jupyterlab/jupyterlab/blob/d3c1ffbcc85db1a85a8c3d733cd66438b302d812/packages/codemirror/src/mode.ts#L132).
-   resolves #483 by adding a method for checking whether Git handles a file as binary or text.

Currently, if the extension is unable to resolve a CodeMirror mode using the [filename alone](https://github.com/jupyterlab/jupyterlab-git/blob/ae3f2a6f1e76f5e57c6a203b988d9811bc5d2452/src/components/diff/PlainTextDiff.tsx#L148), the extension punts on allowing a user to view a diff.

This PR adds support for viewing diffs of dot files, such as `.gitignore`, `.npmignore`, and others, which do not lend themselves to CodeMirror mode resolution by filename.

Screen Shot 2020-02-24 at 6 39 53 PM


kgryte:
@telamonian Added support for checking if a file is binary or text.

kgryte:
Fixed the broken tests. Should be ready for review.

kgryte:
@telamonian Resolved merge conflicts stemming from #493.

kgryte:
@telamonian Awesome! Thanks!

PR #540, opened 2020-02-25
Document settings in the README

kgryte:
This PR

-   documents plugin settings, providing additional context and rationale.
-   reformats existing copy to ensure consistent with the rest of the README.
-   updates existing copy phrasing.


README.md
```diff
@@ -25,6 +25,15 @@ pip install --upgrade jupyterlab-git
 jupyter lab build
 ```
 
+## Settings
+
+Once installed, extension behavior can be modified via the following settings which can be set in JupyterLab's advanced settings editor:
+
+-   **disableBranchWithChanges**: disable all branch operations, such as creating a new branch or switching to a different branch, when there are changed/staged files. Setting this setting to `true` guards against overwriting and/or losing uncommitted changes.
```
fcollonval:
```suggestion
-   **disableBranchWithChanges**: disable all branch operations, such as creating a new branch or switching to a different branch, when there are changed/staged files. Setting this option to `true` guards against overwriting and/or losing uncommitted changes.
```

README.md
```diff
@@ -25,6 +25,15 @@ pip install --upgrade jupyterlab-git
...
+-   **historyCount**: number of commits shown in the history log, beginning with the most recent. Displaying a larger number of commits can lead to performance degradation, so use caution when modifying this setting.
+-   **refreshInterval**: number of milliseconds between polling the file system for changes. In order to ensure that the UI correctly displays the current repository status, the extension must poll the file system for changes. Longer polling times increase the likelihood that the UI does not reflect the current status; however, longer polling times also incur less performance overhead.
+-   **simpleStaging**: enable a simplified concept of staging. When this setting is `true`, all files with changes are automatically staged. When we develop in JupyterLab, we often only care about what files have changed (in the broadest sense) and don't need to distinguish between "tracked" and "untracked" files. Accordingly, this setting allows us to simplify the visual presentation of changes, which is especially useful for those less acquainted with Git.
```
mlucool: I have never loved the simple staging name. It focuses on one key difference, staging being shown, and not the concept - making it easier for those who little to no git experience.

mlucool: Maybe we should include another gif or second binder here? I'll also note the top gif appears to be out of date.

README.md
kgryte: Addressed.

README.md
kgryte: @mlucool I think there may some confusion here, and I apologize if this was not articulated or made clear. When reasoning about "simple Git" during implementation, a decision was made, if I recall correctly given that this work pre-dates my involvement, to, rather than have a single idea as to what a simplified Git UI should be, allow for various aspects of the current UI to be "simplified". By decomposing into separate settings (e.g., `disableBranchWithChanges`, `simpleStaging`, etc), users/orgs could then define what they consider to be "simple Git". This said, the decomposition of the "simple Git" idea into separate settings does not preclude a future setting called `simpleGit` which imposes an opinionated simplified Git interface by toggling one or more individual settings, rather than having to set each individually. However, for the time being, it would seem that what constitutes a simplified interface is up to the user/org to decide.

kgryte: For reference, see https://github.com/jupyterlab/jupyterlab-git/pull/482. This stalled due to a bug and needs to be updated again to showcase more recent additions.

kgryte: Could either add a separate binder or update the current binder to provide a tutorial of sorts which guides the user into toggling the various settings and seeing their effects.

kgryte:
@fcollonval Thanks for reviewing! Should be ready for merging provided no further nits!

README.md
mlucool: Maybe we could document the set of all changes that simplify the UI in some way then? I.e. the refresh interview is unrelated but per above `disableBranchWithChanges` and `simpleStaging` are.

kgryte: @mlucool Agreed. I will submit a follow-up PR once I submit a PR related to aborting a merge if there are conflicts (current WIP). The latter PR will add another setting which is also related to the simplified UI concept.

kgryte:
@telamonian Thanks!

Issue #543, opened 2020-02-26
Unicode values for file names for a file name using Chinese


lresende:
## Description

When using the git extension in Jupyterlab and trying to list the status of files named using Chinse characters only unicode values are displayed.

JupyterLab file browser

<img width="401" alt="image" src="https://user-images.githubusercontent.com/382917/75341170-26f1b400-5894-11ea-926a-30dd588675d0.png">

Git status panel

<img width="406" alt="image" src="https://user-images.githubusercontent.com/382917/75341202-396bed80-5894-11ea-950d-b9cd5f1ec604.png">

## Reproduce

Name a file using Chinese characters and try to view its status on the git panel

## Expected behavior

The filename would display properly as in the FileBrowser

lresende:
Investigating this issue further, it seems to be a general git configuration issue, as you would get the same result if you use the git command line:

```
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	"pipelines/\303\263\303\255.ipynb"
	"\346\237\220\344\272\213.ipynb"

```

Git has a configuration to disable this behavior:

```
git config --global core.quotepath false
```

Which then give you the proper result in the git command line

```
git status
On branch master
Your branch is up to date with 'origin/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	pipelines/óí.ipynb
	某事.ipynb

```

As well as in the jupyterlab git ui:
<img width="399" alt="image" src="https://user-images.githubusercontent.com/382917/75341633-23aaf800-5895-11ea-9bd4-f6f577af81c2.png">

ajbozarth:
I already have a fix for this in my fork and was going to bring it up at scrum today

PR #545, opened 2020-02-26
Don't escape filenames with non-ascii characters

ajbozarth:
By default git escapes non-ascii characters in filenames.
I added an inline config change to the git status call to disable this.
By making this an in-line commit we make sure filenames are human
readable without requiring the user to handle gitconfigs on their own.

PR #548, opened 2020-03-04
Allow filenames with non-ascii characters

ianhi:
This encodes the filenames from `git status` to utf-8. With this change the filenames are properly displayed in the UI and git commands such as add, and push still function as expected as git can accept filenames in utf-8 encoding because "Git is to some extent character encoding agnostic." see: https://git-scm.com/docs/git-commit/2.13.4#_discussion

Fixes https://github.com/jupyterlab/jupyterlab-git/issues/543 and https://github.com/jupyterlab/jupyterlab-git/pull/545
<details><summary>gif of this working</summary>

working
</details>

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -264,7 +264,6 @@ class Git:
                 # if file was renamed, next line contains original path
                 "from": next(line_iterable) if line[0]=='R' else line[3:]
             })       
-
         return {"code": code, "files": result}
 
     async def log(self, current_path, history_count=10):
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -5,7 +5,6 @@ import os
 import re
 import subprocess
 from urllib.parse import unquote
-from codecs import decode, escape_decode
 
 import pexpect
 import tornado
@@ -169,14 +168,14 @@ class Git:
             }
         """
         if single_commit:
-            cmd = ["git", "diff", "{}^!".format(single_commit), "--name-only"]
+            cmd = ["git", "diff", "{}^!".format(single_commit), "--name-only", "-z"]
         elif base and remote:
             if base == "WORKING":
-                cmd = ["git", "diff", remote, "--name-only"]
+                cmd = ["git", "diff", remote, "--name-only", "-z"]
             elif base == "INDEX":
-                cmd = ["git", "diff", "--staged", remote, "--name-only"]
+                cmd = ["git", "diff", "--staged", remote, "--name-only", "-z"]
             else:
-                cmd = ["git", "diff", base, remote, "--name-only"]
+                cmd = ["git", "diff", base, remote, "--name-only", "-z"]
         else:
             raise tornado.web.HTTPError(
                 400, "Either single_commit or (base and remote) must be provided"
@@ -194,7 +193,7 @@ class Git:
                 response["command"] = " ".join(cmd)
                 response["message"] = error
             else:
-                response["files"] = output.strip().split("\n")
+                response["files"] = output.strip("\x00").split("\x00")
 
         return response
 
@@ -237,7 +236,7 @@ class Git:
         """
         Execute git status command & return the result.
         """
-        cmd = ["git", "status", "--porcelain", "-u"]
+        cmd = ["git", "status", "--porcelain", "-u", "-z"]
         code, my_output, my_error = await execute(
             cmd, cwd=os.path.join(self.root_dir, current_path),
         )
@@ -250,22 +249,24 @@ class Git:
             }
 
         result = []
-        line_array = my_output.splitlines()
-        for line in line_array:
-            to1 = None
-            from_path = line[3:]
-            if line[0] == "R":
-                to0 = line[3:].split(" -> ")
-                to1 = to0[len(to0) - 1]
+        line_iterable = iter(my_output.strip("\x00").split('\x00'))
+        result = []
+        for line in line_iterable:
+            x = line[0]
+            y = line[1]
+            if line[0]=='R':
+                #If file was renamed then we need both this line
+                #and the next line, then we want to move onto the subsequent
+                #line. We can accomplish this by calling next on the iterable
+                to = line[3:]
+                from_path = next(line_iterable)
             else:
-                to1 = line[3:]
-            if to1.startswith('"'):
-                to1 = to1[1:]
-            if to1.endswith('"'):
-                to1 = to1[:-1]
-            to1 = decode(escape_decode(to1)[0],'utf-8')
-            from_path = decode(escape_decode(from_path)[0],'utf-8')
-            result.append({"x": line[0], "y": line[1], "to": to1, "from": from_path})
+                #to and from_path are the same
+                from_path = line[3:]
+                to = line[3:]
+            result.append({"x": x, "y": y, "to": to, "from": from_path})
+
+
         return {"code": code, "files": result}
 
     async def log(self, current_path, history_count=10):
@@ -314,10 +315,10 @@ class Git:
 
     async def detailed_log(self, selected_hash, current_path):
         """
-        Execute git log -1 --stat --numstat --oneline command (used to get
+        Execute git log -1 --stat --numstat --oneline -z command (used to get
         insertions & deletions per file) & return the result.
         """
-        cmd = ["git", "log", "-1", "--stat", "--numstat", "--oneline", selected_hash]
+        cmd = ["git", "log", "-1", "--stat", "--numstat", "--oneline", "-z", selected_hash]
         code, my_output, my_error = await execute(
             cmd, cwd=os.path.join(self.root_dir, current_path),
         )
@@ -329,7 +330,7 @@ class Git:
         note = [0] * 3
         count = 0
         temp = ""
-        line_array = my_output.splitlines()
+        line_array = re.split("\x00|\n|\r\n|\r", my_output)
         length = len(line_array)
         INSERTION_INDEX = 0
         DELETION_INDEX = 1
@@ -342,7 +343,7 @@ class Git:
                     note[count] = words[i]
                     count += 1
             for num in range(1, int(length / 2)):
-                line_info = line_array[num].split(maxsplit=2)
+                line_info = line_array[num].split('\t', maxsplit=2)
                 words = line_info[2].split("/")
                 length = len(words)
                 result.append(
@@ -373,14 +374,14 @@ class Git:
         """
         Execute git diff command & return the result.
         """
-        cmd = ["git", "diff", "--numstat"]
+        cmd = ["git", "diff", "--numstat", "-z"]
         code, my_output, my_error = await execute(cmd, cwd=top_repo_path)
 
         if code != 0:
             return {"code": code, "command": " ".join(cmd), "message": my_error}
 
         result = []
-        line_array = my_output.splitlines()
+        line_array = my_output.strip('\x00').split('\x00')
         for line in line_array:
             linesplit = line.split()
             result.append(
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -200,7 +200,7 @@ class Git:
                 response["command"] = " ".join(cmd)
                 response["message"] = error
             else:
-                response["files"] = output.strip("\x00").split("\x00")
+                response["files"] = self.strip_and_split(output)
 
         return response
 
@@ -256,7 +256,7 @@ class Git:
             }
 
         result = []
-        line_iterable = iter(my_output.strip("\x00").split('\x00'))
+        line_iterable = iter(self.strip_and_split(my_output))
         result = []
         for line in line_iterable:
             x = line[0]
@@ -388,7 +388,7 @@ class Git:
             return {"code": code, "command": " ".join(cmd), "message": my_error}
 
         result = []
-        line_array = my_output.strip('\x00').split('\x00')
+        line_array = strip_and_split(my_output)
         for line in line_array:
             linesplit = line.split()
             result.append(
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -336,7 +336,7 @@ class Git:
                 # we need next two lines of output
                 from_path = next(line_iterable)
                 to_path = next(line_iterable)
-                modified_file_name = from_path + " ⮕ " + to_path
+                modified_file_name = from_path + " => " + to_path
                 modified_file_path = to_path
             else:
                 modified_file_name = file.split("/")[-1]
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -258,12 +258,13 @@ class Git:
         line_iterable = iter(strip_and_split(my_output))
         for line in line_iterable:
             from_path = line[3:]
-            if line[0]=='R':
-                #If file was renamed then we need both this line
-                #and the next line, then we want to move onto the subsequent
-                #line. We can accomplish this by calling next on the iterable
-                from_path = next(line_iterable)
-            result.append({"x": line[0], "y": line[1], "to": line[3:], "from": from_path})
+            result.append({
+                "x": line[0],
+                "y": line[1],
+                "to": line[3:],
+                # if file was renamed, next line contains original path
+                "from": next(line_iterable) if line[0]=='R' else line[3:]
+            })       
 
         return {"code": code, "files": result}
 
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -5,6 +5,7 @@ import os
 import re
 import subprocess
 from urllib.parse import unquote
+from codecs import decode, escape_decode
 
 import pexpect
 import tornado
@@ -262,6 +263,8 @@ class Git:
                 to1 = to1[1:]
             if to1.endswith('"'):
                 to1 = to1[:-1]
+            to1 = decode(escape_decode(to1)[0],'utf-8')
+            from_path = decode(escape_decode(from_path)[0],'utf-8')
             result.append({"x": line[0], "y": line[1], "to": to1, "from": from_path})
         return {"code": code, "files": result}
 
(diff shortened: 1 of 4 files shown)


fcollonval:
Thanks @ianhi for this PR. I think the approach is better as per @telamonian [comment](https://github.com/jupyterlab/jupyterlab-git/pull/545#issuecomment-592150779).

Could you add test for this case?
You could look there for examples: https://github.com/jupyterlab/jupyterlab-git/blob/master/jupyterlab_git/tests/test_branch.py

I got a question too. Why are you not converting back the file names before calling the git commands?

ajbozarth:
I checked out and played with you code and I agree that this is a better way of dealing with the issue than the angle I was working.

I would suggest pulling your code into a function like you had in the [comment](https://github.com/jupyterlab/jupyterlab-git/pull/545#issuecomment-594176530) on my PR, because this only addresses one location the problem is found.

From a little bit more looking I also found that line 351 needs to this change to display files listed in each commit in the log section. From my light testing I couldn't see any other locations where it was still displaying using you fix once I added it to that location.

ianhi:
@ajbozarth interestingly I've actually experience the same but in the reverse direction, I'm really coming around to `-z`. This is because I checked out how github desktop handles parsing the status ([here](https://github.com/desktop/desktop/blob/a75fdc4b942e8cbf2a70b5435f37f09632e5b664/app/src/lib/status-parser.ts#L54-L66)) and they say the following:

> There is also an alternate -z format recommended for machine parsing. In that
  format, the status field is the same, but some other things change. First,
  the -> is omitted from rename entries and the field order is reversed (e.g
  from -> to becomes to from). Second, a NUL (ASCII 0) follows each filename,
  replacing space as a field separator and the terminating newline (but a space
  still separates the status field from the first filename). Third, filenames
  containing special characters are not specially formatted; no quoting or
  backslash-escaping is performed.

This is combined with the fact that  `codecs.escape_decode` is actually an undocumented part of the python api (see https://stackoverflow.com/a/23151714/835607) it feels as though making git not modify the file names and slightly change the parsing might be a more robust strategy. The biggest change is that you need to split lines on `\x00` instead of `\n`. With this knowledge we can rewrite `status` as like so:
```python
cmd = ["git", "status", "--porcelain", "-u", "-z"]
...
...
...
result = []
# [:-1] to ignore trailing \x00
line_iterable = iter(my_output.split('\x00')[:-1])
result = []
for line in line_iterable:
    x = line[0]
    y = line[1]
    if line[0]=='R':
        #If file was renamed then we need both this line
        #and the next line, then we want to move onto the subsequent
        #line. We can accomplish this by calling next on the iterable
        to = line[3:]
        from_path = next(line_iterable)
    else:
        #to and from_path are the same
        from_path = line[3:]
        to = line[3:]
    result.append({"x": x, "y": y, "to": to, "from": from_path})
```
@fcollonval this seems better to me because it doesn't use an undocumented function, but I'm happy to do however you'd like. As far as I can tell there are four functions that need to be modified to deal with parsing: `status`,`diff`,`detailed_log`, and `changed_files`. Using `-z` can easily fix filename parsing (up to changing `split()` to `split('\x00')`) for all of these except for `detailed_log` which doesn't properly render all occurrences of filenames even in the terminal. So that function may require the codecs solution no matter what.
As for why I don't have to convert the filenames back I believe this is a consequence of (via git-scm):

> Note that Git at the core level treats path names simply as sequences of non-NUL bytes, there are no path name encoding conversions

So that leaves me with three questions

1. Is `-z` with modified parsing or `escape_decode` preferable
2. If I make a function for the decoding should it go anywhere in particular? Perhaps at the top before `execute`
3. Can you specify a little bit more what you're looking for in tests? My understanding is that you'd like a case to check that the functions all work when passed non-ascii characters.

ajbozarth:
1. I didn't realize that `escape_decode` is an undocumented api, in light of that I agree this isn't the best solution. Based on your feedback, I try editing my pr to handle splitting on `\x00` while using `-z`. I had started to follow that path prior to testing your idea. If I can get it working I will open my PR back up with that fix. If I can't get it to work we'll use this PR (Should be able to determine which we'll do by EOD Tuesday at the latest)
2. If we use this PR, then I would put the function below `__init__` in the `Git` class.
3. As for testing I've been making sure a file with the non-ascii characters can follow the process from creation to git commit successfully and be displayed correctly in the ui along the way.

ianhi:
Sounds good. When I was playing around with `-z` I ended up creating a branch that uses only `-z` to fix non-ascii characters. You can find that here: https://github.com/jupyterlab/jupyterlab-git/compare/master...ianhi:non-ascii-z Feel free to use that as inspiration, or if its easier I can make a pull request from that branch.

ajbozarth:
I ran out of time today to get to this, I'll look back on it on Monday

fcollonval:
@ianhi  and @ajbozarth Thanks for the discussion and pushing on this. I had a quick look at VS Code. And actually they use the `-z` option (the full call being `git status -z -u`). So it may be the best fix.

ajbozarth:
@ianhi I've reviewed the branch you linked and I think it's the best path forward, I do have a couple review comments I can make on it though. If you could either merge that code here for review

jupyterlab_git/git.py
```diff
@@ -249,20 +249,24 @@ def __init__(self, contents_manager):
...
         result = []
-        line_array = my_output.splitlines()
-        for line in line_array:
-            to1 = None
-            from_path = line[3:]
-            if line[0] == "R":
-                to0 = line[3:].split(" -> ")
-                to1 = to0[len(to0) - 1]
+        line_iterable = iter(my_output.strip("\x00").split('\x00'))
+        result = []
```
ajbozarth: unnecessary duplicate of line 251

jupyterlab_git/git.py
```diff
@@ -249,20 +249,24 @@ def __init__(self, contents_manager):
...
+        for line in line_iterable:
+            x = line[0]
```
ajbozarth: why move these into a var if they are only used in the one location? I would leave them inside the append function as they were before

jupyterlab_git/git.py
```diff
@@ -193,7 +193,7 @@ def __init__(self, contents_manager):
                 response["command"] = " ".join(cmd)
                 response["message"] = error
             else:
-                response["files"] = output.strip().split("\n")
+                response["files"] = output.strip("\x00").split("\x00")
```
ajbozarth: given you call `.strip("\x00").split("\x00")` three times in this change it might be worth creating function for it.

jupyterlab_git/git.py
```diff
@@ -249,20 +249,24 @@ def __init__(self, contents_manager):
...
+            y = line[1]
+            if line[0]=='R':
+                #If file was renamed then we need both this line
+                #and the next line, then we want to move onto the subsequent
+                #line. We can accomplish this by calling next on the iterable
+                to = line[3:]
+                from_path = next(line_iterable)
             else:
-                to1 = line[3:]
-            if to1.startswith('"'):
```
ajbozarth: why is this `"` handling no longer necessary?

ianhi:
Ok. I merged that branch into the branch for this pull request and added a file to test the status function. I think I dealt the the `current_directory` argument correctly but if I had to guess where something might be wrong it would be there.

jupyterlab_git/git.py
```diff
@@ -249,20 +249,24 @@ def __init__(self, contents_manager):
...
-                to1 = to1[1:]
-            if to1.endswith('"'):
-                to1 = to1[:-1]
-            result.append({"x": line[0], "y": line[1], "to": to1, "from": from_path})
+                #to and from_path are the same
+                from_path = line[3:]
+                to = line[3:]
+            result.append({"x": x, "y": y, "to": to, "from": from_path})
```
ajbozarth: Based on my understanding on the flow here, this whole chunk inside the for loop could be simplified to:
```
            from_path = line[3:]
            if line[0]=='R':
                #If file was renamed then we need both this line
                #and the next line, then we want to move onto the subsequent
                #line. We can accomplish this by calling next on the iterable
                from_path = next(line_iterable)
            result.append({"x": line[0], "y": line[1], "to": line[3:], "from": from_path})
```

jupyterlab_git/git.py
```diff
@@ -249,20 +249,24 @@ def __init__(self, contents_manager):
...
+
+
```
ajbozarth: remove extra newline

jupyterlab_git/git.py
```diff
@@ -326,7 +330,7 @@ def __init__(self, contents_manager):
         note = [0] * 3
         count = 0
         temp = ""
-        line_array = my_output.splitlines()
+        line_array = re.split("\x00|\n|\r\n|\r", my_output)
```
ajbozarth: why the slightly different split here? Aren't `\r` and `\n` converted to `\x00` by the `-z` flag?

jupyterlab_git/git.py
```diff
@@ -339,7 +343,7 @@ def __init__(self, contents_manager):
                     note[count] = words[i]
                     count += 1
             for num in range(1, int(length / 2)):
-                line_info = line_array[num].split(maxsplit=2)
+                line_info = line_array[num].split('\t', maxsplit=2)
```
ajbozarth: why is the `/t` now necessary?

jupyterlab_git/git.py
ianhi: As far as I can tell git only adds the quotes when you have non-unicode characters and don't use `-z`

ajbozarth: If you're sure that's the case then I'm fine with it but if not then it's probably worth replacing `to` with `to.strip('\"')` to be sure

ianhi:
I have a question about how to best handle the renaming of files in the history tab of the ui, which is derived from the `detailed_log` function. If you move a file into a folder(zoop.py => test_dir/zoop.py) then history currently shows this (from master branch):
image
contrast that with what github desktop shows:
image

If you rename a file then the history tab shows:
image

So currently inconsistent things are being returned for similar actions. What should be returned in both of these cases? Since `detailed_log` is getting modified in this pull request maybe this is a good time to address this? At issue seems to be what should be returned for the value of `modified_file_name` and perhaps what `modified_file_path` ought to be for when a file is moved or renamed. As a user my preference is for the two cases to display as:

`zoop.py ⮕ test_dir/zoop.py`
and 
`γ.gamma ⮕ name_with spaces`
or if the move happens to a very long file path:
`γ.gamma ⮕ .../name_with spaces`

jupyterlab_git/git.py
ianhi: Originally by accident, but on further investigation the `-z` flag splits renaming of files onto multiple lines, so now if there are spaces in a name it could get split weirdly. i.e.:
without `-z`:  
`0    0    file1 => renamed file with spaces`
with `-z` that becomes:  
```
0    0
file1
renamed file with spaces```
so the `maxsplit=2` no longer protects from splitting filenames on spaces. Though this needs further modification to fix how it handles when files are renamed. I added a comment to the main thread here asking how to handle renaming of files in the history tab.

File: jupyterlab_git/git.py (hunk 7 above)
Comment (ianhi) [reply jupyterlab_git/git.py]: Apparently only sometimes. if you try `git log -1 --stat --numstat --oneline -z` you can see that only the first half has the newlines swapped for `\x00`, the second half (from --stat) doesn't seem to be affected by `-z`:
image

Though the output of --stat only seems to be used to grab the final line of that output, which could be generated from the file parsing that already gets done, so this weird splitting could be eliminated by dropping the `--stat` and changing the parsing logic slightly.

**Comment by fcollonval** [comment]
Hey just some information on the format from the [documentation](https://git-scm.com/docs/git-status). Using "-z" with status command implies porcelain format v1:

> There is also an alternate -z format recommended for machine parsing. In that format, the status field is the same, but some other things change. First, the -> is omitted from rename entries and the field order is reversed (e.g from -> to becomes to from). Second, a NUL (ASCII 0) follows each filename, replacing space as a field separator and the terminating newline (but a space still separates the status field from the first filename). Third, filenames containing special characters are not specially formatted; no quoting or backslash-escaping is performed.

So quoting is indeed unneeded.

File: jupyterlab_git/git.py [hunk 9]
```diff
@@ -110,6 +110,13 @@ class Git:
     def __init__(self, contents_manager):
         self.contents_manager = contents_manager
         self.root_dir = os.path.expanduser(contents_manager.root_dir)
+        
+    def strip_and_split(self, s):
```
fcollonval:
```suggestion
    @staticmethod
    def strip_and_split(s):
```

As the object is not needed, it is better to define this as a static method or to place it outside the class as utility function

jupyterlab_git/git.py
ianhi: Good simplification, I changed to be this.

jupyterlab_git/git.py
ianhi: Great point thanks. I moved the function outside the object.

ianhi:
I think I addressed all the comments in latest commit. Sorry it took a few days, coronavirus has really upended life at universities.

ajbozarth:
@ianhi thanks for the update, I'll take a look and review your updates removing `--stat`, as well as your tests, this afternoon. The rest of your changes look good.

jupyterlab_git/git.py
```diff
@@ -311,73 +313,70 @@ def __init__(self, contents_manager):
...
+        line_iterable = iter(strip_and_split(my_output)[1:])
+        for line in line_iterable:
+            insertions, deletions, file = line.split('\t')
+
+            if file == '':
+                # file was renamed or moved
+                # we need next two lines of output
+                from_path = next(line_iterable)
+                to_path = next(line_iterable)
+                modified_file_name = from_path + " ⮕ " + to_path
```
ajbozarth: I'm not sure introducing unicode symbol here is the best plan, I would just use `=>` like the git cli does.

jupyterlab_git/git.py
```diff
@@ -249,20 +255,16 @@ def __init__(self, contents_manager):
...
-                to1 = to1[1:]
-            if to1.endswith('"'):
-                to1 = to1[:-1]
-            result.append({"x": line[0], "y": line[1], "to": to1, "from": from_path})
+            if line[0]=='R':
+                #If file was renamed then we need both this line
+                #and the next line, then we want to move onto the subsequent
+                #line. We can accomplish this by calling next on the iterable
+                from_path = next(line_iterable)
+            result.append({"x": line[0], "y": line[1], "to": line[3:], "from": from_path})
```
telamonian:
```suggestion
            result.append({
                "x": line[0],
                "y": line[1],
                "to": line[3:],
                # if file was renamed, next line contains original path
                "from": next(line_iterable) if line[0]=='R' else line[3:]
            })       
```

This can be a bit clearer. Remove intermediate value `from_path` and use shorter/more succinct comment

jupyterlab_git/git.py
```diff
@@ -249,20 +255,16 @@ def __init__(self, contents_manager):
             }
 
         result = []
-        line_array = my_output.splitlines()
-        for line in line_array:
-            to1 = None
+        line_iterable = iter(strip_and_split(my_output))
+        for line in line_iterable:
             from_path = line[3:]
```
telamonian:
```suggestion
```

Intermediate value `from_path` is not needed. See next suggestion below

jupyterlab_git/git.py
telamonian:
```suggestion
                modified_file_name = from_path + " => " + to_path
```

Although in general I support any innovation in arrow rendering (at least those that make the arrows prettier), unicode in source should sadly still be avoided for the moment, since many tools (eg github) still can't correctly render/use old fonts that don't contain many glyphs

ianhi:
I think that that should address all the requested changes. As for fixing the original issue it fixes all the issues I had. I believe that @ajbozarth is the other user who encountered errors, does it fix yours?

**Unicode:**
 I included several unicode characters in the tests for [status](https://github.com/jupyterlab/jupyterlab-git/pull/548/files#diff-2ff2b18e5a53b3a9b844d1748012b7d8R27-R28) and [detailed_log](https://github.com/jupyterlab/jupyterlab-git/pull/548/files#diff-2ff2b18e5a53b3a9b844d1748012b7d8R27-R28) based on my understanding the request for test [above](https://github.com/jupyterlab/jupyterlab-git/pull/548#issuecomment-594355767). Is this a problem? Originally it was necessary to include non-ascii characters in the tests because they required different behaviour, but with the use of `-z` it's likely no longer necessary to include non-ascii characters in the test cases. My feeling would be to leave them as that feels like a more complete set of tests, but happy to change.

As an aside:
Thanks for the lesson in why not use unicode in source. I don't think I've ever even had to think about that before. Though maybe on github this is a browser issue, it seems to render fine for me: 
image

telamonian:
> I included several unicode characters in the tests for [status](https://github.com/jupyterlab/jupyterlab-git/pull/548/files#diff-2ff2b18e5a53b3a9b844d1748012b7d8R27-R28) and [detailed_log](https://github.com/jupyterlab/jupyterlab-git/pull/548/files#diff-2ff2b18e5a53b3a9b844d1748012b7d8R27-R28) based on my understanding the request for test [above](https://github.com/jupyterlab/jupyterlab-git/pull/548#issuecomment-594355767). Is this a problem?

No, it's not a problem here since it's necessary for these tests, regardless of how we're solving the non-ascii issue (`-z` or otherwise).

> Though maybe on github this is a browser issue, it seems to render fine for me

Here's what it looks like for me:

image

It probably has to do with the particular font being used, which, yeah, is determined by some combination of github's CSS and the quirks of our individual browsers.

@ianhi In general, you should treat "don't use unicode in src" as a rule of thumb rather than as a hard law. If you stick to ascii it ensures that code will render correctly in everyone's tools and editors, but in reality support for non-ascii has improved by leaps and bounds in the last few years. For example, the Greek characters you included in the tests do render fine for me.

I think the deeper issue here is that no font has support for the complete set of unicode glyphs, so it's hard to tell apriori what set of glyphs we should consider to be "safe" for including in src.

PR #520, opened 2020-01-09
[WIP] Update for JupyterLab 2.0

blink1073:
Note: this can't be merged without a release of nbdime: https://github.com/jupyter/nbdime/pull/508

There is also discussion in #517 about when master should target 2.0.

Summary of changes:

- Rename `@phosphor` -> `@lumino` for dependencies and imports
- Use `@jupyterlab/nbformat` and `@jupyterlab/settingregistry` that were split out from `@jupyterlab/coreutils`
- Change icons to use `JLICon`

PR #550, opened 2020-03-04
[WIP ]Labicon migration jlab 2.0

telamonian:


lresende:
It would be great if this could be relative to @blink1073 pr's (or even an official JupyterLab 2.0 branch) that brings support to JupyterLab 2.0 instead of master, which would enable us to do a last release of Master that supports JupyterLab 1.x (See #554)

telamonian:
@lresende Neither this nor #520 are going to be pulled in until 0.10.0 is released

Issue #528, opened 2020-01-27
Jlab freezes if incorrect clone url or credentials or already cloned repo url is given


chrisjohn2306:
## Description

When the user tries to clone a repo that already exists in the base path, the window freezes and no error or status are shown.
Same behaviour is seen if the clone url is incorrect, or the credentials entered by the user is incorrect.
## Reproduce

1. Open the jupyter lab with the extension enabled.
2. Clone a repo in to the home folder
3. (a)After the clone is finished, try to clone the same repo again
(b) In the terminal prompt for entering clone url, enter a invalid one
(c) After entering the correct clone url, in the credential form, enter incorrect username or password
4. Window freezes with no error in browser console or logs

## Expected behavior

Either a dialog box appears saying that the repo already exists or some status that indicates that the operation has failed. 
Input Validation on the entered values 
## Context

- Python package version: 3.5

- Extension version: v0.10.0

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>
<pre>
Paste the output from your browser Javascript console here.
</pre>
</details>

fcollonval:
Thanks for reporting @chrisjohn2306 

Which OS are you using?
Did you have to authenticate through the extension or were your credentials managed by git directly?

chrisjohn2306:
I am using MacOS. When the repeated clone is given it never reaches the authentication part. Window freezes before that.

chrisjohn2306:
While looking more into the issue i could note that, if an existing repo is attempted to clone again, instead of clone a git pull happens. But no message or progress bar is present to indicate that. 
Is it a default behaviour of git?

chrisjohn2306:
I have a fix for this issue along with the fix for the concern raised in https://github.com/jupyterlab/jupyterlab-git/issues/534 regarding the confusing warning message.

I would like to contribute it to the repository. Is there any T&C formalities to contribute to the repo ?

fcollonval:
> I have a fix for this issue along with the fix for the concern raised in #534 regarding the confusing warning message.

Awesome!!

> I would like to contribute it to the repository. Is there any T&C formalities to contribute to the repo ?

The contribution rules are those of JupyterLab: https://github.com/jupyterlab/jupyterlab-git#contributing

So feel free to open a PR with your fix.

chrisjohn2306:
[...]

PR #553, opened 2020-03-08
Check if repository dir doesn't already exists when cloning

lresende:
When trying to clone a repository, if the repository
directory already exists the operation was causing
JupyterLab to hang. With these changes, an error
message is sent if the directory already exists.

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -170,6 +170,13 @@ class Git:
         :param auth: OPTIONAL dictionary with 'username' and 'password' fields
         :return: response with status code and error message.
         """
+
+        # Validate folder don't already exists
+        dir_name = self._get_git_dir_name(unquote(repo_url))
+        full_dir_name = os.path.join(self.root_dir, current_path, dir_name)
+        if os.path.exists(full_dir_name):
+            raise HTTPError(500, 'Directory {} already exists.'.format(dir_name))
+
         env = os.environ.copy()
         if (auth):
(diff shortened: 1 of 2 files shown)


lresende:
This should address the hanging when the repository is already cloned (or folder already exists) as described in #528

telamonian:
I don't love this approach:

- If the directory already exists *but is empty*, then `git clone` is a valid op
- Why is our gui hanging in this case in the first place? On the command line you do get

    ```
    fatal: destination path 'jupyterlab-git' already exists and is not an empty directory.
    ```

    What's the actual root cause of the problem here?

jupyterlab_git/git.py
```diff
@@ -170,6 +170,13 @@ def clone(self, current_path, repo_url, auth=None):
         :param auth: OPTIONAL dictionary with 'username' and 'password' fields
         :return: response with status code and error message.
         """
+
+        # Validate folder don't already exists
```
kgryte:
```suggestion
        # Validate that the folder doesn't already exist
```

telamonian:
@lresende I fixed up #556 so that it should solve the underlying issue causing the "freeze on clone existing" for everyone. I'd prefer to take the approach in that PR (correctly pass along the error message from git itself) rather than the approach in this PR (add our own pre-emptive error handling). Can you please review #556 and confirm that it solves the same problem you were trying to solve here?

lresende:
This is being taken care by #556

PR #556, opened 2020-03-09
Updating Exception Handling Scenarios

chrisjohn2306:
Changes are to handle 
- In Push/Pull Credential dialog box, incorrect warning message is coming
- For Git clone, if an incorrect url is entered, Jupyter Lab freezes
- If exisiting repo is attempted for a Git clone again, Jupyter Lab freezes
- In Git credentials dialog box, if incorrect username/password is entered, Jupyter Lab freezes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,13 +1,24 @@
+# os junk
 .DS_Store
+
+# yarn/typescript build junk
+package-lock.json
+*.tsbuildinfo
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
 *$py.class
+.pytest_cache/
 
 # C extensions
 *.so
 
 # Distribution / packaging
+*.egg
+*.egg-info/
+.eggs/
+.installed.cfg
 .Python
 env/
 build/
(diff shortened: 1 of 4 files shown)


chrisjohn2306:
@fcollonval Can you go review the changes?

lresende:
I tried these changes, and if you attempt to clone a repository that is already cloned in your local environment JupyterLab is still freezing.

chrisjohn2306:
I tried to to clone repo that already is existing in my home directory. Instead of Jlab freezing the error is thrown.
Screenshot 2020-03-10 at 8 32 43 AM

telamonian:
I dug into this and the "freeze on clone existing" issue. The freeze turns out to be caused by an unhandled conditional branch within a non-terminating `while` loop:

https://github.com/jupyterlab/jupyterlab-git/blob/e1fb6a32bc5eebdb74c4bddd12758210bd204c2f/src/widgets/gitClone.tsx#L60-L88

Adding an `else` to go with the `if` at the top level of the loop will make it so that on error we show a dialog with the error, instead of getting stuck in the loop indefinitely:

https://github.com/jupyterlab/jupyterlab-git/blob/0bbade7a0e797183663537877b1c869f77ff99f1/src/widgets/gitClone.tsx#L60-L93

Building on @chrisjohn2306's work, I fixed up his tweaks so that the related code should now work as intended for everyone. Can someone (@chrisjohn2306, @lresende, or anyone who would care to) please confirm that the current version of this PR fixes the "freeze on clone existing" issue for them?

lresende:
@telamonian these changes takes care of the "freeze on clone" when a project is already cloned. Thanks for taking a look at that.

Issue #555, opened 2020-03-08
Discard changes fails due to index.lock already exists


lresende:
## Description

Discard changes fails due to index.lock already exists

## Reproduce

1. Go to a directory that represents cloned git repository
2. Make changes to a given file
3. Click on the icon to pull from the repository
![image](https://user-images.githubusercontent.com/382917/76171722-afe9d480-614b-11ea-8a1c-58b6b35106e5.png)

4. Now, try go discard the changes in order to be able to pull from the repository
![image](https://user-images.githubusercontent.com/382917/76171748-00613200-614c-11ea-8a5f-899c2c2842a4.png)

5. Discard fails due to `index.lock` is already present
![image](https://user-images.githubusercontent.com/382917/76171727-c2640e00-614b-11ea-8360-01d672c1cbf5.png)

## Expected behavior

Discard should discard the changes for the file

## Context

- Python package version:
  - JupyterLab 1.2.7
  - Jupyterlab-git master

## Workaround

It seems that on the second try it actually behaves well, needs to investigate it further.

lresende:
This seems related to the following git config in my environment
```
[branch]
        autosetuprebase = always
```
Maybe we could check for this and decide if we should use `git merge --abort` versus `git rebase --abort`

fcollonval:
This should be fixed by #564

PR #564, opened 2020-03-17
Fix race condition issues on slower FS

mlucool:
This adds locking before executing commands. While this may add locking in some cases that are not a race (i.e. things that do not need to acquire index.lock), this solves issues related to using this with a git directory on NFS.

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -8,6 +8,8 @@ from urllib.parse import unquote
 
 import pexpect
 import tornado
+import tornado.locks
+import datetime
 
 
 # Git configuration options exposed through the REST API
@@ -16,7 +18,14 @@ ALLOWED_OPTIONS = ['user.name', 'user.email']
 # See https://git-scm.com/docs/git-config#_syntax for git var syntax
 CONFIG_PATTERN = re.compile(r"(?:^|\n)([\w\-\.]+)\=")
 DEFAULT_REMOTE_NAME = "origin"
+# Ensure on NFS or similar, that we give the .git/index.lock time to be removed
+MAX_WAIT_FOR_LOCK_S = 5
+# How often should we check for the lock above to be free?
+CHECK_LOCK_INTERVAL_S = 0.1
+# How long to wait to be executed or finished your execution before timing out
+MAX_WAIT_FOR_EXECUTE_S = 20
 
+execution_lock = tornado.locks.Lock()
 
 async def execute(
     cmdline: "List[str]",
@@ -85,19 +94,32 @@ async def execute(
         output, error = process.communicate()
         return (process.returncode, output.decode("utf-8"), error.decode("utf-8"))
 
-    if username is not None and password is not None:
-        code, output, error = await call_subprocess_with_authentication(
-            cmdline,
-            username,
-            password,
-            cwd,
-            env,
-        )
-    else:
-        current_loop = tornado.ioloop.IOLoop.current()
-        code, output, error = await current_loop.run_in_executor(
-            None, call_subprocess, cmdline, cwd, env
-        )
+    await execution_lock.acquire(timeout=datetime.timedelta(seconds=MAX_WAIT_FOR_EXECUTE_S))
+    try:
+        # Ensure our execution operation will succeed by first checking and waiting for the lock to be removed
+        time_slept = 0
+        lockfile = os.path.join(cwd, '.git', 'index.lock')
+        while os.path.exists(lockfile) and time_slept < CHECK_LOCK_INTERVAL_S:
+            await tornado.gen.sleep(CHECK_LOCK_INTERVAL_S)
+            time_slept += CHECK_LOCK_INTERVAL_S
+
+        # If the lock still exists at this point, we will likely fail anyway, but let's try anyway
+
+        if username is not None and password is not None:
+            code, output, error = await call_subprocess_with_authentication(
+                cmdline,
+                username,
+                password,
+                cwd,
+                env,
+            )
+        else:
+            current_loop = tornado.ioloop.IOLoop.current()
+            code, output, error = await current_loop.run_in_executor(
+                None, call_subprocess, cmdline, cwd, env
+            )
+    finally:
+        execution_lock.release()
 
     return code, output, error
 
@@ -978,7 +1000,7 @@ class Git:
                 is_binary = await self._is_binary(filename, "", top_repo_path)
                 if is_binary:
                     raise tornado.web.HTTPError(log_message="Error occurred while executing command to retrieve plaintext diff as file is not UTF-8.")
-                    
+
                 curr_content = await self.show(filename, "", top_repo_path)
             else:
                 raise tornado.web.HTTPError(
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -18,18 +18,18 @@ ALLOWED_OPTIONS = ['user.name', 'user.email']
 # See https://git-scm.com/docs/git-config#_syntax for git var syntax
 CONFIG_PATTERN = re.compile(r"(?:^|\n)([\w\-\.]+)\=")
 DEFAULT_REMOTE_NAME = "origin"
-# Ensure on NFS or similar, that we give the .git/index.lock time to be removed
-MAX_WAIT_FOR_LOCK_S = 5
-# How often should we check for the lock above to be free?
-CHECK_LOCK_INTERVAL_S = 0.1
 # How long to wait to be executed or finished your execution before timing out
 MAX_WAIT_FOR_EXECUTE_S = 20
+# Ensure on NFS or similar, that we give the .git/index.lock time to be removed
+MAX_WAIT_FOR_LOCK_S = 5
+# How often should we check for the lock above to be free? This comes up more on things like NFS
+CHECK_LOCK_INTERVAL_S = 0.1
 
 execution_lock = tornado.locks.Lock()
 
 async def execute(
     cmdline: "List[str]",
-    cwd: "Optional[str]" = None,
+    cwd: "str",
     env: "Optional[Dict[str, str]]" = None,
     username: "Optional[str]" = None,
     password: "Optional[str]" = None,
@@ -94,12 +94,16 @@ async def execute(
         output, error = process.communicate()
         return (process.returncode, output.decode("utf-8"), error.decode("utf-8"))
 
-    await execution_lock.acquire(timeout=datetime.timedelta(seconds=MAX_WAIT_FOR_EXECUTE_S))
+    try:
+        await execution_lock.acquire(timeout=datetime.timedelta(seconds=MAX_WAIT_FOR_EXECUTE_S))
+    except  tornado.util.TimeoutError:
+        return (1, "", "Unable to get the lock on the directory")
+
     try:
         # Ensure our execution operation will succeed by first checking and waiting for the lock to be removed
         time_slept = 0
         lockfile = os.path.join(cwd, '.git', 'index.lock')
-        while os.path.exists(lockfile) and time_slept < CHECK_LOCK_INTERVAL_S:
+        while os.path.exists(lockfile) and time_slept < MAX_WAIT_FOR_LOCK_S:
             await tornado.gen.sleep(CHECK_LOCK_INTERVAL_S)
             time_slept += CHECK_LOCK_INTERVAL_S
 
--- a/jupyterlab_git/tests/test_execute.py
+++ b/jupyterlab_git/tests/test_execute.py
@@ -1,47 +1,27 @@
 import pytest
-import tornado
-from unittest.mock import MagicMock, patch, DEFAULT
+from unittest.mock import patch
 
-from jupyterlab_git.git import execute
+from jupyterlab_git.git import execute, execution_lock
 
 
 @pytest.mark.asyncio
-async def test_execute_waits_on_index_lock():
-    lock_exists = False
+async def test_execute_waits_on_index_lock(tmp_path):
+    lock_file = tmp_path / ".git/index.lock"
+    lock_file.parent.mkdir(parents=True, exist_ok=True)
+    lock_file.write_text("")
 
-    def does_lockfile_exist(p):
-        if lock_exists and p == "/dummy/path/.git/index.lock":
-            return True
-
-        return DEFAULT
-
-    def success_when_no_lock(*args, **kwargs):
-        if lock_exists:
-            return tornado.gen.maybe_future((1, "", "Lock found"))
-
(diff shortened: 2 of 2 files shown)


mlucool:
Let me know if you generally like this. Not sure why it can't find `tornado.locks` as it has been there a while: https://www.tornadoweb.org/en/stable/locks.html

jupyterlab_git/git.py
```diff
@@ -8,6 +8,7 @@
 
 import pexpect
 import tornado
```
telamonian:
```suggestion
import tornado
import tornado.locks
```

It looks like the problem in the test is being caused by the fact that `locks` is a seperate module that doesn't get imported by default on `import tornado`. If you're familiar with scipy, it's just like how you would need to do `import scipy.stats` before you can actually reference `scipy.stats` (or whichever scipy module)

telamonian:
Nice work. I like the thinking behind this PR. We will need a better description of the error/issue this is trying to solve (might be nicest if it was written down in a separate issue). I'd also like to better understand what locking (if any) the git CLI itself does.

Ideally, this PR should also include a test to prove that it fixes the problem it's trying to solve. One idea for the test: explicitly lock a resource for ~5 seconds, then immediately start another operation that would create a race condition if locking is not present.

mlucool:
> I'd also like to better understand what locking (if any) the git CLI itself does.

Git locking: https://docs.microsoft.com/en-us/azure/devops/repos/git/git-index-lock?view=azure-devops

Waiting on locks: https://stackoverflow.com/a/36364687

So we can improve this by telling execute which commands care about a lock file or not (e.g. I don't believe status does), but I think thats an over optimization at this point and error prone.

> We will need a better description of the error/issue this is trying to solve
I think https://github.com/jupyterlab/jupyterlab-git/issues/555 is the same issue. Symptoms are at least

> this PR should also include a test to prove that it fixes the problem it's trying to solve

I'll add one

mlucool:
@telamonian do you have an example where execute is tested? `pytest --cov=jupyterlab_git --cov-report html jupyterlab_git` makes it seem like we always mock that out

telamonian:
> do you have an example where execute is tested? `pytest --cov=jupyterlab_git --cov-report html jupyterlab_git` makes it seem like we always mock that out

Hmm, not off the top of my head. Frederic's the expert, since he wrote execute and most of the related code (and the related tests). @fcollonval Any ideas?

telamonian:
I thought of a potential concern: should read-only ops (in particular `git status`) ignore locking? Basically, I'm wondering if there are any potential interactions between this PR and how our UI updates.

mlucool:
Per my comment:
>While this may add locking in some cases that are not a race (i.e. things that do not need to acquire index.lock)

Yes, `git status` will now unnecessarily wait for the lock. We can remove it in some cases trivially (i.e. opt out) but that feels like an pre-mature optimization based on my local testing. If you feel strongly, I'm happy to make this change

mlucool:
@telamonian let me know what you'd like me to do here to get this merged so that it's included in 0.10.

fcollonval:
> 
> 
> > do you have an example where execute is tested? `pytest --cov=jupyterlab_git --cov-report html jupyterlab_git` makes it seem like we always mock that out
> 
> Hmm, not off the top of my head. Frederic's the expert, since he wrote execute and most of the related code (and the related tests). @fcollonval Any ideas?

Hey sorry for the delay. 
Unfortunately as stated in the out-of-date [roadmap](https://github.com/jupyterlab/jupyterlab-git/blob/master/ROADMAP.md). There is no integration test. Meaning that all git command are mocked. But when I introduced the asynchronous calls, I went to mock the `execute` function directly.

But you could test your nice feature by calling directly the `execute` function and mock the `run_in_executor`; the definition to patch is [tornado.ioloop.IOLoop.run_in_executor](https://www.tornadoweb.org/en/stable/ioloop.html#tornado.ioloop.IOLoop.run_in_executor).

fcollonval:
Testing is a bit more complicated that what I say. But that commit shows how to start achieving it: https://github.com/fcollonval/jupyterlab-git/commit/6ede2a8bed6ab9302ba7c62b3dc6018d40ecebbd

jupyterlab_git/git.py
```diff
@@ -85,19 +94,32 @@ def call_subprocess(
...
-            password,
-            cwd,
-            env,
-        )
-    else:
-        current_loop = tornado.ioloop.IOLoop.current()
-        code, output, error = await current_loop.run_in_executor(
-            None, call_subprocess, cmdline, cwd, env
-        )
+    await execution_lock.acquire(timeout=datetime.timedelta(seconds=MAX_WAIT_FOR_EXECUTE_S))
```
fcollonval:
```suggestion
    try:
        await execution_lock.acquire(timeout=datetime.timedelta(seconds=MAX_WAIT_FOR_EXECUTE_S))
    except  tornado.util.TimeoutError:
        return (1, "", "Unable to get the lock on the directory")
```

We should return a proper error message if the lock can not be acquired.

jupyterlab_git/git.py
```diff
@@ -16,7 +18,14 @@
 # See https://git-scm.com/docs/git-config#_syntax for git var syntax
 CONFIG_PATTERN = re.compile(r"(?:^|\n)([\w\-\.]+)\=")
 DEFAULT_REMOTE_NAME = "origin"
+# Ensure on NFS or similar, that we give the .git/index.lock time to be removed
+MAX_WAIT_FOR_LOCK_S = 5
+# How often should we check for the lock above to be free?
+CHECK_LOCK_INTERVAL_S = 0.1
+# How long to wait to be executed or finished your execution before timing out
+MAX_WAIT_FOR_EXECUTE_S = 20
```
fcollonval: There is an inconsistency here. The waiting time for the lock should be greater than the one for execution. Otherwise, for a long run other calls will most likely be dropped due to a failure in acquiring the lock.

fcollonval:
One additional comment:

The argument `cwd` of `execute` should be mandatory, now:
```diff
-    cwd: "Optional[str]" = None,
+    cwd: str,
```

jupyterlab_git/git.py
```diff
@@ -16,7 +18,14 @@
 # See https://git-scm.com/docs/git-config#_syntax for git var syntax
 CONFIG_PATTERN = re.compile(r"(?:^|\n)([\w\-\.]+)\=")
 DEFAULT_REMOTE_NAME = "origin"
+# Ensure on NFS or similar, that we give the .git/index.lock time to be removed
+MAX_WAIT_FOR_LOCK_S = 5
```
fcollonval: Upon better review this one is not used.

telamonian:
I'm targetting this for the next release (which should be 0.20.0). Once this PR is pulled in we'll also backport it to 0.11.0

jupyterlab_git/git.py
mlucool: This should have been used. Fixed. The idea is you have 20s to wait for your turn to execute, and when you run, it will wait up to 5s for the `index.lock`.

mlucool:
@fcollonval this should be all set now. I'll admit the unit test is mediocre because we don't have integration testing. It does prove that if a lock exists, it will sleep and simulates that if a lock exists, a command fails.

mlucool:
> I'm targetting this for the next release (which should be 0.20.0). Once this PR is pulled in we'll also backport it to (jlab 1.0 compatible) 0.11.0

There is no need to backport this to 0.11 for me at this time. I'm hoping to be on 2.x in the near future, but I'll let you know if something changes. If you want to backport anyway, it may be a good idea.

jupyterlab_git/tests/test_execute.py
```diff
@@ -0,0 +1,47 @@
...
+
+                # Simulate answer1 not having released lock
+                lock_exists = True
+                answer2 = await execute(cmd, **kwargs)
+
+                assert executor.call_count == 2
+                assert answer1[0] == 0
+                assert answer2[0] == 1
+                # Did we really try to wait on the lock?
+                sleep.assert_called()
```
fcollonval:
```suggestion
import pytest
from unittest.mock import patch

from jupyterlab_git.git import execute, execution_lock


@pytest.mark.asyncio
async def test_execute_waits_on_index_lock(tmp_path):
    lock_file = tmp_path / ".git/index.lock"
    lock_file.parent.mkdir(parents=True, exist_ok=True)
    lock_file.write_text("")

    async def remove_lock_file(*args):
        assert "unlocked" not in repr(execution_lock)  # Check that the lock is working
        lock_file.unlink()  # Raise an error for missing file
        
    with patch("tornado.gen.sleep") as sleep:
        sleep.side_effect = remove_lock_file   # Remove the lock file instead of sleeping
        
        assert "unlock" in repr(execution_lock)
        cmd = ["git", "status"]
        kwargs = {"cwd": f"{tmp_path!s}"}
        await execute(cmd, **kwargs)
        assert "unlock" in repr(execution_lock)
        
        assert not lock_file.exists()
        sleep.assert_called_once()
```

mlucool: I was avoiding using any disk, but if you think thats ok I'm happy to go with your suggestion. 

FWIW, per above, I don't think git status modifies the lock and we may want to let get status skip having to wait in the future if it does cause some slowness

fcollonval: > I was avoiding using any disk, but if you think thats ok I'm happy to go with your suggestion.

Ja, it is fine. At some point, we should start integration tests those will be much slower.
 
> FWIW, per above, I don't think git status modifies the lock and we may want to let get status skip having to wait in the future if it does cause some slowness

You are correct. I updated the suggestion to restore the `git dummy` call. The goal is not here to check if the command executed correctly or not. So this is fine.

kgryte:
It is not clear to me that we should continually retry until command success in the event that an `index.lock` is already present, as opposed to parsing error messages and displaying a dialog in the UI which instructs a user to retry the command after waiting a specified period of time.

The main issue I see with this PR is the possibility of a race condition. We don't "lock" the UI to prevent further commands, and, from my [reading](https://github.com/jupyterlab/jupyterlab-git/pull/564/files#diff-041cb4cfb7143f3bbcdd07d28dffa92cR107), we run the risk that the pending command could be executed **after** a subsequent command (e.g., another commit).

If we want to add retry logic, we should presumably lock the UI to prevent further interactions. IMO, that is not particularly desirable, in comparison to pushing the retry to the user, which is exactly what happens when using Git at the command-line.

mlucool:
> It is not clear to me that we should continually retry until command success in the event that an index.lock is already present

We only loop on waiting for the lock, not rerunning the command

> we run the risk that the pending command could be executed after a subsequent command (e.g., another commit).

This enforces that each command wait for `execution_lock.acquire`. I assumed, maybe wrongly, they would acquire the lock in FIFO and thus preserve ordering. If this is not true, we should enforce this somehow.

The reason for this commit has less to do with needing retry, but because some of the action from the UI require multiple commands to be sent (e.g. a `git add` then a `git commit`)

kgryte:
@mlucool Yes, you are correct; from my reading, `tornado` uses something akin to a FIFO queue. In which case, the race condition I alluded to earlier is not applicable.

However, another concern is what happens if a user shuts down a kernel before an `index.lock` file is removed and the queued command/commands has/have executed?

Meaning, is it possible that, e.g., one or more pending commands (e.g., one or more commit commands for "saving" changes) could fail to be executed in the event that the kernel closes during the interim waiting for an `index.lock` file to be removed?

If, instead of locking on the backend, where the locking behavior is hidden from the user, the extension erred immediately and displayed an error modal to the user, recommending manually retrying an action, then, IMO, while arguably more laborious, expectations could be better managed.

In short, I am largely concerned about all the ways that this change could cause the extension to fail. Are there edge cases that we are missing?

Lastly, I am leery about hardcoding intervals for wait times. Wait times would seem to be highly particular to a user's environment and the relevant repository.

For example, one of my side projects is rather large, and I frequently encounter the presence of an `index.lock` which sometimes takes upward of 30-60 seconds to disappear, depending on the current resource constraints on my laptop.

Personally, I don't have a good sense as to what could be considered reasonable.

mlucool:
> However, another concern is what happens if a user shuts down a kernel before an index.lock file is removed and the queued command/commands has/have executed?

I thought extension added routes and interactions are not done via a kernel?

All actions that take more than one command which affects `index.lock` (e.g. git add + commit in simple staging mode) fail because they happen too fast on NFS in my case. These would have to be made atomic from the UI POV and the backend would have to enforce waiting and ordering between those. I'll also note that having more than one UI instance means you can have race conditions between them and is another reason to put it into the backend.

> Lastly, I am leery about hardcoding intervals for wait times. Wait times would seem to be highly particular to a user's environment and the relevant repository.

You are right. We should let users change these. There is no "right" amount of time to wait with very different environments. I think we pick a good default in terms of human timeframes and then let advanced users change this. Should we both `MAX_WAIT_FOR_EXECUTE_S` and `MAX_WAIT_FOR_LOCK_S `?

saulshanabrook:
> These would have to be made atomic from the UI POV and the backend would have to enforce waiting and ordering between those. 

It seems to make sense that if we are allowing a long running process to take place on the server, then we should make sure to make this clear to the user. So we should add some way for them to see what actions are being run on the server, or block the UI while one action is being run.

If that could be done after this PR is merged, instead of in this PR itself, that seems like it could be fine as well.

mlucool:
> So we should add some way for them to see what actions are being run on the server, or block the UI while one action is being run.

This already happens for at least some commands (a popup with a spinner). I agree that we should have feedback for all commands to make it more clear to a user that something is being done.

FWIW, in practice, this wait is still subhuman scale for the cases I tested (fast network, NFS, smallish repo). Any case that would be on human scales (e.g. very large repos that @kgryte pointed out), would either fail (e.g. `git add` and `git commit` would step on each other) or take approximately as long as they would from the CLI.

kgryte:
@telamonian You have any thoughts on this?

Proposal:

1.  Modify PR to allow retry timeout configurability.
2.  Follow-up PR to suspend UI interactions until pending command complete (e.g., a modal with a spinner). I believe this should be straightforward, as we just need to enable a UI element until receiving an HTTP response.

The need for the latter is to prevent the user from closing JupyterLab before pending commands have had a chance to complete (e.g., before an `index.lock` file is removed and `git add && git commit` run, thus potentially leading to the discarding of user changes).

mlucool:
I'm happy with @kgryte proposal if that means @telamonian / @fcollonval will accept this PR. It's been stuck for a while, and it would be good wrap up the work here.

fcollonval:
Thanks @mlucool for the patience. I'll merge this and add an issue with the proposal of enhancement from @kgryte.

fcollonval:

PR #569, opened 2020-03-23
Display message when server extension unavailable

lresende:
When, for some reason, the jupyterlab_git jupyter server_extension
is not installed or available the user experience was just receiving a
json response error message. 

This pr extends the code around the first invocation to the backend
api and add better error handling around 404 errors to provide a better
user experience.

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -12,6 +12,7 @@ import { ISignal, Signal } from '@phosphor/signaling';
 import { httpGitRequest } from './git';
 import { IGitExtension, Git } from './tokens';
 import { decodeStage } from './utils';
+import { Dialog, showErrorMessage } from '@jupyterlab/apputils';
 
 // Default refresh interval (in milliseconds) for polling the current Git status (NOTE: this value should be the same value as in the plugin settings schema):
 const DEFAULT_REFRESH_INTERVAL = 3000; // ms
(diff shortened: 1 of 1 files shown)


karlaspuldaro:
Changes look good. 
Tested locally by uninstalling the extension and got the error dialog:
image

src/model.ts
```diff
@@ -1014,10 +1020,20 @@ export class GitExtension implements IGitExtension {
   private async _getServerRoot(): Promise<string> {
     try {
       const response = await httpGitRequest('/git/server_root', 'GET', null);
+      if (response.status === 404) {
+        throw new ServerConnection.ResponseError(
+          response,
+          'Required Git serverextension not available.'
```
kgryte: What about expanding the error message a bit to provide guidance as to next steps?

```suggestion
          'Git server extension is unavailable. Please ensure you have installed the JupyterLab Git server extension by running: pip install --upgrade jupyterlab-git. To confirm that the server extension is installed, run: jupyter serverextension list.'
```

lresende: Perfect, updated the message.

telamonian:
This is excellent, thanks for contributing it. Should hopefully make for a much smoother user experience when the serverextension isn't set up correctly

PR #571, opened 2020-03-24
Fix style of FileList and FileItem

### LINKED ISSUE #572: Flaws in FileList styling related to height and scrollbars
## Description

Currently the styling of the FileList component is a bit flawed. For example, it doesn't expand to fill available vertical space:

image

Also, the scrollbar next to FileList also always shows, regardless of whether there are enough items to need a scrollbar.

## Reproduce

NA

## Expected behavior

NA

## Context

NA

telamonian:
fixes #572

Also fixes some issues with FileItemSimple styling by merging it with FileItem. Here's what the file items look like with this PR in "simple staging" mode:

image

@kgryte @lresende Can one of you please review this?


src/components/FileItem.tsx
```diff
@@ -27,11 +28,18 @@ export const STATUS_CODES = {
...
   actions?: React.ReactElement;
-  contextMenu: (event: React.MouseEvent) => void;
+  contextMenu?: (event: React.MouseEvent) => void;
   file: Git.IStatusFile;
+  markBox?: boolean;
   model: GitExtension;
-  selected: boolean;
-  selectFile: (file: Git.IStatusFile | null) => void;
+  selected?: boolean;
+  selectFile?: (file: Git.IStatusFile | null) => void;
```
fcollonval: Would it be better to set an function doing nothing rather than making `contextMenu` and `selectFile` optional?

telamonian: My current mental model of FileItem is as an object with configurable support for optional features. For example, in the Simple Staging case it may never make sense for a FileItem to have selection support. Thus, it makes sense for eg `selectFile` to be optional, rather than require a dummy

PR #575, opened 2020-03-25
Fix status error on empty string

### LINKED ISSUE #574: Error on status in CI
```
    Traceback (most recent call last):
      File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tornado/web.py", line 1703, in _execute
        result = await result
      File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/jupyterlab_git/handlers.py", line 131, in post
        result = await self.git.status(current_path)
      File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/jupyterlab_git/git.py", line 261, in status
        "x": line[0],
    IndexError: string index out of range

The command "python tests/test-browser/run_browser_test.py" exited with 1.
```

_Originally posted by @fcollonval in https://github.com/jupyterlab/jupyterlab-git/pull/573#issuecomment-603733675_

This was hit by @telamonian too on the latest PR merge: https://travis-ci.org/github/jupyterlab/jupyterlab-git/builds/666657840?utm_source=github_status&utm_medium=notification

fcollonval:
Fix #574

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -101,16 +101,12 @@ async def execute(
 
     return code, output, error
 
-def strip_and_split(s: "str") -> "List[str]":
-    """Strip trailing \x00 and split on \x00 a string.
 
+def strip_and_split(s):
+    """strip trailing \x00 and split on \x00
     Useful for parsing output of git commands with -z flag.
     """
-    s = s.strip("\x00")
-    if len(s) > 0:
-        return s.split("\x00")
-    else:
-        return list()
+    return s.strip("\x00").split("\x00")
 
 
 class Git:
@@ -260,7 +256,7 @@ class Git:
             }
 
         result = []
-        line_iterable = iter(strip_and_split(my_output))
+        line_iterable = line_iterable = (line for line in strip_and_split(my_output)) if line)
         for line in line_iterable:
             result.append({
                 "x": line[0],
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -101,12 +101,17 @@ async def execute(
 
     return code, output, error
 
-def strip_and_split(s):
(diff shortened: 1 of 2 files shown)


telamonian:
This is an intermittent bug that only shows up in the CI; I haven't been able to reproduce this bug locally. Before I merge this PR, I'd like to first understand the underlying cause of the bug a little better. Under what conditions does `git status --porcelain -u -z` have the problematic output?

fcollonval:
Unfortunately I was not able to find the conditions precisely.

But what is at least bad in `strip_and_split` is that it return a list containing a empty string when passing a empty string rather than a empty list. And this is the reason of the error seen in the CI.

This is the test added in this PR.

fcollonval:
Ok I just did some additional test on windows. If you open the extension on a clean git repository (nothing untracked or modified), you will hit the error. In the UI, nothing happens (as expected) so this may have silence an earlier discovery.

The error is due to the empty string being return in such case:
```
In [5]: r = subprocess.Popen(["git", "status", "-z", "-u", "--porcelain"], stdout=subprocess.PIPE)

In [6]: r.communicate()
Out[6]: (b'', None)
```

telamonian:
Since this is an issue specific to our status command, can you change your fix to only affect the `Git.status` method? I was thinking something like reverting `strip_and_split` to head and then in `Git.status`:

```python
line_iterable = (line for line in strip_and_split(my_output)) if line)
```

fcollonval:
I try to challenge the other use of `strip_and_split`. With the git commands used, I indeed got an error for the `diff` and `log` cases. However one thing that we should keep in mind is that a direct `git diff` on a new repository is a valid call and returned also an empty string. But as the calls of that commands are all setting some options requiring a revision to be present, we should be safe.

PR #579, opened 2020-03-26
Switch sponsorship icon

saulshanabrook:
Switches sponsorship icon to use handshake with money https://www.flaticon.com/free-icon/sponsorship_1999134?term=sponsorship&page=1&position=2


saulshanabrook:
I think the github preview is incorrect, because hopefully when this is merged it will pick up the relative URL properly.

This is how it looks locally:

Screen Shot 2020-03-26 at 3 26 01 PM

saulshanabrook:
Eh, it doesn't look so good that small! Let's try 🤝 instead.

saulshanabrook:
This is how it looks now, after the change.

Screen Shot 2020-05-25 at 2 41 23 PM

Does anyone have objections to this new icon?

Issue #578, opened 2020-03-26
Detailed log failed when changed files include binary files


fcollonval:
## Description

Fail to get detailed log when there are binary files involved.

![image](https://user-images.githubusercontent.com/8435071/77625221-7c78b980-6f43-11ea-9998-a4daef9a0efd.png)

When dealing with binary files, `git log --numstat -z --oneline` return `-` as number of changes => conversion to integer fails:

```
jupyterlab_git/git.py, line 350, in detailed_log
total_insertions += int(insertions)
ValueError: invalid literal for int() with base 10: '-'
```

See git command output below.

## Reproduce

Request detail log in history for commit 6db57bf (update animated demo gif in README to show 0.10.0 UI)

## Context

Tested with the master on Binder

- Python package version:
09c6533
- Extension version:
09c6533

<details><summary>Git Command Output</summary>
<pre>
jovyan@jupyter-jupyterlab-2djupyterlab-2dgit-2d97r8agpd:~$ git log -1 --numstat -z --oneline 6db57bf
6db57bf (tag: v0.10.0) update animated demo gif in README to show 0.10.0 UI^@2  2
       README.md^@-    -       docs/figs/demo-0-10-0.gif^@-    -       ^@gifs/nbdiff.gif^@docs/figs/nbdiff.gif^@
</pre>
</details>

<details><summary>Request answer</summary>
<pre>
​{"message": "Unhandled error", "reason": null, "traceback": "Traceback (most recent call last):\n  File \"/srv/conda/envs/notebook/lib/python3.7/site-packages/tornado/web.py\", line 1699, in _execute\n    result = await result\n  File \"/srv/conda/envs/notebook/lib/python3.7/site-packages/jupyterlab_git/handlers.py\", line 170, in post\n    result = await self.git.detailed_log(selected_hash, current_path)\n  File \"/srv/conda/envs/notebook/lib/python3.7/site-packages/jupyterlab_git/git.py\", line 350, in detailed_log\n    total_insertions += int(insertions)\nValueError: invalid literal for int() with base 10: '-'\n"}
</pre>
</details>

PR #580, opened 2020-03-27
Allow binary files in detailed log + update tests

ianhi:
This should fix: https://github.com/jupyterlab/jupyterlab-git/issues/578 as the only non-numeric characters expected should be hyphens (https://git-scm.com/docs/git-log#Documentation/git-log.txt---numstat)

> Similar to --stat, but shows number of added and deleted lines in decimal notation and pathname without abbreviation, to make it more machine friendly. For binary files, outputs two - instead of saying 0 0.

I also added trailing `\x00` to input for tests to accurately reflect what  `git ___ -z` outputs, as well as added a binary file test case to detailed log.

Finally I cleaned up the double definition of `line_iterable` in status

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -330,6 +330,8 @@ class Git:
         line_iterable = iter(strip_and_split(my_output)[1:])
         for line in line_iterable:
             insertions, deletions, file = line.split('\t')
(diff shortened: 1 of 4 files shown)


fcollonval:

Issue #584, opened 2020-04-01
Git Init Menu


mlucool:
The menu has "Git Init" in it.  I suggest two separate things:
1. Remove git init when inside a git repo already (its rare to have one git repo int another)
2. Allow users, via a config option, to change the default command from `git init` to something else.

The reason for 2, is I typically want `git init && setup_notebook_filters`.

What do people think?

fcollonval:
>     Remove git init when inside a git repo already (its rare to have one git repo int another)
👍 
I'm not a big fan of hiding UI element. I would suggest disabling the command in such case as done for the `add remote` command:
https://github.com/jupyterlab/jupyterlab-git/blob/2787178ecdf50a8adb3ed576ea03b6de4feddef5/src/gitMenuCommands.ts#L118

>     Allow users, via a config option, to change the default command from `git init` to something else.
👎 
In #494, it was decided to suppress the ability to run custom commands in terminal to follow the rules of JupyterLab. So I vote to reject this proposal for the same security reason.

mlucool:
> In #494, it was decided to suppress the ability to run custom commands in terminal to follow the rules of JupyterLab. So I vote to reject this proposal for the same security reason.

That is a good point. Do you have any suggestions? What if we add the ability to set in in the server config (`jupyter_notebook_config.py`)? I think that should mitigate the security concern. 

My use case is for init to setup [nbstripout](https://github.com/kynan/nbstripout) (update .gitattibutes) and on clone to rerun the setup to re-add to the git filters (they are not stored for security reasons). So I'd want some way to say when you init or clone - run this command (or run this command after init or clone is fine too).

jaipreet-s:
> Remove git init when inside a git repo already (its rare to have one git repo int another)

👍 as well. For reference, the _Git Clone_ button used to be disabled if already inside a Git repo, to discourage nesting of Git repos #260,  though I'm not sure if this is still the case after the recent refactoring)

fcollonval:
There is a hanging feature request for cleaning notebooks (https://github.com/jupyterlab/jupyterlab-git/issues/392). So it may be better to introduce this feature specifically through an extension option - I imagine that using a tooglable menu item would be appropriate.

There is also some issues with `nbdime` and `nbstripout` on Windows (see https://github.com/jupyterlab/jupyterlab-git/issues/471#issuecomment-578417925).
[...]

PR #596, opened 2020-04-18
add clone to commands and to git menu

ianhi:
I added the the clone command to the git menu:
image

If `init` is in this menu then I think it makes sense to also include `clone`. Also, making `clone` available as a command will make it easier to incorporate a clone button to the updated not in repo panel.

### DIFF
--- a/src/gitMenuCommands.ts
+++ b/src/gitMenuCommands.ts
@@ -147,7 +147,7 @@ export function addCommands(
       }
     }
   });
-  
+
   /** Add git clone command */
   commands.addCommand(CommandIDs.gitClone, {
     label: 'Clone',
--- a/src/gitMenuCommands.ts
+++ b/src/gitMenuCommands.ts
@@ -152,6 +152,7 @@ export function addCommands(
   commands.addCommand(CommandIDs.gitClone, {
     label: 'Clone',
     caption: 'Clone a repository from a URL',
+    isEnabled: () => model.pathRepository === null,
     execute: async () => {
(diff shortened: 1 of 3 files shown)


kgryte:
This PR may be affected by the outcome of https://github.com/jupyterlab/jupyterlab-git/issues/584.

ianhi:
Looks as though the conclusion there is that commands like `clone` ought to be disabled when in a git repository, this would also bring it into line with the behavior of the clone button. I will add this.

ianhi:
I also noticed that the git clone button above the filebrowser seems to only have a theme for a light jupyterlab theme. I actually thought the enabled version was disabled when I first saw it on account of how dark it was. Here it is in dark mode when:
enabled:
image
Disabled:
image

The plus and refresh file list have color: bdbdbd
and the new folder and upload icons have color: e0e0e0

How about adding a dark mode git clone icon with one of those colors?

fcollonval:
> How about adding a dark mode git clone icon with one of those colors?

Since JLab 1.2, JupyterLab has moved to svg inline icon. So styling icon is done via a class name in the svg for easier and better theming.

If you are interested, you could look at the following commits for an example with the diff icon:
https://github.com/jupyterlab/jupyterlab-git/pull/550/commits/d3b7fdda200a4866dc7c6657b130e2634f9ef8a0
https://github.com/jupyterlab/jupyterlab-git/pull/550/commits/5501a2844611af15bf787f6d8db1431de2342af3

fcollonval:

Issue #310, opened 2019-03-25
git merge


bmll-chris:
Add functionality to be able to git merge branches by selecting from list of available branches.

Zhanhao527:
Is this functionality added

jbacheli:
+1

This would be an interesting feature to minimise conflicts for users who are not comfortable with command lines.

PR #600, opened 2020-04-19
[WIP] Add merging functionality and UI

ianhi:
https://github.com/jupyterlab/jupyterlab-git/issues/310

This PR is an attempt to implement a merging UI similar to what github desktop has. I don't think it's complete but wanted to put it here for any feedback so that if I'm going in the wrong direction I can stop that sooner rather than later. In particular, before I go about writing tests and such. 

I drew pretty heavily from the github desktop UI, but didn't replicate their pre-emptive checking of if a merge is conflict free. Though this looks like it could be figured out with a series of git commands: https://stackoverflow.com/questions/501407/is-there-a-git-merge-dry-run-option

Here's the current UI, I think my one big worry is that was I was too invasive in modifying the existing `newBranchDialog` and should have just made a mergeDialog object. (To be clear the actual UI of the `newBranchDialog` is unchanged)

**Button in BranchMenu to trigger a dialog**:
image

**Dialog that pops up**:
Before a branch is selected:
image

After a branch is selected:
image

**merge conflict**:
The error reporting here could definitely use some work. I think using something other than the default error box will allow for better listing of the files that were the problem.

failed-merge

#### merge settings
I just wrote the merge command as `git merge branch` and didn't play arround with the other arguments like `-ff`. I also didn't try to extract the diff stat that merge ouputs in order to display it somewhere because this would have necessitated popping up another dialog.

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -732,6 +732,18 @@ class Git:
             return {"code": code, "command": " ".join(cmd), "message": error}
         return {"code": code}
 
+    async def merge(self, merge_from, top_repo_path):
+        """
+        Execute git merge command & return the result.
+        """
+        cmd = ["git", "merge", merge_from]
+        code, output, error = await execute(cmd, cwd=top_repo_path)
+
+        if code != 0:
+            return {"code": code, "command": " ".join(cmd), "message": error}
+        return {"code": 0}
+
+
     async def commit(self, commit_msg, top_repo_path):
         """
         Execute git commit <filename> command & return the result.
--- a/jupyterlab_git/handlers.py
+++ b/jupyterlab_git/handlers.py
@@ -367,6 +367,25 @@ class GitCheckoutHandler(GitHandler):
         self.finish(json.dumps(body))
 
 
+class GitMergeHandler(GitHandler):
+    """
+    Handler for git merge '<merge_from> <merge_into>'. Merges into current working branch
+    """
+    @web.authenticated
+    async def post(self):
+        """
+        POST request handler, merges branches
+        """
+        data = self.get_json_body()
+        merge_from = data["merge_from"]
+        top_repo_path = data["top_repo_path"]
+        body = await self.git.merge(merge_from, top_repo_path)
+
+        if body["code"] != 0:
+            self.set_status(500)
+        self.finish(json.dumps(body))
+
+
 class GitCommitHandler(GitHandler):
     """
     Handler for 'git commit -m <message>'. Commits files.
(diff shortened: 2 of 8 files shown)


kgryte:
I would suggest creating a separate dialog.

IMO one of the current design problems within this extension is that components tend to be shoehorned into serving multiple purposes, thus increasing code complexity and affecting maintainability.

My preference would be components which are dedicated to single purposes. Here, that purpose would be to provide merge functionality.

ianhi:
Thanks for the feedback, that makes sense to me. My reasoning for doing it this this way was to avoid replicating the filtering and branch selection and some of the styling. As this branch selection seems like the core shared feature of both of these dialogs.

 Would you suggest having an entirely separate widget that does its own listing and selection of branches? Or is it worth extracting parts that can be shared between the newBranch and merging dialogs? i.e. having a single component that displays lists of branches and incorporating that into both dialogs.

mlucool:
Notebooks aren't great with regular git merge. You may want to consider integrating nbdime for the notebook merge: https://nbdime.readthedocs.io/en/latest/#why-is-nbdime-needed

fcollonval:
> Would you suggest having an entirely separate widget that does its own listing and selection of branches? Or is it worth extracting parts that can be shared between the newBranch and merging dialogs? i.e. having a single component that displays lists of branches and incorporating that into both dialogs.

Smart laziness should drive developper work 😃 ; meaning make the code as modular as possible to avoid copy-paste (because you are copy-pasting bugs 😉). For ReactJS, this means composing component rather than inheriting (see the [documentation](https://reactjs.org/docs/composition-vs-inheritance.html)). So creating small reusable component is the way to go.

> Note: prefer functional component rather than ES6 classes for simple small component (e.g. [ActionButton](https://github.com/jupyterlab/jupyterlab-git/blob/master/src/components/ActionButton.tsx))

fcollonval:
> Notebooks aren't great with regular git merge. You may want to consider integrating nbdime for the notebook merge: https://nbdime.readthedocs.io/en/latest/#why-is-nbdime-needed

@mlucool I think @ianhi just want to provide a way to carry a simple merge. Resolving conflict is a though business (although it will be a great addition). It should be carry out in subsequent PRs.

> Note: technologies to have a resolve conflicts UI are already there for plain text (the code include the full merge add-on of code mirror) and notebooks (nbdime is included). So this should ease subsequent PRs for conflict resolution.

ianhi:
> provide a way to carry a simple merge

This is indeed my goal. 

You can configure git to use nbmerge as the merge driver for ipynb files. Unfortunately there does seems to be a way to do this as an argument to git merge. Instead you need to modify a .gitattributes file or `.git/config` ([ref](https://nbdime.readthedocs.io/en/latest/vcs.html#id3))

> To associate the merge driver with a file type, add the following entry to the appropriate .gitattributes file:
> 
>     *.ipynb merge=jupyternotebook
> 
Introducing changes to settings of a git repo seems a bit invasive. Maybe this extension could toggle this setting before doing a merge `git-nbmergedriver config --enable` and then disable it afterwards.

But I think it makes sense to limit this PR to just implementing standard git merge.

ianhi:
Oops. clicked the wrong button. didn't intend to close

PR #618, opened 2020-04-23
Update git-icon svg width/height size

### LINKED ISSUE #604: Left rail git icon is too tall
## Description

The git-icon used in the left rail of Jupyterlab is larger than its container.  The size can end up confusing Cypress automated tests which need to click other icons around the git icon.  By default, Cypress will not click an element if something else is on top of the element you want to click.  This was creating seemingly random test failures in our production canaries.  The screenshot below should help illustrate the issue:

Screen Shot 2020-04-21 at 3 52 48 PM

While this doesn't necessarily affect the user experience, it does make it harder to perform end-to-end tests on our application.

## Reproduce

1. pip install jupyterlab-git
2. jupyter serverextension enable jupyterlab_git
3. jupyter labextension install @jupyterlab/git
4. Load the page
5. Inspect the git icon in the left bar.  Make sure to select the `<svg>` element with browser's devtools.

## Expected behavior

I would expect the `<svg>` to have the same or smaller height than its wrapping `<div>`

## Context

- Python package version: 0.10.0

- Extension version: 0.10.0

kjdoyle:
In git-icon.svg it had hard coded width/height of 114.8625. Changed to standard size of 20.

Fixes #604

### DIFF
--- a/style/images/git-icon.svg
+++ b/style/images/git-icon.svg
@@ -10,8 +10,8 @@
    version="1.1"
    inkscape:version="0.91 r13725"
    xml:space="preserve"
-   width="114.8625"
-   height="114.8625"
+   width="20"
+   height="20"
    viewBox="0 0 114.8625 114.8625"
    sodipodi:docname="Git-Icon-Black.svg">
    <metadata id="metadata8">


ajbozarth:
After discussing this a bit with @kjdoyle over slack I believe this may be an issue that should be solved in jupyterlab. Currently lab only sets a width for the sidebar icons in css since each of the icons only sets the width in the svg file. Though we could solve this here by removing the height from this svg file (to mimic the lab ones). I would suggest that a better solution would be to add `height: 20px` to this line in jupyterlab since that's the height all the core lab icons in that tool bar default to. https://github.com/jupyterlab/jupyterlab/blob/4720c1121a306f3f77769a6c09a9f941ea8b034b/packages/ui-components/src/style/icon.ts#L391
@telamonian since you are the icons guys which is a better solution changing the svg here or adding height to the css in lab? I'll also add a suggested change to this PR if you believe it should be fixed here in the svg

style/images/git-icon.svg
```diff
@@ -10,8 +10,8 @@
    version="1.1"
    inkscape:version="0.91 r13725"
    xml:space="preserve"
-   width="114.8625"
-   height="114.8625"
+   width="20"
+   height="20"
```
ajbozarth, line 14: If @telamonian think this should be an svg fix here rather than addressing this in lab (as I mentioned in my previous comment) then I believe this would be a more consistent fix with icon in lab
```suggestion
   width="20"
```

kjdoyle: Thanks makes sense to me, but maybe width="16" to match the other icons that were all changed to that width. Will wait to hear about the overall handling though.

ajbozarth:
I opened an issue on lab to fix this after running into it in elyra. If we want to include a work around here we can just add the following css:
```
.lm-TabBar-tabIcon svg {
  height: auto;
}
```

Issue #617, opened 2020-04-23
Styling of filenames in error/warning dialogs


ianhi:
## Description
Currently in the Error Switch branch dialog or in the Discard warning dialog the filename is not visually distinct from the rest of the text.

![image](https://user-images.githubusercontent.com/10111092/80055326-68e35300-84ef-11ea-9b1e-3f9b92aca212.png)

![image](https://user-images.githubusercontent.com/10111092/80055036-b9a67c00-84ee-11ea-9586-4551c8acacb0.png)

### Expected behavior
The file would stand out so I can easily identify what file(s) are causing the problem/being discarded.

the `showErrorMessage` used in the branch error only accepts a string as input, but the `showDialog` used by discard can take a react element, so with a small change the filename can be made to stand out more:
```diff
diff --git a/src/components/FileList.tsx b/src/components/FileList.tsx
index 0d5c8cc..2b4a5d1 100644
--- a/src/components/FileList.tsx
+++ b/src/components/FileList.tsx
@@ -237,7 +237,7 @@ export class FileList extends React.Component<IFileListProps, IFileListState> {
   discardChanges = async (file: string) => {
     const result = await showDialog({
       title: 'Discard changes',
-      body: `Are you sure you want to permanently discard changes to ${file}? This action cannot be undone.`,
+      body: <span>Are you sure you want to permanently discard changes to <b>{file}</b>? This action cannot be undone.</span>,
       buttons: [Dialog.cancelButton(), Dialog.warnButton({ label: 'Discard' })]
     });
     if (result.button.accept) {
```
![image](https://user-images.githubusercontent.com/10111092/80055227-2faae300-84ef-11ea-806a-0fef8c30fa46.png)

fcollonval:
Don't hesitate to open a PR.

ianhi:
Done. I thought best practice was to open an issue first to allow for discussion? Is it ok to sometimes go straight to a PR?

fcollonval:
Thanks for the PR.

Yes opening an issue before a PR is the good practice. As the issue list is first place people will look before reporting something. So you did perfectly 👏

PR #619, opened 2020-04-23
improve error dialogs

ianhi:
close: https://github.com/jupyterlab/jupyterlab-git/issues/617

I added styling to the Discard dialog and the switching branch error dialog to make the filenames more visually distinct from the rest of the text. I also modified the text of a git checkout error to something that is hopefully clearer to beginners. In particular I followed the convention already used of saying "switch" instead of "checkout" and I added discarding the changes as an option for resolving the conflict.

**New:**
image

image

**Old:**
image

image

### DIFF
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -3,7 +3,7 @@ import { classes } from 'typestyle';
 import List from '@material-ui/core/List';
 import ListItem from '@material-ui/core/ListItem';
 import ClearIcon from '@material-ui/icons/Clear';
-import { showErrorMessage } from '@jupyterlab/apputils';
+import { Dialog, showDialog, showErrorMessage } from '@jupyterlab/apputils';
 import { Git, IGitExtension } from '../tokens';
 import {
   activeListItemClass,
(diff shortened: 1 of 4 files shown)


ianhi:
The new commit makes it so that the New Branch Dialog doesn't close too early. There were two things happening:
1. Currently if you don't enter a name and click "Create Branch" the dialog will just close without doing anything because `git checkout  ` doesn't do anything. The [documentation](https://git-scm.com/docs/git-checkout#_description) on this case is pretty fun:

> You could omit \<branch\>, in which case the command degenerates to "check out the current branch", which is a glorified no-op with rather expensive side-effects to show only the tracking information, if exists, for the current branch.

2. If you enter an invalid branch name such as `definitely invalid.lock` and try to create a branch the New Branch Dialog will close and then an error will be displayed explaining why a branch couldn't be created
invalid_branch_early_close

I did three things:
1. Disabled the create branch button when no name has been entered.
2. Close the New Branch Dialog only after it is clear that a new branch has been created
3. Removed the 'fatal:' from the error text as it wasn't adding more information to the error box.

Resulting in:
invalid_branch_later_closing

ianhi:
A failed git pull/push could also be improved:
Failed pull:
image

could offer a button in the dialog to run the `set-upstream-to` command
push:
image
vs: 
image

Unlike pull this is an edited version of the error text that git generates. Would be nice to prompt that a remote needs to be added.

fcollonval:

src/components/BranchMenu.tsx
```diff
@@ -314,7 +314,41 @@ export class BranchMenu extends React.Component<
...
      */
     function onError(err: any): void {
-      showErrorMessage('Error switching branch', err.message);
+      if (err.message.includes('following files would be overwritten')) {
+        showDialog({
+          title: 'Unable to switch branch',
+          body: (
+            <React.Fragment>
+              <p>
+                Your changes to the follow files would be overwritten by
```
ianhi, line 323: oops, just noticed this grammar error. In that this was just merged whats the best way to fix this?
```suggestion
                Your changes to the following files would be overwritten by
```

Issue #627, opened 2020-04-30
modifying a staged file does not add it to Changed files list


ianhi:
## Description
If you make changes to a file, stage it (`git add`) and then make further changes without re-staging then the file will only show up in the Staged list of the changes tab:
![image](https://user-images.githubusercontent.com/10111092/80665433-7f495b80-8a67-11ea-9cec-2211e94b504f.png)
This makes it so you can't see the diff of the uncommitted changes. But more importantly this makes the information in the GUI diverge from the information reported by the Git cli. The output of `git status` lists the file as both staged and unstaged:
![image](https://user-images.githubusercontent.com/10111092/80665207-d1d64800-8a66-11ea-988a-f381a427d286.png)

I think that the status for this file is being parsed correctly in python, but that the `decodeStage` function in typescript does not handle this case:
https://github.com/jupyterlab/jupyterlab-git/blob/8d56986409461cd9968d7208da4946fca8c70392/src/utils.ts#L37-L50

I think this function needs a 4th state it can return, perhaps: staged-and-unstaged
Then the diff of the changed file should be WORKING vs HEAD (maybe with an optional setting to show WORKING vs INDEX)
and for staged it should be INDEX vs HEAD

## Reproduce

1. add some content to `file.txt`
2. `git add file.txt`
3. make modifications to `file.txt`
4. Compare output of `git status` to the changes tab

## Expected behavior
The file appears in both the Staged and Changed lists.

## Context

- Python package version:
0.20.0rc0

- Extension version:
0.20.0-rc.0

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>
<pre>
Paste the output from your browser Javascript console here.
</pre>
</details>

PR #629, opened 2020-05-01
show both staged and unstaged changes for a file

ianhi:
Fixes: https://github.com/jupyterlab/jupyterlab-git/issues/627

As discussed in the issue if you have both staged and unstaged changes to a file:
image

then the file will only show in the Staged Files list in jupyterlab-git. 

I added another option to the Status token that accounts for this state and modified the FileList and FileItem accordingly. Note how the same file appears in both Staged and Unstaged lists with different status codes, and these give different results for the diff.
staged-unstaged

When you select one instance of the file both list items are selected. I can't decide if this is a feature or a bug. But it seems that this didn't affected any of the context menu commands that depend on tracking which file is selected.

In simple staging mode it reduces to showing the diff from WORKING to HEAD:
image

### DIFF
--- a/src/components/FileItem.tsx
+++ b/src/components/FileItem.tsx
@@ -34,6 +34,7 @@ export interface IFileItemProps {
   onDoubleClick: () => void;
   selected?: boolean;
   selectFile?: (file: Git.IStatusFile | null) => void;
+  staged?: boolean;
 }
 
 export interface IGitMarkBoxProps {
@@ -74,9 +75,10 @@ export class FileItem extends React.Component<IFileItemProps> {
   }
 
   render() {
-    const status =
-      this.getFileChangedLabel(this.props.file.y as any) ||
-      this.getFileChangedLabel(this.props.file.x as any);
+    const status_code = this.props.staged
+      ? this.props.file.x
+      : this.props.file.y;
+    const status = this.getFileChangedLabel(status_code as any);
 
     return (
       <li
@@ -110,9 +112,7 @@ export class FileItem extends React.Component<IFileItemProps> {
         />
         {this.props.actions}
         <span className={this.getFileChangedLabelClass(this.props.file.y)}>
-          {this.props.file.y === '?'
-            ? 'U'
-            : this.props.file.y.trim() || this.props.file.x}
+          {this.props.file.y === '?' ? 'U' : status_code}
         </span>
       </li>
     );
--- a/src/components/FileItem.tsx
+++ b/src/components/FileItem.tsx
@@ -34,7 +34,6 @@ export interface IFileItemProps {
   onDoubleClick: () => void;
   selected?: boolean;
   selectFile?: (file: Git.IStatusFile | null) => void;
-  staged?: boolean;
 }
 
 export interface IGitMarkBoxProps {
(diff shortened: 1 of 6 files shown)


ianhi:
> Do you have time to work on this and to rebase on the master?

Yes, will try to do this evening.

ianhi:
What should Discard changes do to partially staged files? Currently is also discards staged changes?

ianhi:
Rebased. Not sure what to do about discarding changes. There is a fun failure mode where if you stage the creation of a file, make further changes, then revert those changes you get the below error:
image
image

I think this happens because the discard function first calls `reset` then tries to `checkout` a file that no longer exists.

fcollonval:
@ianhi thanks a lot for the quick action. I modify the `discardChanges` function to skip calling checkout in the case you mentioned. I also modify the case where the index contains modifications but the file was deleted in the working folder to be coherent.

Is this PR good to go for you like that?

ianhi:
Oops sorry I accidentally commented out your action button lines! Looking at this again there are two things I might change:

1. Put the discard changes action button on the left of the buttons. This will make the button positions line up for staged and unstaged files. 
2. Make discarding changes to a partially staged file only remove the unstaged changes.

My intuition as user is that discarding changes through a button in the unstaged section would only remove the the unstaged changes. So removing both staged and unstaged changes is dangerous because I can lose more changes than I expect.
discarding-changes

I think my position boils down to: Hitting discard changes on a FileItem in the unstaged section should call `git checkout` but not `git reset`. It looks like the `git reset` is in there because https://github.com/jupyterlab/jupyterlab-git/blob/bd872968e4f8a96357af8296294689f7a74288c3/src/components/FileList.tsx#L236 the function is meant to apply to both staged and unstaged files. Though there doesn't currently seem to be a way to unstage changes to a staged file. Can we just get rid of the `git reset` and change `DiscardChanges` -> `DiscardUnstagedChanges`?

Or perhaps discard changes should result in a [git restore](https://git-scm.com/docs/git-restore) command. Unfortunately that was only introduced in git version 2.23, But we can easily get the same effect with:
```bash
git stash push file
git stash drop
```

fcollonval:
> * Put the discard changes action button on the left of the buttons. This will make the button positions line up for staged and unstaged files.

I will not change it. The current order follows the one of VS code. And this seems to be ok if double clicking opens the diff view, then the user action for a given file will be adding or discarding the changes. So it sounds logical the two buttons are next to each other.

> * Make discarding changes to a partially staged file only remove the unstaged changes.

Done, staged file are reset only and unstaged file are checkout only.

We cannot get ride of the method as we need to cover the simple UI case in which discarding changes means reset + checkout for partially staged files (although this should not happen when using only the UI).

ianhi:
:tada: Just tried it out again and everything seems to work as expected, so good to go for me.

Thanks for the help on this one.

ianhi:
Oops wait I take that back, if the only staged file is a partially staged file then you will be unable to commit. I will fix and then I think it will be good to go

ianhi:
I also realized that there was no way to access the option to discard changes to a staged files in simple mode. So I added a discard action to files in simple staging mode and also added a contextmenu.

fcollonval:

PR #647, opened 2020-05-26
All icons as svg

### LINKED ISSUE #429: Use more svg icons for better theming
Continue work of #426. But it may be worth to wait for jlab 2.0 release that will simplify handling svg icons.

See @telamonian comment

I plan to open an issue to use more svg for icons in a further PR. But I'll put your comment in it to maybe delay that for JLab 2.0.

_Originally posted by @fcollonval in https://github.com/jupyterlab/jupyterlab-git/pull/426#issuecomment-544136921_

### LINKED ISSUE #604: Left rail git icon is too tall

fcollonval:
Use SVG for all icons with default width = 16px.

Naming has also be improved to be more coherent with syntax in JupyterLab core.

Superseeds #550 and #618

Fixes #429 and fixes #604


kgryte:
This has been discussed elsewhere. Moving to icons to embedded SVG has performance implications.

@fcollonval Have you investigated these performance implications? This could include time to paint, DOM updates, node selection, and reflows.

JLab already has performance concerns, and I'd like to avoid exacerbating the problem.

Notably, especially for repeated icons, such as in the branch menu and file list components, increasing the number of DOM nodes will impact DOM traversal and rendering performance.

I recognize that theming is a reasonable concern; however, we need to be pragmatic in ensuring that we're doing things wisely when balancing concerns,and not pursuing an implementation simply because JLab core does something--core's concerns may or may not perfectly overlap with extension concerns.

I would like see a more complete analysis of this PR's implications (e.g., as a function of the number of branches, changed files, etc) before green lighting this.

Perhaps I will be pleasantly surprised in terms of performance, but my experience in the web data visualization world suggests otherwise.

Perhaps you have already investigated performance implications in your icon work @telamonian?

fcollonval:
Thanks @kgryte for the comments.

I understand the concern. So an intermediate solution (as trade-off between theming and performance) could be to render all icons as SVG inline elements except the one for the files and branch lists. What do you think?

I read a bit about SVG icon and performance issues. I saw that Max already improved the situation in https://github.com/jupyterlab/jupyterlab/pull/8125 and that he has more idea https://github.com/jupyterlab/jupyterlab/issues/8237. So indeed let's wait for his comments.

Question for @telamonian : I read that [`<use>`](https://fvsch.com/svg-icons/) allows for more efficient reuse of svg inside a page. I did not follow the discussion on `LabIcon` but for what I saw in the current implementation. You are already specifying uuid for each icon. But you are using a clone-node action to replace icon. Would it be more efficient to load at construction time the LabIcon as `symbol` using the uuid as _id_ in the web page (keeping track of it if to apply changes if needed) and load the real icon using `use` to that _id_?

fcollonval:
Quick remark, I looked by curiosity at VS Code. They use font icons mostly (in particular for files listing) and some background images.

src/components/ActionButton.tsx
```diff
@@ -16,9 +16,9 @@ export interface IActionButtonProps {
    */
   disabled?: boolean;
   /**
-   * Icon name
+   * Icon
    */
-  iconName: string;
+  icon: LabIcon.IMaybeResolvable;
```
telamonian, line 21:
```suggestion
  icon: LabIcon;
```

All of the `ActionButton` instances are upgraded in this PR to actual `LabIcon` instances, so there's no need for `IMaybeResolvable` here.

In general, my advice would be to avoid `IMaybeResolvable` like the plague. Its use cases are very narrow

src/components/ActionButton.tsx
```diff
@@ -37,14 +37,15 @@ export interface IActionButtonProps {
...
-      className={classes(actionButtonStyle, props.className)}
-      title={props.title}
-      onClick={props.onClick}
+      disabled={disabled}
+      className={classes(actionButtonStyle, className)}
+      title={title}
+      onClick={onClick}
     >
-      <LabIcon.resolveReact icon={props.iconName} />
+      <LabIcon.resolveReact icon={icon} elementPosition="center" tag="span" />
```
telamonian, line 48:
```suggestion
      <icon.react elementPosition="center" tag="span" />
```

Don't bother with `resolveReact` here. You already know/can enforce via `IActionButtonProps` that all `icon` here are instances of `LabIcon`.

src/components/FilePath.tsx
```diff
@@ -38,7 +37,12 @@ export const FilePath: React.FunctionComponent<IFilePathProps> = (
 
   return (
     <React.Fragment>
-      <span className={getFileIconClass(props)} />
+      <LabIcon.resolveReact
```
telamonian, line 40: This is a nice use of `resolveReact`. Its use is appropriate here since either/both of `filetype.icon` and `filetype.iconClass` may be undefined

src/components/NewBranchDialog.tsx
```diff
@@ -278,7 +279,7 @@ export class NewBranchDialog extends React.Component<
         key={branch.name}
         onClick={this._onBranchClickFactory(branch.name)}
       >
-        <span className={classes(listItemIconClass, 'jp-Icon-16')} />
+        <branchIcon.react className={listItemIconClass} tag="span" />
```
telamonian, line 282: Athan made a convincing argument (on the basis of performance) as for why the branch icon should be a background image and not an inline svg here. What's the impetus for changing it now?

ping @kgryte

kgryte: I think @fcollonval was trying to be consistent, but IMO we should sacrifice dev consistency for performance concerns and revert to using background images where performance concerns are valid, such as here and anywhere where we can expect several duplicate icons (e.g., file list, etc).

src/components/ActionButton.tsx
fcollonval: Thanks for the tip

src/components/NewBranchDialog.tsx
fcollonval: Yep consistency was the reason. So I will move back to the original code in light of the potential performance issue.

fcollonval:
I updated this PR to use SVG icons everywhere except on file and branch lists. To have icon style change on those lists, I set a CSS filter - that does not conform to theme as SVG but this is at least better especially for the selected item.

fcollonval:
> Some users are facing performance issue for large number of modified files. That will most likely be solved by rendering only the subset of React component needed. Therefore we could also switch to use SVG icons everywhere as the number of DOM nodes will be kept under control.

telamonian:
> Some users are facing performance issue for large number of modified files. That will most likely be solved by rendering only the subset of React component needed.

If you're talking virtualization, I'm hard at work on the basis for a new virtualized [file listings widget](https://github.com/jpmorganchase/regular-table/blob/master/examples/file_browser.html). There's a live demo [here](https://bl.ocks.org/texodus/raw/a7b7588c899e3953dd8580e81c51b3f9/): functional, though not yet feature complete. To my shame the latest release runs only on chrome, but I have already submitted a firefox fix PR.

This will be production ready in some form or another in a month or two. Until then, the status quo of filetype icons as `background-image`s is fine

fcollonval:
> @fcollonval If you haven't already, can you please do one last spot check for visual regressions for this PR in both chrome and firefox? Once you give me the okay for that I'll pull this PR in

I have no seen visual regression. Here are some screenshots to compare.

| browser | theme | master | PR |
| --- | --- | --- | --- |
| Chrome | Light | image | image |
|  | Dark | image | image |
| Firefox | Light | image | image |
|  | Dark | image | image |

PR #649, opened 2020-05-27
Fix #551

### LINKED ISSUE #551: Problem with install in the build step
Having `jlpm install` in the [package.json's build](https://github.com/jupyterlab/jupyterlab-git/blob/master/package.json#L16) is somewhat problematic. I happened to have set `NODE_ENV=production yarn run build`, which yarn took to mean uninstall all my dev dependencies.

Not a high priority, but thought it was worth flagging.

fcollonval:
To have install + build, just execute `jlpm`.

As the clean command is not set in the _prepare_ stage, this avoid multiple compilation (the reason it was remove in 49e3261e9214b27441f0d0e1de7d1a0aa674a600).

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,10 +101,10 @@ cd jupyterlab-git
 
 # Install the server extension in development mode and enable it
 pip install -e .[test]
-jupyter serverextension enable --py jupyterlab_git
+jupyter serverextension enable --py jupyterlab_git --sys-prefix
 
 # Build the labextension and dev-mode link it to jlab
-jlpm build
+jlpm
 jupyter labextension link .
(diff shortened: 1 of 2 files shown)


kgryte:
Why `prepare` and not `postinstall`?

fcollonval:
> Why `prepare` and not `postinstall`?

I followed the practice for JLab extensions as established by the extension template.

kgryte:
We should update the README, as well.

fcollonval:
> We should update the README, as well.

Good catch, thanks

fcollonval:

PR #650, opened 2020-05-27
Test for `response.ok` not `status === 200`

### LINKED ISSUE #642: Add remote succeeds despite prompting an error message
The following line is wrong (and should never be used):

https://github.com/jupyterlab/jupyterlab-git/blob/e923a5beeb5f29ab16085e4a28efeae785128280/src/model.ts#L358

The post request returns the correct 201 status that results in prompting an error message when actually everything went well.

fcollonval:
Fixes #642


fcollonval:

PR #653, opened 2020-05-28
Duplicate prettier configuration

fcollonval:
Prettier configuration is done through file .prettierrc and .prettierignore.

This should be removed.


fcollonval:

PR #655, opened 2020-05-28
Ease release through Github workflow

fcollonval:
@telamonian I propose this PR to ease the release; i.e. it will release each time a GitHub release is published (and a git tag will be also created) - the version of npm and Python are tested to avoid publishing non-matching packages.


fcollonval:

Issue #640, opened 2020-05-19
Change the commit button title attribute to reflect why it is disabled


fcollonval:
Even if there was no error it might be nice to change the title attribute of the commit button to reflect why it is disabled. It can be disabled due to no text in commit message, or no files staged.
https://github.com/jupyterlab/jupyterlab-git/blob/e923a5beeb5f29ab16085e4a28efeae785128280/src/components/CommitBox.tsx#L70
https://github.com/jupyterlab/jupyterlab-git/blob/e923a5beeb5f29ab16085e4a28efeae785128280/src/components/CommitBox.tsx#L90-L97

_Originally posted by @ianhi in https://github.com/jupyterlab/jupyterlab-git/issues/639#issuecomment-630302921_

ianhi:
How about also changing the disabled styling of the button from grey to semi transparent? 
This is what jupyterlab does. 
https://github.com/jupyterlab/jupyterlab/blob/f0cb758333df2bb622689bcc8b92e63bb9c533c6/packages/apputils/style/toolbar.css#L86-L88

vs the current disabled styling:
https://github.com/jupyterlab/jupyterlab-git/blob/e923a5beeb5f29ab16085e4a28efeae785128280/src/style/CommitBox.ts#L102-L106

kgryte:
@ianhi I don't see the advantage of changing the transparency.

PR #669, opened 2020-06-07
Use title to describe why commit button is disabled

ianhi:
fix: https://github.com/jupyterlab/jupyterlab-git/issues/640

Add some conditional logic to choosing the title of the commit button so it is clear to users why it is disabled.


fcollonval:

Issue #628, opened 2020-04-30
Impossible to update an open diff


ianhi:
## Description
If you have a tab with the diff of a file and then make further changes to the file you will be unable to see these changes reflected in the diff until you close the diff tab and create a new one. 
![diff-not-updating](https://user-images.githubusercontent.com/10111092/80667105-38aa3000-8a6c-11ea-8c6b-cc21d6ef8039.gif)

Clicking the diff action button again does not update the diff on account of the checking done here:
https://github.com/jupyterlab/jupyterlab-git/blob/8d56986409461cd9968d7208da4946fca8c70392/src/components/diff/DiffWidget.tsx#L31-L35

## Reproduce

1. Make changes to a file
2. Open a diff of that file
3. make more changes
4. try to update the diff without closing the diff tab.

## Expected behavior
When clicking the Diff button the diff will be recomputed.

Two nice to have features would be:
1. Option to refresh the diff from a context menu when right clicking the diff widget
2. diffs that are visible are recomputed every so often, or whenever the diff is focused. 

## Context

- Python package version:
0.20.0rc0

- Extension version:
0.20.0-rc.0

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>
<pre>
Paste the output from your browser Javascript console here.
</pre>
</details>

kgryte:
Ideally, this would not be configurable. We should not display stale diffs.

jaipreet-s:
Agreed with @kgryte that this shouldn't be configurable. 

https://github.com/jupyterlab/jupyterlab-git/blob/8d56986409461cd9968d7208da4946fca8c70392/src/components/diff/DiffWidget.tsx#L31-L35

The intent of this line was to not open a new tab every-time the diff button is clicked. I do think we should retain that UX while addressing this to avoid an explosion of diff tabs.

 Now, if a `mainAreaItem` with an `id` is found, We could address this in two ways to ensure the content gets updated with the latest diff.

1. Close that one and create a new one its place, rather than activating it, or
2. Just update the content of the `mainAreaItem` and then activate it.

P.S. Thanks for all the recent reports and PRs @ianhi !

ianhi:
[...]

Issue #666, opened 2020-06-05
Adding diff provider not working


fcollonval:
Looking at the code following #664, I realized the API to add diff provider is actually not used and not working 😭

To be more precise, the following method is not populating a list used by the diff logic...

https://github.com/jupyterlab/jupyterlab-git/blob/6169af07d306eb07f3aa034e042a103a772a7649/src/model.ts#L727-L737

The code should be refractored to use it for notebook diff and documented in the readme. And this should definitely be possible.

> The API should be updated. The callback should return the react component to be displayed and handle by this extension not void => the API of that component should also be defined.

PR #671, opened 2020-06-08
Update Diffs when the underlying file is changed

ianhi:
Partial fix for: https://github.com/jupyterlab/jupyterlab-git/issues/628
update-plain-text-diff
Following @fcollonval's suggestion I used the `<UseSignal>` component to trigger an update of the diff when the file is changed. 

However, thus I only fixed the updating for non-notebook files. The `NbDiff` `PerformDiff` method contains a `SetState` call https://github.com/jupyterlab/jupyterlab-git/blob/6169af07d306eb07f3aa034e042a103a772a7649/src/components/diff/NbDiff.tsx#L202-L204

so calling it inside of `render` results in an infinite loop :(
I thought of three ways to approach this issue but I'm unsure if any of them are appropriate solutions. Any input would be most welcome:

1. Follow `PlainTextDiff` approach and hold the `nbmodel` outside of the state as a private variable.
    - my sense is that this is an incorrect way to use React.
2. Wrap `nbdModel` in an object that implements an `update` method that could be called.
3. Don't use the `UseSignal` component and instead set up a listener to the signal in the constructor of the individual diff providers. 
     - This puts more onus on each diff provider but would remove the recalculation of diffs from the `render` method.

### DIFF
--- a/src/components/FileList.tsx
+++ b/src/components/FileList.tsx
@@ -14,6 +14,7 @@ import { openDiffView } from './diff/DiffWidget';
 import { ISpecialRef } from './diff/model';
 import { FileItem } from './FileItem';
 import { GitStage } from './GitStage';
+import { FileBrowserModel } from '@jupyterlab/filebrowser';
 
 export namespace CommandIDs {
   export const gitFileOpen = 'git:context-open';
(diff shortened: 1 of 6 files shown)


fcollonval:
Thanks @ianhi for starting this.

As mentioned in #666, the diff provider logic should be reworked. And this PR shows that something needs to be structured to set it right. I'm keen to force diff provider to provide a `refresh` method in their component. This will only enforce an API on something that they need to code anyway.

Issue #668, opened 2020-06-07
Discard changes should update open editors


ianhi:
## Description
If you discard changes to a file you are currently working on then the editor doesn't update to reflect this.

![editor-not-updating](https://user-images.githubusercontent.com/10111092/83974037-48f9ca00-a8b8-11ea-9f73-c358ca13d0d9.gif)

If you discard and then save the file then jupyterlab detects the difference and asks if you want to overwrite or revert. I think that the revert option should be chosen automatically as  soon as the changes are discarded. Or, at the very least jlab-git should make this dialog open via `docmanager.save`
![discard-revert](https://user-images.githubusercontent.com/10111092/83974039-4b5c2400-a8b8-11ea-9130-02d5330c57fe.gif)

## Reproduce

1. open a file to edit
2. make some changes + save
3. use jlab-git to discard changes
4. editor doesn't reflect updates

## Expected behavior
Editor would change to reflect to reflect modifications to the file. I think this method is the one to use: https://github.com/jupyterlab/jupyterlab/blob/b1e2b83047421bf7196bec5f2a94d0616dcb2329/packages/docregistry/src/context.ts#L299-L306 though I don't know how to get the context for a an editor that may or may not be open.

## Context
Jupyterlab: 2.1.4

- Python package version: master brnach

- Extension version: master branch

ianhi:
I think the functions to call to get to the revert method are:
- IDocumentManager.findWidget https://jupyterlab.github.io/jupyterlab/docmanager/interfaces/idocumentmanager.html#findwidget
- get the context https://jupyterlab.github.io/jupyterlab/docregistry/interfaces/idocumentwidget.html#context
- revert https://jupyterlab.github.io/jupyterlab/docregistry/interfaces/documentregistry.icontext.html#revert

ianhi:
Similarly switching branches should update all the open editors for affected files

PR #676, opened 2020-06-13
Update open files when Git commands modify them

ianhi:
Close: https://github.com/jupyterlab/jupyterlab-git/issues/668

If any files are open in jupyterlab that are modified by:
- reset to commit
- revert commit
- switch branch
- discard changes

then they will be updated to reflect the new contents. Unless they have unsaved changes, in which case the open document will not be modified.

### Approach
Use the `changed_files` command to get the list of files changed by a git command and then revert those files if they are opened.

I didn't just loop over all open documents for two reasons:
1. I didn't think of doing that that until I started writing this just now
2. (Post hoc) This is is more parsimonious - if there are large notebooks that are unaffected by the Git command then it may be slow to reload them

### Tests
I updated the tests to mock the `changed_files` endpoint and mocked the `docmanager`. While my mock of the docmanager seems to work I'm not sure it's the best way to approach the mock. But unfortunately I'm not sure how I ought to do it differently. 

### In action

**Swapping branches**
revert-on-branch-change
**Discarding changes**

revert-on-discard
**Reverting to a commit**
revert-on-reset
**Reverting a specific commit**
revert-on-revert-commit

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -184,7 +184,7 @@ class Git:
 
         return response
 
-    async def changed_files(self, base=None, remote=None, single_commit=None):
+    async def changed_files(self, current_path, base=None, remote=None, single_commit=None):
         """Gets the list of changed files between two Git refs, or the files changed in a single commit
 
         There are two reserved "refs" for the base
@@ -204,7 +204,7 @@ class Git:
             }
         """
         if single_commit:
-            cmd = ["git", "diff", "{}^!".format(single_commit), "--name-only", "-z"]
+            cmd = ["git", "diff", single_commit, "--name-only", "-z"]
         elif base and remote:
             if base == "WORKING":
                 cmd = ["git", "diff", remote, "--name-only", "-z"]
@@ -219,7 +219,7 @@ class Git:
 
         response = {}
         try:
-            code, output, error = await execute(cmd, cwd=self.root_dir)
+            code, output, error = await execute(cmd, cwd=os.path.join(self.root_dir, current_path))
         except subprocess.CalledProcessError as e:
             response["code"] = e.returncode
             response["message"] = e.output.decode("utf-8")
--- a/jupyterlab_git/tests/test_diff.py
+++ b/jupyterlab_git/tests/test_diff.py
@@ -14,7 +14,7 @@ from .testutils import FakeContentManager, maybe_future
 async def test_changed_files_invalid_input():
     with pytest.raises(tornado.web.HTTPError):
         await Git(FakeContentManager("/bin")).changed_files(
-            base="64950a634cd11d1a01ddfedaeffed67b531cb11e"
+            current_path="test-path", base="64950a634cd11d1a01ddfedaeffed67b531cb11e"
         )
 
 
@@ -29,7 +29,7 @@ async def test_changed_files_single_commit():
 
         # When
         actual_response = await Git(FakeContentManager("/bin")).changed_files(
-            single_commit="64950a634cd11d1a01ddfedaeffed67b531cb11e"
+            current_path="test-path", single_commit="64950a634cd11d1a01ddfedaeffed67b531cb11e^!"
         )
 
         # Then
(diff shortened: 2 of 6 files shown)


ianhi:
I'm working on a rebase of this post #630 and have a quick question for either @fcollonval or @kgryte. Should the code that reverts the file should be inside the modal's try block? It seems to me the answer is yes, however if I do that then I will need to put the check on `response.ok` inside the try block which it looks as though @kgryte avoided. Is it a misuse of the `try..finally` to have return statements and error throwing inside the try block? For example I would be changing the code to look like this:
```typescript
    try {
      const files = (await this.changedFiles(null, null, commitId))['files'];
      const response = await httpGitRequest('/git/delete_commit', 'POST', {
        commit_id: commitId,
        top_repo_path: path
      });
      if (!response.ok) {
        return response.json().then((data: any) => {
          throw new ServerConnection.ResponseError(response, data.message);
        });
      }
      if (files) {
        files.forEach(file => {
          this._revertFile(file);
        });
      }
      await this.commit(message);
      return response;
    } catch (err) {
      throw new ServerConnection.NetworkError(err);
    } finally {
      this._removeTask(tid);
    }
```

vs this:
https://github.com/jupyterlab/jupyterlab-git/blob/4c3dbe2007a75eaea5dae13f1f020ff786fd7407/src/model.ts#L624-L640

kgryte:
@ian-r-rose I recommend **only** wrapping what is absolutely necessary in a `try/catch` block. Or stated otherwise, the statement (here, a network request) which could potentially throw an error. This simplifies error handling and mitigates the potential for conflated errors.

In the above, including substantial code blocks (i.e., many lines of code) increases the risk of trapping unintentional errors, such as syntax errors, etc, in which the application continues to run as normal due to the trapping or even responds with an erroneous error message making debugging more difficult.

ianhi:
@kgryte  thanks for the help. 

> Or stated otherwise, the statement (here, a network request) which could potentially throw an error. This simplifies error handling and mitigates the potential for conflated errors.

Do you mean that reverting the files should be outside the try block? That seems problematic because reverting files should (I think) be hidden behind the modal to prevent further user modifications while files are being modified. Alternatively I could move the revertFiles call outside of the try..finally and give it it's own modal, but then there would end up being two modal's spawned which does not seem ideal.

kgryte:
Ah, right. You could just move the `_removeTask` statement outside of the finally block. Sorry for not addressing the thrust of your question. Meaning,  if the modal needs to bridge multiple actions, just ensure that those actions happen between the `addTask` and `removeTask` statements. How you want to individually wrap the statements in-between is up to you. You can individually wrap them in `try/catch` blocks and then `removeTask` in consolidated error handling logic.

ianhi:
> You could just move the _removeTask statement outside of the finally block

Ah ok that sounds reasonable. I was worried that it was important to have `removeTask` in a finally block to make sure the modal is removed, but maybe that is not necessary?.

kgryte:
Nope. :) Just "bracket" the operations in which the modal should be present. For many of the operations for which I added the modal, there was only one operation for which we needed to show the modal. Hence, our ability to use `finally`. The main thing is that, during error handling, if, say, the first operation fails, you may need to early return and perform clean-up in the `catch` clause. In which case, you may have multiple `removeTask` statements, depending on how you handle failures.

fcollonval:
A small note about future change. I plan to push a proposal to handle connection and response error directly within `httpGitRequest`. This will ease readability of the `model` code and remove plenty of duplicate statements.

Related to that @kgryte, would it make sense to create `decorator` kind of method to handle the `_addTask`/`_removeTask`:

```js
executeTask(name, fn, ...args) {
this._addTask(name);
  try{
    fn(...args)
  } finally {
    this._removeTask(name)
  }
}
```

That will simplify the code and improve its robustness. But in light of the problem expose here. It could be better to have:

```js
executeTask(name, fnAndArgs[]: Array<{fn, args[]}>) {
this._addTask(name);
  try{
    fnAndArgs.forEach({fn, args} => { fn(..args) });
  } finally {
    this._removeTask(name)
  }
}
```

kgryte:
Personally, I don't see the need for using decorators, particularly when I may want exact control over when tasks are added and removed from the queue.

Pushing some of the error handling logic into `httpGitRequest` seems fine to me, so long as we have a mechanism for bubbling errors up in to UI components (e.g., global event listeners, etc).

fcollonval:

PR #678, opened 2020-06-14
Add Git version and OS to issue template

### LINKED ISSUE #673: Add Git version and OS to issue template
A few recent issues have been caused by users having old versions of Git installed so having this information would be helpful. Checking the Git version could also be added as a troubleshooting step in the template. Asking for the OS may also be helpful as it is my understanding that there are sometimes speed differences in Git commands as a function of OS.

fcollonval:
Fixes #673

@ianhi could you review this one please?

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -8,6 +8,7 @@ Welcome! Before creating a new issue:
 * Look at the README *Troubleshooting* section
 * Search for relevant issues
 * Check that you have updated both the jupyterlab extension and the python package to the same version
+* Check that your git command has version 2 or higher (using `git --version`)
 -->
 
 ## Description
@@ -35,6 +36,9 @@ Welcome! Before creating a new issue:
 <!-- Results of `conda list jupyterlab-git` or `pip show jupyterlab-git` -->
 - Extension version:
 <!-- Results of `jupyter labextension list` -->
+- Git version:
+<!-- Results of `git --version` -->
+- OS:
 
 <details><summary>Command Line Output</summary>
 <pre>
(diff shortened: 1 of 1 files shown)


.github/issue_template.md
```diff
@@ -35,6 +36,9 @@ Welcome! Before creating a new issue:
 <!-- Results of `conda list jupyterlab-git` or `pip show jupyterlab-git` -->
 - Extension version:
 <!-- Results of `jupyter labextension list` -->
+- Git version:
+<!-- Results of `git --version` -->
+- OS:
```
ianhi, line 41:
```suggestion
- Operating System and version:
```
This is more explicit and follows the issue template used by jupyterlab
image

.github/issue_template.md
```diff
@@ -8,6 +8,7 @@ Welcome! Before creating a new issue:
 * Look at the README *Troubleshooting* section
 * Search for relevant issues
 * Check that you have updated both the jupyterlab extension and the python package to the same version
```
ianhi, line 10: If adding a hint for how to check git version then there should probably also be a hint for these. Or similar to these command the hint for `git --version` should only appear in the context section

.github/issue_template.md
```diff
@@ -8,6 +8,7 @@ Welcome! Before creating a new issue:
...
+* Check that your git command has version 2 or higher (using `git --version`)
```
ianhi, line 11: Better matching the phrasing of the previous bullet point
```suggestion
* Check that you have installed Git version 2 or higher (using `git --version`)
```

fcollonval:
> If adding a hint for how to check git version then there should probably also be a hint for these. Or similar to these command the hint for `git --version` should only appear in the context section 

As the commands are mentionned in the context section. I went for a simple introduction without commands.

fcollonval:
Thanks for the review @ianhi

Issue #686, opened 2020-07-06
Error push in branch with "/" character


charliwardbd:
In my repo if i create a repo that the name contains "/" character like "feature/testing" when i try to push i get the next error 

and the push never end. 

Im using jupyterlab-git==0.10.0

![Captura de pantalla 2020-07-06 a las 13 33 00](https://user-images.githubusercontent.com/24396580/86589178-786e2600-bf8d-11ea-8b51-4ed67fd0ce19.png)

E 11:27:14.282 LabApp] Uncaught exception POST /git/push?1594034834216 (172.17.0.1)
    HTTPServerRequest(protocol='http', host='localhost:8888', method='POST', uri='/git/push?1594034834216', version='HTTP/1.1', remote_ip='172.17.0.1')
    Traceback (most recent call last):
      File "/usr/local/lib/python3.5/dist-packages/tornado/web.py", line 1592, in _execute
        result = yield result
      File "/usr/local/lib/python3.5/dist-packages/tornado/gen.py", line 1133, in run
        value = future.result()
      File "/usr/lib/python3.5/asyncio/futures.py", line 274, in result
        raise self._exception
      File "/usr/lib/python3.5/asyncio/tasks.py", line 239, in _step
        result = coro.send(None)
      File "/usr/local/lib/python3.5/dist-packages/jupyterlab_git/handlers.py", line 441, in post
        current_path, current_local_branch
      File "/usr/local/lib/python3.5/dist-packages/jupyterlab_git/git.py", line 900, in get_upstream_branch
        error, " ".join(command)
    Exception: Error [fatal: no such branch: 'prueba'
    ] occurred while executing [git rev-parse --abbrev-ref prueba@{upstream}] command to get upstream branch.
[E 11:27:14.283 LabApp] Uncaught exception in write_error
    Traceback (most recent call last):
      File "/usr/local/lib/python3.5/dist-packages/tornado/web.py", line 1592, in _execute
        result = yield result
      File "/usr/local/lib/python3.5/dist-packages/tornado/gen.py", line 1133, in run
        value = future.result()
      File "/usr/lib/python3.5/asyncio/futures.py", line 274, in result
        raise self._exception
      File "/usr/lib/python3.5/asyncio/tasks.py", line 239, in _step
        result = coro.send(None)
      File "/usr/local/lib/python3.5/dist-packages/jupyterlab_git/handlers.py", line 441, in post
        current_path, current_local_branch
[...]

PR #422, opened 2019-10-13
Test on multiple python 3.x

fcollonval:

PR #691, opened 2020-07-10
FIx pushing when there are slashes in branch names or in remote name

ianhi:
Fixes: https://github.com/jupyterlab/jupyterlab-git/issues/686

- Used an extra git command to allow for `/`s in both the branch and remote names
- tried to bring the returning of errors more in line with the rest of the extension
- Returned more info from the functions in `git.py` to allow the handlers/frontend to deal with rewriting the error messages as necessary.

I changed what is returned by `get_upstream_branch` but this is not used anywhere else in the extension so this shouldn't be a problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 python:  
+  - '3.5'
   - '3.6'
   - '3.7'
   - '3.8'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:  
-  - '3.5'
   - '3.6'
   - '3.7'
   - '3.8'
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -890,12 +890,12 @@ class Git:
         to the `branch` command to get the name.
         See https://git-blame.blogspot.com/2013/06/checking-current-branch-programatically.html
         """
-        command = ["git", "symbolic-ref", "HEAD"]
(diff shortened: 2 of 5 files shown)


ianhi:
oops - turns out you need to check all of the tests... I'll get back to this.

charliwardbd:
Thanks. I will test asap!

ianhi:
Some tests are failing due to python 3.5 not supporting f-strings. From: https://github.com/jupyterlab/jupyterlab-git/pull/422#issuecomment-590197671
>  > I'll fix the conflicts. Do we still want to test on Python 3.5 as it reached end of life and conda package are no more built for it?
>
> Makes sense to drop 3.5. Thanks.

I'll add that here

ianhi:
force pushed to pick up the icons as SVG changes that were causing test failures

ianhi:
Hooray the tests all pass now! What a journey that was.

One thing to note is that there was some special casing going on before for a remote branch that was in the same repository. I think my method should work for this but I wasn't actually able to test this on a real git repo because I couldn't figure out how to set it up. If anyone knows how to do that I will check it but several SO dives in I can't figure it out.

@charliwardbd if you're able to install the version in this PR it would be great if you could report whether it fixes you issue (You will need to be on JLab 2)

charliwardbd:
Sorry, i can't test it now. Actually i'm using version 1.2.6

fcollonval:
@all-contributors please add @ianhi for code

fcollonval:

Issue #674, opened 2020-06-12
Leaving git repo with branch selection menu open kills extension


ianhi:
## Description
Leaving a git repo while the branch menu is open will make the extension unusable:
![broken-branch-menu](https://user-images.githubusercontent.com/10111092/84471959-69c57500-ac54-11ea-9d11-85b97f685abb.gif)

In contrast if I close the branch menu before changing folders then everything works as expected:
![branch-menu-no-errors](https://user-images.githubusercontent.com/10111092/84472020-89f53400-ac54-11ea-80a9-6ffa889b81a1.gif)

## Reproduce

1. Create a git repo with two branches
2. open the branch selection dropdown
3. use jlab filebrowser to move out of git repo
4. extension no longer renders

## Context

- Python package version: master branch

- Extension version: master branch

Git version: 2.25.1
Browser: Firefox 75

Command Line Output: Nothing

<details><summary>Browser Output</summary>
<pre>
Please update the following components: CollapsibleSection react-dom.development.js:11494
The above error occurred in the <NewBranchDialog> component:
    in NewBranchDialog (created by BranchMenu)
    in div (created by BranchMenu)
    in BranchMenu (created by Toolbar)
    in div (created by Toolbar)
    in div (created by Toolbar)
    in Toolbar (created by GitPanel)
    in div (created by GitPanel)
    in GitPanel

Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://fb.me/react-error-boundaries to learn more about error boundaries. react-dom.development.js:19814
    React 12
    unstable_runWithPriority scheduler.development.js:643
    React 4
    unstable_runWithPriority scheduler.development.js:643
    React 6
    componentDidMount GitPanel.tsx:73
    invokeSlot index.es6.js:459
    emit index.es6.js:416
    emit index.es6.js:99
    _setStatus model.ts:1019
    refreshStatus model.ts:840
TypeError: "this.props.model.currentBranch is null"
    _renderItems NewBranchDialog.tsx:220
    render NewBranchDialog.tsx:190
    React 12
    unstable_runWithPriority scheduler.development.js:643
    React 6
    componentDidMount GitPanel.tsx:73
    invokeSlot index.es6.js:459
    emit index.es6.js:416
    emit index.es6.js:99
    _setStatus model.ts:1019
    refreshStatus model.ts:840
index.es6.js:211
    exceptionHandler index.es6.js:211
    invokeSlot index.es6.js:462
[...]

PR #692, opened 2020-07-10
check if in repo before rendering branch menu

ianhi:
Fixes: https://github.com/jupyterlab/jupyterlab-git/issues/674

Prevents the extension crashing when navigating away from a repo when the branch menu is open. I think the branchmenu was always being rendered even when not in a repo.

@kgryte could you review?


fcollonval:

PR #700, opened 2020-07-21
Custom actions

mlucool:
Fixes: https://github.com/jupyterlab/jupyterlab-git/issues/584

This adds the ability for users to set a custom class to take actions. As of now, we only have a post_init action, but this can be trivially extended to give more power to end users who have non-generic needs.

Example `~/.jupyter/jupyter_notebook_config.json`:
```
  "JupyterLabGitConfig": {
      "custom_git_actions_class": "org.dir.MyClass"
  }
```

### DIFF
--- a/jupyterlab_git/__init__.py
+++ b/jupyterlab_git/__init__.py
@@ -19,7 +19,7 @@ class JupyterLabGit(Configurable):
         help='Actions to be taken after a git command. Each action takes a list of commands to execute (strings). Supported actions: post_init',
         config=True,
         trait=List(
-            trait=Unicode,
+            trait=Unicode(),
             help='List of commands to run. E.g. ["touch baz.py"]'
         )
         # TODO Validate
@@ -35,7 +35,7 @@ def load_jupyter_server_extension(nbapp):
     """Load the Jupyter server extension.
     """
 
-    user_custom_actions = JupyterLabGit(config=nbapp.config).actions
-    git = Git(nbapp.web_app.settings['contents_manager'], user_custom_actions)
+    config = JupyterLabGit(config=nbapp.config)
+    git = Git(nbapp.web_app.settings['contents_manager'], config)
     nbapp.web_app.settings["git"] = git
     setup_handlers(nbapp.web_app)
--- a/jupyterlab_git/__init__.py
+++ b/jupyterlab_git/__init__.py
@@ -2,35 +2,27 @@
 """
 # need this in order to show version in `jupyter serverextension list`
 from ._version import __version__
-from traitlets import Type
+from traitlets import List, Dict, Unicode
 from traitlets.config import Configurable
 
 from jupyterlab_git.handlers import setup_handlers
 from jupyterlab_git.git import Git
 
-class GitCustomActions(object):
-    """
-    Class to handle custom actions. This is a base class that is expected to be replaced via the jupyter_notebook_config
-    """
-
-    async def post_init(self, cwd):
-        """
-        :param cwd: Directory where git init was run
-        """
-        pass
-
-class JupyterLabGitConfig(Configurable):
+class JupyterLabGit(Configurable):
     """
     Config options for jupyterlab_git
 
     Modeled after: https://github.com/jupyter/jupyter_server/blob/9dd2a9a114c045cfd8fd8748400c6a697041f7fa/jupyter_server/serverapp.py#L1040
     """
 
-    # See: https://traitlets.readthedocs.io/en/stable/trait_types.html#classes-and-instances
-    custom_git_actions_class = Type(
-        default_value=GitCustomActions,
+    actions = Dict(
+        help='Actions to be taken after a git command. Each action takes a list of commands to execute (strings). Supported actions: post_init',
         config=True,
-        help='Delegate custom git actions'
+        trait=List(
+            trait=Unicode,
+            help='List of commands to run. E.g. ["touch baz.py"]'
+        )
+        # TODO Validate
     )
 
 def _jupyter_server_extension_paths():
@@ -43,7 +35,7 @@ def load_jupyter_server_extension(nbapp):
     """Load the Jupyter server extension.
     """
 
-    user_custom_actions = JupyterLabGitConfig(config=nbapp.config).custom_git_actions_class()
+    user_custom_actions = JupyterLabGit(config=nbapp.config).actions
     git = Git(nbapp.web_app.settings['contents_manager'], user_custom_actions)
     nbapp.web_app.settings["git"] = git
     setup_handlers(nbapp.web_app)
--- a/jupyterlab_git/__init__.py
+++ b/jupyterlab_git/__init__.py
@@ -2,10 +2,36 @@
 """
 # need this in order to show version in `jupyter serverextension list`
 from ._version import __version__
+from traitlets import Type
+from traitlets.config import Configurable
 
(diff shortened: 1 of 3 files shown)


mlucool:
I'm add to add documentation to the README and test cases if this is the right design. Thought I'd bring this up for feedback first

fcollonval:
@mlucool Thanks for this PR. Using server configuration is safe enough - managed installation will keep control :wink: 

Regarding the implementation, I suggest to not be as generic for two reasons:
- We don't want to deal with complex custom Python code plugged-in
- We want to keep some control over those side-effects - like be able to implement a timeout feature around them.

So I propose to use the following configuration to a list of subprocess commands to be executed in the repository folder:

```js
{
  "JupyterLabGit": {
    "actions": {
      "postInit": [{
        "command": Array<string>, // subprocess command
        "args": Dict[string, string]  // subprocess args
      }]
    }
  }
}
```

Some comments on the schema:

- Use `JupyterLabGit` shorter name - we know that we are speaking about configuration
- Add a subentry `actions` so this will allow to organize the configuration parameters - if it grows in the future
- Add actions subentry with clear name about when it is gonna be executed (e.g. you could add a `postClone` entry)

What do you think?

> To make it works easily with `Configurable`, I suggest to use `AutoInstance` implemented in [jupyter-project](https://github.com/fcollonval/jupyter-project/blob/master/jupyter_project/config.py).

mlucool:
> Regarding the implementation, I suggest to not be as generic for two reasons:

I don't totally agree, but I'm happy to oblige. One key reason is that the API is very hard to establish - but maybe we just promise to give the script nothing and we call it in the right cwd. I also find in process error handling mostly easier. Still, we ca go with arbitary shell commands only, but I think we just support anything a user puts and use `shlex` to break it up or use the `shell=True` ( i.e. '/my/script.py --foo' is a valid command to pass).

> Some comments on the schema:

These all seem reasonable

fcollonval:
> but I think we just support anything a user puts and use shlex to break it up or use the shell=True ( i.e. '/my/script.py --foo' is a valid command to pass).

Good point, this will ease the setting.

mlucool:
@fcollonval updated. Any more feedback or I will finish cleaning this up (more doc, seeing if its easy enough to test). I can add this to places besides init (e.g. clone) as well.

mlucool:
@fcollonval did you have further feedback?

jupyterlab_git/__init__.py
```diff
@@ -16,6 +34,8 @@ def _jupyter_server_extension_paths():
 def load_jupyter_server_extension(nbapp):
     """Load the Jupyter server extension.
     """
-    git = Git(nbapp.web_app.settings['contents_manager'])
+
+    user_custom_actions = JupyterLabGit(config=nbapp.config).actions
+    git = Git(nbapp.web_app.settings['contents_manager'], user_custom_actions)
```
fcollonval, line 39: I would pass the all configuration object to anticipate other configuration settings.

```suggestion
    config = JupyterLabGit(config=nbapp.config)
    git = Git(nbapp.web_app.settings['contents_manager'], config)
```

jupyterlab_git/git.py
```diff
@@ -870,14 +872,33 @@ async def init(self, current_path):
...
+            actions_list = self.user_custom_actions[name]
+            for action in actions_list:
+                # We trust the actions as they were passed via a config and not the UI
+                code, stdout, stderr = await execute(
+                    shlex.split(action), cwd=cwd
+                )
+                # After any failure, stop
+                if code != 0:
+                    break
+        return code, stdout, stderr
```
fcollonval, line 900: Opened question: Should the stdout be concatenate for all actions?

jupyterlab_git/git.py
```diff
@@ -870,14 +872,33 @@ async def init(self, current_path):
         Execute git init command & return the result.
         """
         cmd = ["git", "init"]
+        cwd = os.path.join(self.root_dir, current_path)
         code, _, error = await execute(
-            cmd, cwd=os.path.join(self.root_dir, current_path)
+            cmd, cwd=cwd
         )
+        if code == 0:
+            code, _, error = await self._maybe_run_actions('post_init', cwd)
```
fcollonval, line 880: Opened question: Should stdout logged to help debugging?

jupyterlab_git/git.py
```diff
@@ -870,14 +872,33 @@ async def init(self, current_path):
...
+        code = 0
+        stdout = None
+        stderr = None
+        if name in self.user_custom_actions:
+            actions_list = self.user_custom_actions[name]
+            for action in actions_list:
+                # We trust the actions as they were passed via a config and not the UI
+                code, stdout, stderr = await execute(
+                    shlex.split(action), cwd=cwd
+                )
```
fcollonval, line 896:
```suggestion
                try:
                    code, stdout, stderr = await execute(
                        shlex.split(action), cwd=cwd
                    )
                except CalledProcessError:
                    break
```

Should we also protect against subprocess failure or do we consider it corrupt the base git action?

jupyterlab_git/__init__.py
mlucool: Addressed. Not I had to name it `_config` in `git.py` or a test was failing.

jupyterlab_git/git.py
mlucool: Do we have any examples of logging? For now, I sent all data (stdout/stderr/code/command) back to the front end where we can later show whatever we'd like to the user.

jupyterlab_git/git.py
mlucool: I caught all exceptions here and bailed.

mlucool:
@fcollonval this is ready for re-review.

> For the test, could you test also a failure of execute with a non-zero code and with the emission of an exception? This is quite critical and will help use ensuring robustness of the extension.

Tested no post_init, post_init that works, and post_init that fails

jupyterlab_git/git.py
```diff
@@ -872,13 +874,50 @@ async def init(self, current_path):
...
+                    actions.append({
+                        'cmd': action,
+                        'code': code,
+                        'stdout': stdout,
+                        'stderr': stderr
+                    })
+                    # After any failure, stop
+                except Exception as e:
+                    code = 1
+                    action.append({
```
fcollonval, line 911:
```suggestion
                    actions.append({
```

This is obviously not tested 😜

jupyterlab_git/tests/test_init.py
```diff
@@ -0,0 +1,88 @@
...
+        } == actual_response
+
+
+@pytest.mark.asyncio
+async def test_init_and_post_init_fail():
+    with patch("jupyterlab_git.git.execute") as mock_execute:
+        # Given
+        mock_execute.side_effect = [
+            maybe_future((0, "", "")),
+            maybe_future((1, "", "not_there: command not found")),
```
fcollonval, line 63: You can test the `Exception` emission by duplicating this test and set an Expection object as second element in the list.

jupyterlab_git/git.py
mlucool: Good catch!

jupyterlab_git/tests/test_init.py
mlucool: That tip was a big time saver - tested!

mlucool:
@fcollonval I believe all review feedback is taken care of

fcollonval:

Issue #703, opened 2020-07-24
Git status API recursively lists all untracked files causing browsers to hang


jaipreet-s:
## Description

The current implementation of git status uses the `-u` flag which recursively lists _all_ untracked files in _all_ sub-directories in a Git repository. Jupyter users often work with data sets that involve a large number of small files. When such datasets are untar'd into a working Git repository, this causes the `/git/status` API to return an unbounded response and freezes the entire JupyterLab front-end as well as the browser.

This operation is *not* paginated and the amount of files that are returned is unbounded.  For e.g., see the notebook [here](https://github.com/udacity/sagemaker-deployment/blob/master/Project/SageMaker%20Project.ipynb). This downloads the IMDb dataset. Downloading the data causes **the entire JupyterLab application to become unresponsive and the browser to hang**. I noticed my browser taking ~10GB of RAM and causing general machine slowness. 

## Reproduce

(Warning: This will hang your browser) 

Download and untar the [IMDB Dataset](http://ai.stanford.edu/~amaas/data/sentiment/) in your working directory.

Alternatively, 
1. `git clone https://github.com/udacity/sagemaker-deployment`
2. Run the `SageMaker Project` notebook in /Project

## Expected behavior

In general, any unpaginated approach is undesirable and the default behavior of the extension should provide sensible defaults that are performant. For e.g, in the `/git/all_history` API, the output from `git log` is intentionally restricted to 10 by default to avoid such scenarios (See https://github.com/jupyterlab/jupyterlab-git/blob/master/jupyterlab_git/git.py#L317-L325).  

Since there is no native pagination in the `git status` command and the adverse effects of this issue,  the proposed approach is to just use `-u='normal'` or omit the option altogether. (https://git-scm.com/docs/git-status#Documentation/git-status.txt---untracked-filesltmodegt). This option doesn't recursive list every untracked file in each sub-directory and still returns a reasonable default . This is the same UX as just doing `git status` in a terminal. Power users can fall back to the terminal to display each untracked file using additional flags in the API.
 
## Context

- Python package version: 0.10.1

- Extension version: 0.10.1

- Git version: 2.13

- Operating System and its version: Amazon Linux

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>
<pre>
Paste the output from your browser Javascript console here.
</pre>
</details>

[...]

PR #704, opened 2020-07-24
Don't track all untracked files in all subdirectories

jaipreet-s:
Please see #703 for more context around this issue before this PR.

The -u option has side-effects that make JupterLab unusable and hang user browsers.

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -243,7 +243,7 @@ class Git:
         """
         Execute git status command & return the result.
         """
-        cmd = ["git", "status", "--porcelain", "-u", "-z"]
+        cmd = ["git", "status", "--porcelain", "-z"]
         code, my_output, my_error = await execute(
             cmd, cwd=os.path.join(self.root_dir, current_path),
         )
--- a/jupyterlab_git/tests/test_status.py
+++ b/jupyterlab_git/tests/test_status.py
@@ -89,7 +89,7 @@ async def test_status(output, diff_output, expected):
         mock_execute.assert_has_calls(
             [
                 call(
-                    ["git", "status", "--porcelain", "-u", "-z"],
+                    ["git", "status", "--porcelain", "-z"],
                     cwd=os.path.join(root, repository),
                 ),
                 call(


ianhi:
Without the `-u` option git status will return some folders in the untracked files list. Does this break the display of untracked files and how do the current `git add` methods interact with them? 

image


It's definitely most important to solve the: jlab-git hangs the browser problem, but long term it would be nice to have the git UI have a better understanding of folders. i.e. allowing drop downs of a folder or committing an entire folder.

jaipreet-s:
> Without the -u option git status will return some folders in the untracked files list. Does this break the display of untracked files and how do the current git add methods interact with them?

I don't think so, as the -u option was added as part or #519 and the Git GUI has been functional since before that. I also checked locally and the UI shows directories in untracked, similar to what you get with `git status` on the terminal.

fcollonval:
I don't think this is the way to improve the situation. Not being explicit about files selection is really annoying - this is the reason -u was included. Moreover, it makes the advanced view inconsistent as all files will be shown as soon as you stage the folder. And as pointed be @ianhi this does not solve the issue if git is listing the folder files - like if you set a .gitkeep in the folder in which your data will be downloaded.

So a good fix will be to use a virtual file tree. And a fist improvement would be to unmount the FileList when the Git tab is not visible.

jaipreet-s:
Hey @fcollonval , thanks for checking in!

I'm not sure if you've had a look at the description of the issue in https://github.com/jupyterlab/jupyterlab-git/issues/703 but there are *adverse* side effects of including the `-u` option. I personally tried it and it froze up my browser completely and hanged my computer. I do recommend you try the reproduction steps to just gauge the impact.

There are clearly performance issues in the file tree but these are exacerbated by recursively listing all untracked files in a Git repository.  

1. It's easy for a user to get themselves into this situation. A common Jupyter user action is to download and untar a dataset. And just the act of doing that ends up hanging their browser in many cases.
2. Adding -u makes the situation N times worse because it will recursively list all files in all sub-directories.
3. -u is not paginated and will return unbounded lists.
4. The default behavior of `git status` is to _not_ list all untracked files, which is what this file browser is providing a UI over, and we need a good reason to stray from the default.

We have to provide sensible defaults to users. While I agree there is a larger discussion that needs to be happen around performance of the file tree, we need to provide guardrails so users don't get themselves into these situations (as Ian said _It's definitely most important to solve the: jlab-git hangs the browser problem,_) . I don't think we should include `-u` option until we have have tested the file tree's performance with large number of files and can handle these cases of unbounded lists.

Issue #683, opened 2020-06-24
Ignore Notebook UI


mlucool:
## Description

It's useful to declare some notebooks should not be stored in git. In git, this is accomplished by adding something to a `.gitignore` file. I propose we add an ability to right click on one or more files in the git extension and choose to ignore it. This would add a single line to `.gitignore`. To be clear, I am not proposing we commit the change - that's up to the user.

ianhi:
That seems useful. A set of options might look like:
- Add file to gitignore
- Add filetype to gitignore
- (for folders only) add folder to gitignore

kgryte:
@mlucool I am guessing you have in mind a "non-technical" JupyterLab user? Otherwise, for those familiar with Git and using `.gitignore` files, they could simply edit a `.gitignore` file directly in JLab to ignore a file, file type, and/or folder.

For the non-technical user, I am not sure if adding something to `.gitignore` but not committing the changed `.gitignore` is ideal. Suppose a user opts to ignore a folder via a context menu, but never commits the changes to the `.gitignore`. Later, they switch computers (work to home), create a folder which would have been ignored on the other computer, and it gets accidentally committed because they forgot to "ignore" it before staging and committing all files.

Or suppose a user ignores a file, never commits the changes to `.gitignore`, someone else collaborating on the same repository creates a file with the same name, commits it, and then the original user pulls changes and now has a clash and a potential undesired outcome.

Granted, the above can already happen when collaborating and directly using `.gitignore` files. My main hangup is that a user may be led to believe that, by clicking "ignore", a file, file type, and/or folder will be ignored forever, and, thus, be a bit surprised when they pull down changes and see a file type or folder they assumed would be ignored.

In general, while the above may be "edge" cases, I'd like to avoid creating UI behaviors which can lead to unexpected outcomes. I would think that, assuming we were to add support for ignoring files, file types, and/or folders, then we should automatically prompt a user to commit the change to `.gitignore`. This forces the user to be explicit and should help mitigate unexpected outcomes.

Also not clear to me how patterns should be automatically generated. If we are in a nested folder and a user opts to ignore all files having extension `*.foo`, then should the pattern added to the `.gitignore` file be based on the nested folder or apply to the entire repository? Similar for a file. Should only the file basename be added to the `.gitignore` file or should the complete relative path? Same thing for a folder. Just the folder name (in which case all matches folders are ignored) or the path?

[...]

PR #705, opened 2020-07-25
Git ignore UI

echarles:
Fixes https://github.com/jupyterlab/jupyterlab-git/issues/683

ezgif-4-6ad422d60c0f

This add a context `Ignore` menu for untracked files. When the user clicks on `Ignore`, the server is requested to append a line in the .gitignore file (will create it if it does not exist) and then the .gitignore file is opened and shown to the user.


mlucool:
Looks great!

Nitpick: I think there is an extra blank line between entries. Most projects [gitignore](https://github.com/jupyterlab/jupyterlab-git/blob/master/.gitignore) uses whitespace for grouping. In this case, I would consider them all one group.

echarles:
Thx @mlucool

yeah I don't like the extra blank line. It is there to avoid appending on the same last line if no `\n` is present at the end of the gitignore file. I was planning to be more clever in a second pass (should have mentioned that...)

I will wait a bit any feedback and will implement those for final review.

fcollonval:
Thanks a lot @echarles. This looks really nice. And if indeed you could improve the code to introduce the line return only if needed. This will be awesome.

echarles:
The las commits implement:

- A menu to show .gitignore file
- \n is added if not present, so new entries are grouped
- Ability to add the file or its extension

ezgif-4-0fdf2e68aacc

mlucool:
Do you think ignore extension is common? If so, to prevent user error, maybe we should disable ignore extension at on ipynb and maybe py files? Also maybe we should say `ignore *.[ext] files` or something like that?

kgryte:
@mlucool Ignoring all files having a given extension is supported by the GitHub desktop app, which is discussed in the linked issue and I believe the inspiration for the behavior here.

echarles:
Yes, I have quickly added that based on @ianhi input in https://github.com/jupyterlab/jupyterlab-git/issues/683#issuecomment-663864249. I can remove it if we don't want that.

I am trying to get a better message for the right-click menu with something like the following which does not work - any idea to get a more dynamic right click menu with `ignore *.[ext] files` ?

```javascript
      commands.addCommand(CommandIDs.gitIgnore, {
        label: 'Ignore',
        caption: `Ignore ${this.state.selectedFile.to}`,
```

If we keep the `Ignore extension`, I can exclude that menu for a predefined list of extension (I see a requirement coming to make that configurable, but if possible let's put that out-of-scope).

Waiting on more input/feedback here.

ianhi:
> I am trying to get a better message for the right-click menu with something like the following which does not work - any idea to get a more dynamic right click menu with ignore *.[ext] files ?

What about that is not working? Also https://jupyterlab.github.io/jupyterlab/modules/_coreutils_src_index_.pathext.html#extname may be a helpful thing to use.

Two tweaks I might make to phrasing in order to make it more obvious what is happening to users less experienced with Git:
1. `Ignore` -> `Ignore this file`
2. Add a parenthetical such as `(add to .gitignore)`

Overall I'm a big fan of where this going!

ianhi:
Although, rather than speculate as to what would be less confusing to users it may be better to ask one. I think @sjhaque14 is a good representative of users as she is a competent python user who uses jupyterlab for her work, but is fairly new to using Git. 

Sabina I'm curious what you would expect the result of the proposed menu commands to be? Also, would you expect the change to the `.gitignore` to be commited or not (see https://github.com/jupyterlab/jupyterlab-git/issues/683#issuecomment-649227764)?

echarles:
Thx @ianhi for the help and inputs.

In the meantime, I have done a few more experiments to bring the file/extension name in the context menu (when the user right-clicks a file). It turns out the current Lumino Widget constructs is such that new commands needs to be created with the correct label... which we do not want (commands are meant to remain controlled in number and unique in intention).

We still have the option to have static and user-relevant naming as suggested which would be good enough. I will tag this ready for review in a few days.

sjhaque14:
Thanks @ianhi for the mention. From a first glance, it is clear to me how to use the basic function – you go to the "Git" menu and select "Open .gitignore" to view the `.gitignore` file, or you can right click certain files and select "Ignore" to add its name to the `.gitignore` file. I was not clear on what the "Ignore Extensions" menu item is supposed to do. I initially thought that this meant users who had built jupyter extensions could add those extensions associated with a file to the `.gitignore`. But looking at it again, I understand that it adds the file extension (e.g. `.ipynb`) to the `.gitignore`, thus ignoring anything with that extension. That's certainly a very useful feature, and I don't think its inclusion in this proposed fix would detract from people being able to understand the basic function. Also, I do not expect that performing this menu item would automatically commit the changes – I would expect to have to commit the changes in a separate step.

fcollonval:
> In the meantime, I have done a few more experiments to bring the file/extension name in the context menu (when the user right-clicks a file). It turns out the current Lumino Widget constructs is such that new commands needs to be created with the correct label... which we do not want (commands are meant to remain controlled in number and unique in intention).

@echarles, I think the solution is to set the label and/or caption using a function

```js
caption: () => `Ignore ${this.state.selectedFile.to}`
```

echarles:
> @echarles, I think the solution is to set the label an>d/or caption using a function
> caption: () => `Ignore ${this.state.selectedFile.to}`

I have tried that and it also gives null pointer (without the function it fails directly without displaying the list, with the function it fails when you right click)

I have updated the label and the `Ignore this file extension` menu is only shown if the file has an effective extension different from ipynb and py.

kgryte:
@echarles Why the exceptions for `.py` and `.ipynb` files?

Given that (a) Jupyter can be used for languages other than Python and thus scenarios in which a user may want to ignore all Python files is possible and (b) I may, e.g., be working on a JLab extension but never want notebooks added to the repo (in fact, excluding temporary demo notebooks is something I do relatively frequently when developing extensions), I don't see the need to make those two file extensions exceptions.

echarles:
> @echarles Why the exceptions for .py and .ipynb files?

Have taken that input from @mlucool on https://github.com/jupyterlab/jupyterlab-git/pull/705#issuecomment-664407312

mlucool:
I was somewhat surprised to by ability to exclude an extension as it seems error prone. PyCharm for example does not do extension level ignores in their UI:
image

In retrospect, I don't think `.py` should be excluded. Given jupyter is made for notebooks, it would feel odd to make it easy to add `*.pynb` to `.gitignore`.

That's just my 2c, it would be good to see if we can get consensus on this issue.

kgryte:
Not sure it was a requirement, but was instead posed as a question.

fcollonval:
@echarles I found the problem for the label: `setState` is not instantaneous. Therefore, when the context menu is rendered, the state is not set. A step towards correcting it:

```diff
--- a/src/components/FileItem.tsx
+++ b/src/components/FileItem.tsx
@@ -33,7 +33,7 @@ export interface IFileItemProps {
   model: GitExtension;
   onDoubleClick: () => void;
   selected?: boolean;
-  selectFile?: (file: Git.IStatusFile | null) => void;
+  selectFile?: (file: Git.IStatusFile | null) => Promise<void>;
 }
 
 export interface IGitMarkBoxProps {
@@ -87,9 +87,9 @@ export class FileItem extends React.Component<IFileItemProps> {
         }
         onContextMenu={
           this.props.contextMenu &&
-          (event => {
+          (async event => {
             if (this.props.selectFile) {
-              this.props.selectFile(this.props.file);
+              await this.props.selectFile(this.props.file);
             }
             this.props.contextMenu(event);
           })
diff --git a/src/components/FileList.tsx b/src/components/FileList.tsx
index e8a8592..1c7dcbd 100644
--- a/src/components/FileList.tsx
+++ b/src/components/FileList.tsx
@@ -2,6 +2,7 @@ import { Dialog, showDialog, showErrorMessage } from '@jupyterlab/apputils';
 import { PathExt } from '@jupyterlab/coreutils';
 import { IRenderMimeRegistry } from '@jupyterlab/rendermime';
 import { ISettingRegistry } from '@jupyterlab/settingregistry';
+import { PromiseDelegate } from '@lumino/coreutils';
 import { Menu } from '@lumino/widgets';
 import * as React from 'react';
 import { GitExtension } from '../model';
@@ -168,7 +169,10 @@ export class FileList extends React.Component<IFileListProps, IFileListState> {
 
     if (!commands.hasCommand(CommandIDs.gitIgnoreExtension)) {
       commands.addCommand(CommandIDs.gitIgnoreExtension, {
-        label: 'Ignore this file extension (add to .gitignore)',
+        label: () =>
+          `Ignore ${PathExt.extname(
+            this.state.selectedFile.to
+          )} extension (add to .gitignore)`,
         caption: 'Ignore this file extension (add to .gitignore)',
         execute: async () => {
           await this.props.model.ignore(this.state.selectedFile.to, true);
@@ -357,8 +361,12 @@ export class FileList extends React.Component<IFileListProps, IFileListState> {
     await this.addFile(...this.markedFiles.map(file => file.to));
   };
 
-  updateSelectedFile = (file: Git.IStatusFile | null) => {
-    this.setState({ selectedFile: file });
+  updateSelectedFile = (file: Git.IStatusFile | null): Promise<void> => {
+    const stateSet = new PromiseDelegate<void>();
+    this.setState({ selectedFile: file }, () => {
+      stateSet.resolve();
+    });
+    return stateSet.promise;
   };
 
   get markedFiles() {
```

This works. But it brings React error about virtual event handling.

echarles:
@fcollonval Thx a lot for shimming in. I have opened https://github.com/jupyterlab/jupyterlab-git/issues/712 to track the fileitem state issue. This can be solved outside of this PR.

@mlucool My last commit allows to exclude `py` files (only `ipynb` extensions are excluded, which sounds logic for a notebook).

I think this PR takes everyone inputs and expectations, so it is ready for review.

jupyterlab_git/git.py
```diff
@@ -1112,6 +1112,48 @@ def remote_add(self, top_repo_path, url, name=DEFAULT_REMOTE_NAME):
...
+                open(gitignore, "w").close()
+            elif os.path.getsize(gitignore) != 0:
+                f = open(gitignore, "r")
+                if (f.read()[-1] != "\n"):
+                    f = open(gitignore, "a")
+                    f.write('\n')
+                f.close()
+            return {"code": 0}
+        except:
+            return {"code": -1}
```
fcollonval, line 1135:
```suggestion
        try:            
            gitignore = pathlib.Path(top_repo_path) / ".gitignore"
            if not gitignore.exists():
                gitignore.touch()
            elif gitignore.stat().st_size > 0:
                content = gitignore.read_text()
                if (content[-1] != "\n"):
                    with gitignore.open("a") as f:
                        f.write('\n')
        except BaseException as error:
            return {"code": -1, "message": str(error)}
        else:
            return {"code": 0}
```

I'm a big supporter of [pathlib.Path](https://docs.python.org/3/library/pathlib.html). It has nice feature like `touch`.

You should use context opening as much as possible. This ensures a proper file closure (the original code as some unbalanced open/close).

jupyterlab_git/git.py
```diff
@@ -1112,6 +1112,48 @@ def remote_add(self, top_repo_path, url, name=DEFAULT_REMOTE_NAME):
...
+            res = await self.ensure_gitignore(top_repo_path)
+            if res["code"] != 0:
+                return res
+            gitignore = os.path.join(top_repo_path, '.gitignore')
+            f = open(gitignore, "a")
+            f.write(file_path + "\n")
+            f.close()
+            return {"code": 0}
+        except:
+            return {"code": -1}
```
fcollonval, line 1155:
```suggestion
            gitignore = pathlib.Path(top_repo_path) / ".gitignore"
            with gitignore.open("a") as f:
                f.write(file_path + "\n")
        except BaseException as error:
            return {"code": -1, "message": str(error)}}
        else:
            return {"code": 0}
```

jupyterlab_git/handlers.py
```diff
@@ -534,6 +534,34 @@ async def post(self):
...
+        """
+        data = self.get_json_body()
+        top_repo_path = data["top_repo_path"]
+        file_path = data.get("file_path", None)
+        use_extension = data.get("use_extension", False)
+        if file_path:
+            if use_extension:
+                parts = os.path.splitext(file_path)
+                if len(parts) == 2:
+                    file_path = "**/*" + parts[1]
```
fcollonval, line 555: For file extension like `.tar.bz2`, `parts` will be `("name.tar", ".bz2")`. Is ignoring `.bz2` what is expected or is it ignoring `.tar.bz2`?

If this is the second option, I advice using again `pathlib.Path`. So you could request the full list of extensions from the `suffixes` property.

```py
pathlib.Path(file_path).suffixes
```

src/commandsAndMenu.ts
```diff
@@ -190,6 +192,21 @@ export function addCommands(
     }
   });
 
+  /** Add git open gitignore command */
+  commands.addCommand(CommandIDs.gitOpenGitignore, {
+    label: 'Open .gitignore',
+    caption: 'Open .gitignore',
+    isEnabled: () => model.pathRepository !== null,
+    execute: async () => {
+      model.ensureGitignore();
```
fcollonval, line 201:
```suggestion
      await model.ensureGitignore();
```

src/tokens.ts
```diff
@@ -278,6 +279,19 @@ export interface IGitExtension extends IDisposable {
    * @param path Path from which the top Git repository needs to be found
    */
   showTopLevel(path: string): Promise<Git.IShowTopLevelResult>;
+
+  /**
+   * Ensure a.gitignore file
```
fcollonval, line 284:
```suggestion
   * Ensure a .gitignore file exists
```

src/tokens.ts
```diff
@@ -278,6 +279,19 @@ export interface IGitExtension extends IDisposable {
...
+   *
+   */
+  ensureGitignore(): Promise<Response>;
+
+  /**
+   * Add an entry in .gitignore file
+   *
+   * @param filename The name of the entry to ignore
```
fcollonval, line 292:
```suggestion
   * @param filename The name of the entry to ignore
   * @param useExtension Ignore all file with the same extension than filename
```

src/components/FileList.tsx
```diff
@@ -197,7 +242,12 @@ export class FileList extends React.Component<IFileListProps, IFileListState> {
   /** Handle right-click on an untracked file */
   contextMenuUntracked = (event: React.MouseEvent) => {
     event.preventDefault();
-    this._contextMenuUntracked.open(event.clientX, event.clientY);
+    const extension = PathExt.extname(this.state.selectedFile.to);
+    if (extension.length > 0 && extension !== 'ipynb' && extension !== 'py') {
+      this._contextMenuUntracked.open(event.clientX, event.clientY);
+    } else {
+      this._contextMenuUntrackedMin.open(event.clientX, event.clientY);
+    }
```
fcollonval, line 250: I agree with @kgryte. We should not forbid ignoring `ipynb` or `py` files. In particular for the notebook extension, Jupytext users may want to ignore all ipynb files.

But I understand the necessity to protect newbies against that. So what about asking a confirmation for those use cases? With some warning message like:

```ts
Are you sure you want to ignore <b>all ${extension} files</b> within this git repository?
```

> The best place for such confirmation is in the command directly.

jupyterlab_git/git.py
echarles: Thx for the review. Last commit implements this.

src/commandsAndMenu.ts
echarles: Thx for the review. Last commit implements this.

src/tokens.ts
echarles: Thx for the review. Last commit implements this.

src/components/FileList.tsx
echarles: I have added a dialog box to ask confirmation when user wants to ignore a file extension.

src/tokens.ts
echarles: Thx for the review. Last commit implements this.

jupyterlab_git/git.py
echarles: Thx for the review. Last commit implements this.

jupyterlab_git/handlers.py
echarles: Thx for the review. Last commit implements this.

echarles:
@fcollonval Thx a lot for your reviews. My commit implements your suggestions.

src/model.ts
```diff
@@ -1169,6 +1169,65 @@ export class GitExtension implements IGitExtension {
   }
 
   /**
+   * Make request to ensure gitignore.
+   *
+   * @param filename Optional name of the files to add
```
kgryte, line 1174:
```suggestion
```

You can remove this parameter annotation, as this function does not appear to have the documented parameter.

src/model.ts
```diff
@@ -1169,6 +1169,65 @@ export class GitExtension implements IGitExtension {
...
+    });
+
+    this.refreshStatus();
+    return Promise.resolve(response);
+  }
+
+  /**
+   * Make request to ignore one file.
+   *
+   * @param filename Optional name of the files to add
```
kgryte, line 1202: The JSDoc comment here should be updated to reflect the current API.

src/tokens.ts
```diff
@@ -19,6 +19,7 @@ export interface IGitExtension extends IDisposable {
   /**
    * The current branch
    */
+
```
kgryte, line 22:
```suggestion
```

Not sure why this empty line was added.

src/tokens.ts
```diff
@@ -280,6 +281,19 @@ export interface IGitExtension extends IDisposable {
...
   /**
+   * Ensure a .gitignore file exists
+   */
+  ensureGitignore(): Promise<Response>;
+
+  /**
+   * Add an entry in .gitignore file
+   *
+   * @param filename The name of the entry to ignore
+   * @param useExtension Ignore all file with the same extension than filename
```
kgryte, line 292:
```suggestion
   * @param useExtension Ignore all files having the same extension as filename
```

echarles:
Thx @kgryte for your reviews. I have committed your suggestions.

kgryte:
@fcollonval You have any other feedback for this PR?

fcollonval:
> @fcollonval You have any other feedback for this PR?

This is good to go for me.

kgryte:
LGTM. Will merge now..,

echarles:
Thx all for the inputs, reviews and for merging this.

fcollonval:

PR #721, opened 2020-08-09
Guess default push target

fcollonval:
Fresh rebase of #412

### DIFF
--- a/jupyterlab_git/handlers.py
+++ b/jupyterlab_git/handlers.py
@@ -13,6 +13,11 @@ from packaging.version import parse
 from ._version import __version__
 from .git import DEFAULT_REMOTE_NAME
 
+
+# Git configuration options exposed through the REST API
+ALLOWED_OPTIONS = ['user.name', 'user.email']
+
+
 class GitHandler(APIHandler):
     """
     Top-level parent class.
@@ -265,13 +270,14 @@ class GitAddAllUntrackedHandler(GitHandler):
 class GitRemoteAddHandler(GitHandler):
     """Handler for 'git remote add <name> <url>'."""
 
-    def post(self):
+    @web.authenticated
+    async def post(self):
         """POST request handler to add a remote."""
         data = self.get_json_body()
         top_repo_path = data["top_repo_path"]
         name = data.get("name", DEFAULT_REMOTE_NAME)
         url = data["url"]
-        output = self.git.remote_add(top_repo_path, url, name)
+        output = await self.git.remote_add(top_repo_path, url, name)
         if(output["code"] == 0):
             self.set_status(201)
         else:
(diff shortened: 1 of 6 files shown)


fcollonval:
@ianhi This should fix #725

ianhi:
I don't think it fixed it for me:
pushing-4

jupyterlab_git/handlers.py
```diff
@@ -436,32 +445,76 @@ async def post(self):
...
+                    current_local_branch, 
+                    current_path, 
+                    data.get("auth", None), 
+                    set_upstream=True,
+                )
+            else:
                 response = {
                     "code": 128,
                     "message": "fatal: The current branch {} has no upstream branch.".format(
                         current_local_branch
```
ianhi, line 511: Maybe we should modify this error message to explain to the user why the guessing of the push target failed. Something along the lines of this
```suggestion
                response = {
                    "code": 128,
                    "message": "The current branch {} has no upstream branch. You have {} remotes so a remote cannot be chosen automatically without having set a pushdefault.".format(
                        current_local_branch,
                        len(remotes)
```

fcollonval:
@ianhi, do you have multiple remotes? Could you post the push request response for the defunct case?

jupyterlab_git/handlers.py
fcollonval: That error message will be used also for a repository without remote.

So I will not change the error message. But prompt the user to choose from the remote list if it exists. Hence your use case will be covered. What do you think?

ianhi:
I only have one remote on that repo

```bash
(jlab-git) ian@pop-os:~/Documents/jupyter/git-expt$ git remote -vv
origin  git@github.com:ianhi/git-expt.git (fetch)
origin  git@github.com:ianhi/git-expt.git (push)
```
I think this is due to a weird interaction between #630 and existing modal for git push/pull here:
https://github.com/jupyterlab/jupyterlab-git/blob/abfdcf20df226bfeb0927adac7abc80618afaf78/src/commandsAndMenu.ts#L303-L333

Adding some `console.log`s shows that when I click off the dialog the `await` finishes but `result.button.accept` remains false. Then that same interaction can continue indefinitely due to the while loop

ianhi:
Ooh actually this predates #630 checking out an arbitrary commit from before that (2f9ad78074bd8219200587020236ac77daa761be) I also see this issue. 

I think the dialog is being closed due to a jupyterlab problem. See the unmerged https://github.com/jupyterlab/jupyterlab/pull/8438 which stops dialogs from being closed by clicking outside them if they do not have a cancel button.

ianhi:
I think the solution to #725 is to eliminate the old modal and adapt push/pull to use the new modal + toast notifications. But that could be done separately to this PR?

jupyterlab_git/handlers.py
ianhi: > But prompt the user to choose from the remote list if it exist

That is the best solution by far. My suggestion was to make it more clear if there were to be no further frontend changes.

fcollonval:
> I think the solution to #725 is to eliminate the old modal and adapt push/pull to use the new modal + toast notifications. But that could be done separately to this PR?

Thanks @ianhi for the detailed analysis. At the beginning I hoped to use the new model. But unfortunately it is not based on some global function that emits it. Moreover as the action is executed from a command and not a react component, this makes the switch quite complicated. So I try staying on the not robust actual code.

The ideal code word look like:

```ts
waitForCall("Pushing...", model.push, args).catch(error => 
if error due to auth then
  Request credentials
  Re try
else if multiple remotes then
  User needs to pick one
  Re try
)
```
And so this should be independent of the waiting ui component.

jupyterlab_git/handlers.py
ianhi: Will having the option to prompt the user to choose from a remote list run into the same issues as the Git Credentials w.r.t the new modal?

PR #722, opened 2020-08-09
Bump to 0.21.0a0

fcollonval:


ianhi:
@fcollonval it looks like the the version.ts was not updated by the github action

https://github.com/jupyterlab/jupyterlab-git/blob/06326ae5f76e09546cab232838c1756b6d1b8b6a/src/version.ts#L2

is that important to keep updated?

Issue #728, opened 2020-08-13
Python Styling Proposal


mlucool:
Hi,

I'd like to propose we move all python code to be styled by [black](https://github.com/psf/black/). It will increase readability across the code base IMO. The downside is black is not at 1.0 yet, but we can run black just on changed files and set minimum versions in setup.py (see [projects using black](https://github.com/psf/black#used-by)).

What do others think?

PR #763, opened 2020-09-05
Format code using black

ianhi:
closes: https://github.com/jupyterlab/jupyterlab-git/issues/728

1. Formatted existing code `black .`
2. added `black . --check` to github actions
3. Saved formatting commit SHA in a file following https://github.com/psf/black/tree/20.8b1#migrating-your-code-style-without-ruining-git-blame
4. added pre-commit to setup.py and updated install instructions (https://github.com/psf/black/tree/20.8b1#version-control-integration)

@mlucool @fcollonval


ianhi:
Tests fail because black is not available on python 3.5. This can be worked around with some if statements but it seems to get a little bit ugly as github actions don't support line level if statements. Given that python 3.5 is at the end of life in 9 days (september 13) I propose dropping it from the tests. Thoughts?

fcollonval (approved):
Thanks @ianhi 

I agree with you. We can drop testing against 3.5 and I went ahead by adding 3.9rc.

I reorganize a bit the CI script so only jupyterlab needs to be explicitly installed before the test step (to get `jlpm`).

mlucool:
This is great! Does this commit add in a way to verify that future commits use black? i.e. have `black --check` as a test or similar?

fcollonval:

PR #766, opened 2020-09-06
Fix text diff for 1st commit and deleted files

fcollonval:
This fixes the display of text diff in history for two cases:
- First commit (no previous commit)
- Commit at which the file is deleted

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -1085,6 +1085,7 @@ class Git:
                 "fatal: Path '{}' does not exist (neither on disk nor in the index)".format(
                     filename
                 ),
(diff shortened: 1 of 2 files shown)


ianhi:
Nice!

One note: I think this is still missing one case with deleted files. It works as I'd expected from the history panel, but if you try to get the diff of the file from the commit panel you will get an error
diff-of-deleted

fcollonval:
Good catch @ianhi. I will extend this PR to cover that case.

Thank for reviewing

fcollonval:
I added the versioned file being deleted case

ianhi (approved):
This is great thanks!

fcollonval:

PR #767, opened 2020-09-08
Virtualized files list

### LINKED ISSUE #667: a high number of untracked files make the extension and jupyter lab unresponsive
I have a repo with a lot (10 000+) untracked files (mostly intermediate figures and tables I do not need to track, not excluded through a .gitignore) which get listed in the corresponding section of the UI but make the extension and the whole jupyter lab unresponsive.

Is there any remedy?

### LINKED ISSUE #703: Git status API recursively lists all untracked files causing browsers to hang

fcollonval:
Use a virtual component to display files list.

Fixes #667 
Closes #703 
Closes #704

TODO
- [x] Set nice height strategy
- [x] Set windowing on history file list
- [x] Set windowing on branch list

### DIFF
--- a/jupyterlab_git/handlers.py
+++ b/jupyterlab_git/handlers.py
@@ -175,6 +175,9 @@ class GitDetailedLogHandler(GitHandler):
         selected_hash = data["selected_hash"]
         current_path = data["current_path"]
         result = await self.git.detailed_log(selected_hash, current_path)
+
+        if result["code"] != 0:
+            self.set_status(500)
         self.finish(json.dumps(result))
 
 
--- a/package.json
+++ b/package.json
@@ -73,6 +73,8 @@
     "react": "~16.9.0",
     "react-dom": "~16.9.0",
     "react-textarea-autosize": "^7.1.2",
+    "react-virtualized-auto-sizer": "^1.0.2",
+    "react-window": "^1.8.5",
     "typestyle": "^2.0.1"
   },
   "devDependencies": {
@@ -86,6 +88,8 @@
     "@types/react": "~16.8.13",
     "@types/react-dom": "~16.0.5",
     "@types/react-textarea-autosize": "^4.3.5",
+    "@types/react-virtualized-auto-sizer": "^1.0.0",
+    "@types/react-window": "^1.8.2",
     "@typescript-eslint/eslint-plugin": "^2.25.0",
     "@typescript-eslint/parser": "^2.25.0",
     "all-contributors-cli": "^6.14.0",
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -3,6 +3,7 @@ import List from '@material-ui/core/List';
 import ListItem from '@material-ui/core/ListItem';
 import ClearIcon from '@material-ui/icons/Clear';
 import * as React from 'react';
+import { FixedSizeList, ListChildComponentProps } from 'react-window';
 import { classes } from 'typestyle';
 import {
   activeListItemClass,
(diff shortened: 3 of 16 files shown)


fcollonval:
This is an early prototype - styling issues need to be addressed (i.e. height handling). But at least it shows that windowing the files list will fix the performance issue.

fcollonval:
@ianhi I'll go ahead and merge this one as I need it for other improvements. But if you could check you don't see any glitches on the master after that, I'll appreciate.

ianhi:
:+1:  will do

ianhi:
This fixes the total freezing of the UI and I don't think it introduced any bugs -  so hooray!

We should maybe still consider compressing the result of status? When I create 10000 files the status request starts to get large and while it's still pretty quick at `65` ms if it gets much longer I suspect it will start to feel laggy:
image

At least one person reported that the payload was very large for them: https://github.com/jupyterlab/jupyterlab-git/issues/667#issuecomment-642940456

ZhangHan:
@fcollonval Great this gets fixed in 0.2.X! Is there any concern back porting this fix to 0.10.x and include in a new release?

fcollonval:
> @fcollonval Great this gets fixed in 0.2.X! Is there any concern back porting this fix to 0.10.x and include in a new release? 

We don't have bandwidth to support older versions. But if you are willing to help. You could trigger a backport with the following comment:

> @meeseeksdev backport to 0.10.x

If you succeed to make it works. I can make a release.

Small question why not on 0.11.x as that is the latest version for JupyterLab 1?

PR #783, opened 2020-09-26
Use only SVG icons

fcollonval:
Now that the elements lists are windowed to limit the number of rendered elements, it is safe to use svg icons everywhere.

In particular, this will benefit to get new file icon from JupyterLab core directly in the extension.


src/style/FileItemStyle.ts
```diff
@@ -22,12 +22,12 @@ export const fileStyle = style(
 );
 
 export const selectedFileStyle = style({
-  color: 'white',
+  color: 'var(--jp-ui-inverse-font-color1)',
```
telamonian, line 25: The text and icons for the "selected" styles should always stay as white. This is because the "selected" background color is always `--jp-brand-color1`, which is a bright blue color that typically does not change from theme to theme

telamonian (changes requested):
There are some cases where the font/icon color should always be white, and not tied to any theme var

src/style/FileItemStyle.ts
fcollonval: This sounds in contradiction with the principle of theming and with the comment in the core package:

```css
  /*
   * Use these against the brand/accent/warn/error colors.
   * These will typically go from light to darker, in both a dark and light theme.
   */

  --jp-ui-inverse-font-color0: rgba(255, 255, 255, 1);
```

https://github.com/jupyterlab/jupyterlab/blob/8eba49bbb3981fefd16c55f70fb03e6bc03855ad/packages/theme-light-extension/style/variables.css#L127-L128

But I see in the core that the two definitions exists:
- All variables: https://github.com/jupyterlab/jupyterlab/blob/141d0142de90bf9756e10c7a7ddd9fecfd76c5ab/packages/extensionmanager/style/base.css#L97
- Force white: https://github.com/jupyterlab/jupyterlab/blob/cb3595e5e8c02ba2c28f748de6ab4846a35ccdad/packages/apputils/style/commandpalette.css#L92

telamonian: I agree that it's contradictory/messy. For example, check out [the definition of `jp-ui-inverse-font-color0` from the dark theme](https://github.com/jupyterlab/jupyterlab/blob/02fb48085a774dacb8af5a9c02fd132ac7721d7f/packages/theme-dark-extension/style/variables.css#L128-L133):

```
  /*
   * Use these against the brand/accent/warn/error colors.
   * These will typically go from light to darker, in both a dark and light theme.
   */


  --jp-ui-inverse-font-color0: rgba(0, 0, 0, 1);
```

In contradiction of the comment, `jp-ui-inverse-font-color0` is pure black. Which is not a very readable font color for text on a bright blue background. So for now, the pragmatic solution is to just set the relevant font colors to white.

telamonian: One solution would be to introduce a `jp-inverse-brand-colorX` variable series to the `variables.css` file in the themes. But there are already too many css vars in each theme, many of which are redundant, unused, or unlikely to change.

So the "right" solution will probably be to do an audit of the css vars we use for theming, then work out a reduced set of vars that resolve this and other contradictions in the theming system. But this will be an API breaking change, so will have to wait until at least jlab 4.0.

So for now just do the pragmatic thing and set "selected" font/icon colors to white.

ianhi: I agree that that pragmatic thing is to return to white. I just tried out a few different themes and while it looks ok/good in some:
default dark:
image

theme-darcula:
image

In others it is unreadable:
jupyterlab-solarized-dark
image

fcollonval: Thanks both of you for all the information and screenshots. So let's get back to white.

For the buttons I wonder if we should not move to `--md-color-xxx` to increase homogeneity with core JLab: 

https://github.com/jupyterlab/jupyterlab/blob/master/packages/application/style/buttons.css

PR #790, opened 2020-10-02
sort main branch in NewBranchDialog

ianhi:
As of yesterday the new default branch on Github will be called `main` ([ref](https://github.blog/changelog/2020-10-01-the-default-branch-for-newly-created-repositories-is-now-main/)). This means that a sizeable number of new repos will have `main` as the default branch when people copy-paste the github new repo instructions:
image

To account for that this PR adds the `main` branch to special cases when sorting branches in the newBranchDialog. The sorting order will now be:
1. `current branch`
2. `master` (if exists)
3. `main` (if exists)
4. other

### DIFF
--- a/src/components/NewBranchDialog.tsx
+++ b/src/components/NewBranchDialog.tsx
@@ -283,6 +283,11 @@ export class NewBranchDialog extends React.Component<
       } else if (b.name === 'master') {
         return 1;
       }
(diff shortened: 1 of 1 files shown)


fcollonval (approved):
Thanks for following up the upcoming changes in GitHub.

PR #794, opened 2020-10-06
Style issues with SVG icons

### LINKED ISSUE #793: Visual regressions of the icons in the files list
## Icons resize when expanding the panel

icons_resize

Recorded using binder in Firefox 81 on Ubuntu. Same happens in Chrome.
Binder is pinned at JupyterLab 2.2.0

## List is scrollable and needlessly hides elements 

scrolling

Recorded using binder in Firefox 81 on Ubuntu. Same happens in Chrome but the scrollbars are even more annoying.
Binder is pinned at JupyterLab 2.2.0

## Icons are huge and the list is completely unusable when zoomed-in Chrome.

zoom

Chrome 85, Ubuntu; zooming back to the 100% level does not help. Does not happen in Firefox.

fcollonval:
Fixes #793 

Got the scrolling weird effect to disappear by switching to pure div elements rather than ul/li for changed files list.

### DIFF
--- a/src/components/FileItem.tsx
+++ b/src/components/FileItem.tsx
@@ -160,7 +160,7 @@ export class FileItem extends React.PureComponent<IFileItemProps> {
     const status = this._getFileChangedLabel(status_code as any);
 
     return (
-      <li
+      <div
         className={this._getFileClass()}
         onClick={
           this.props.selectFile &&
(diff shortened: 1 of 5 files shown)


fcollonval:
/binder

@krassowski Could you please test this PR to check it fixes the style problems you found?

krassowski:
It behaves much better. All the icon sizing issues are gone, and scrolling seems better, although it is still not perfect - if the list is long enough, a second scrollbar appear:

scroll_2

fcollonval:
Thanks @krassowski 

Regarding the scroll bar, double scroll bars is per design needed. But the second should be on the all panel if needed.

I created a new issue for it. So I can merge this and release a patch version.

Issue #641, opened 2020-05-23
Allow to ignore repositories


memeplex:
I've a very large repository that hangs the entire Jupyter UI every time I open its directory. Some other tools that work with git allow to ignore some directories/repositories.

fcollonval:
@memeplex Thanks for reporting. Could you please use the issue template next time and report the context:
```
## Context

<!--Complete the following for context, and add any other relevant context-->

- Python package version:
<!-- Results of `conda list jupyterlab-git` or `pip show jupyterlab-git` -->
- Extension version:
<!-- Results of `jupyter labextension list` -->

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>
<pre>
Paste the output from your browser Javascript console here.
</pre>
</details>
```

memeplex:
Since it's more of a simple feature request I thought that kind of context wasn't relevant, but here you go:

- Python package version: 0.20.0
- Extension version: 0.20.0

<details><summary>Command Line Output</summary>
<pre>
[D 09:47:16.044 LabApp] Searching ['/home/carlos', '/home/carlos/.jupyter', '/usr/etc/jupyter', '/usr/local/etc/jupyter', '/etc/jupyter'] for config files
[D 09:47:16.044 LabApp] Looking for jupyter_config in /etc/jupyter
[D 09:47:16.044 LabApp] Looking for jupyter_config in /usr/local/etc/jupyter
[D 09:47:16.045 LabApp] Looking for jupyter_config in /usr/etc/jupyter
[D 09:47:16.045 LabApp] Looking for jupyter_config in /home/carlos/.jupyter
[D 09:47:16.045 LabApp] Looking for jupyter_config in /home/carlos
[D 09:47:16.045 LabApp] Looking for jupyter_notebook_config in /etc/jupyter
[D 09:47:16.045 LabApp] Looking for jupyter_notebook_config in /usr/local/etc/jupyter
[D 09:47:16.045 LabApp] Looking for jupyter_notebook_config in /usr/etc/jupyter
[D 09:47:16.045 LabApp] Looking for jupyter_notebook_config in /home/carlos/.jupyter
[D 09:47:16.046 LabApp] Loaded config file: /home/carlos/.jupyter/jupyter_notebook_config.json
[...]

PR #812, opened 2020-10-31
Reduce polling

fcollonval:
This reduce refresh polling.

Refresh is blocked if:
- Not in a git repository
- The Git tab is hidden (this can be turned off with the setting `refreshIfHidden`)


ianhi:
Is it possible to immediately start a refresh when the git tab is brought back into view? Otherwise it may end up being that you switch and then need to wait three seconds for a refresh

ianhi:
This could also be used to implement the ability to ignore repositories (https://github.com/jupyterlab/jupyterlab-git/issues/641) There could a setting that contains a list of filepaths to ignore.

fcollonval:
> 
> 
> Is it possible to immediately start a refresh when the git tab is brought back into view? Otherwise it may end up being that you switch and then need to wait three seconds for a refresh

It is a good suggestion. I'll try to find a way.

> This could also be used to implement the ability to ignore repositories (#641) There could a setting that contains a list of filepaths to ignore.

I tend to set `won't fix` on #641 as the major problem was jupyterlab hanging because of a big repo. So with this PR and the virtual files list, I think if the UI is still hanging it should be investigated instead of hidden.

fcollonval:
> Is it possible to immediately start a refresh when the git tab is brought back into view?

Done 😄

ianhi:
Looks mostly good except I seem to be getting double status requests:

image

ianhi:
Oh hmm. I see the same double status POST on master as well. Is that intentional or is that a bug?

schema/plugin.json
```diff
@@ -41,6 +41,12 @@
       "description": "Number of (most recent) commits shown in the history log",
       "default": 25
     },
+    "refreshIfHidden": {
+      "type": "boolean",
+      "title": "Refresh if Git tab is hidden",
+      "description": "Whether to refresh even if the Git tab is hidden",
```
ianhi, line 47: `refresh` doesn't have a specific meaning in terms of Git, so it may be less clear what this means someone who hsan't developed this extension. This maybe clarifies it bit?

```suggestion
      "title": "Refresh if the Git tab is hidden",
      "description": "Whether to check Git status when the Git tab is not visible. Choose `false` for higher performance.",
```

fcollonval: Thanks for the suggestion. It sounds clearer.

fcollonval:
> Oh hmm. I see the same double status POST on master as well. Is that intentional or is that a bug?

Good catch, I'll take a look

fcollonval:
I replace the test by this one

```
      const headChanged =
        this._currentBranch.name !== data.current_branch.name ||
        this._currentBranch.top_commit !== data.current_branch.top_commit;
```

I think testing for the branch name and top commit get us cover.

PR #813, opened 2020-10-31
Add indicators on push/pull icons if actions is required

### LINKED ISSUE #582: UX: Git Push/Pull
## Description

When push does not error, the UI always says:
image

While this is not wrong, it's misleading when nothing was pushed. The git CLI says:
```
$ git push
Everything up-to-date
```

We should do something similar in the UI so people know that everything was up to date and nothing happened. The same is also true for the pull messages.

### LINKED ISSUE #780: Add indicators on pull/push branch status
## Description

Currently there is no visual feedback for the branch status if
- branch is ahead of upstream
- branch is behind upstream

It would be nice to add an indicator on the push/pull icons when the current branch is ahead or behind the upstream.

fcollonval:
Fixes #780

Side effect: update jest

TODO:
- [x] Execute fetch to get the latest ahead/behind status
- [x] Show push badge for new branch without upstream
- [x] Fix #582


jupyterlab_git/git.py
```diff
@@ -298,11 +302,25 @@ async def status(self, current_path):
...
-            cwd=os.path.join(self.root_dir, current_path),
-        )
+        cwd = os.path.join(self.root_dir, current_path)
+        # Start by fetching to get accurate ahead/behind status
+        cmd = [
+            "git",
+            "fetch",
+            "--all",
+            "--prune",
+        ]  # Run prune by default to help beginners
```
ianhi, line 312: Will this slow things down? On my laptop with a reasonably fast internet connection (fast for america :cry: ) I get:
`time git fetch --prune --all`
gives:
```
Fetching origin

real	0m0.607s
user	0m0.028s
sys	0m0.034s
```

fcollonval: It will definitely depend on the internet connection and the number of changes to be downloaded.

From experience in VS Code, I never got the impression things got stuck. So I propose to move forward with it. And if users starts to complain, we could move it in its own poll running parallel to the status refresh with a dedicated interval. What do you think?

fcollonval: At the end I followed your advice and set up a separate poll to fetch remote. This has the side-effect of stopping that poll if the repository does not have remote defined.

Issue #687, opened 2020-07-07
Add Delete option to untracked files in git menu


ianhi:
## Description
I'd like to be able to delete untracked files from the Git panel. I often have the Git panel open instead of the filebrowser and just want to quickly delete a file I may have accidentally created but don't want to add to a gitignore. This option would avoid this pattern:
1. See file I don't want in the git menu - decide I want to get rid of it
2. Return to file browser
3. right click and delete
4. return to git panel.

Both Github Desktop and VSCode handle `Discard Changes` to an untracked file by deleting the file. VSCode also adds this error message:
![image](https://user-images.githubusercontent.com/10111092/86692723-0838c200-bfd8-11ea-88da-4b295a5d58cd.png)

ianhi:
This should be accomplished using the docmanager `deleteFile` method https://jupyterlab.github.io/jupyterlab/interfaces/_docmanager_src_index_.idocumentmanager.html#deletefile

https://github.com/jupyterlab/jupyterlab-git/pull/676 will make the docmanager accessible so maybe after/if that is merged this should be addressed.

PR #815, opened 2020-10-31
add context menu option to delete untracked files

ianhi:
Closes: https://github.com/jupyterlab/jupyterlab-git/issues/687

Made a new command for this so we can present a warning dialog. Using the `docmanager:delete-file` command directly doesn't allow for a warning dialog.

### DIFF
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -78,6 +78,7 @@ export namespace CommandIDs {
   // Context menu commands
   export const gitFileDiff = 'git:context-diff';
   export const gitFileDiscard = 'git:context-discard';
+  export const gitFileDelete = 'git:context-delete';
   export const gitFileOpen = 'git:context-open';
   export const gitFileUnstage = 'git:context-unstage';
   export const gitFileStage = 'git:context-stage';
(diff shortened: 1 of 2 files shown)


ianhi:
delete-untracked

src/commandsAndMenu.tsx
```diff
@@ -478,6 +479,36 @@ export function addCommands(
...
+          <span>
+            Are you sure you want to permanently delete
+            <b>{file.to}</b>? This action cannot be undone.
+          </span>
+        ),
+        buttons: [Dialog.cancelButton(), Dialog.warnButton({ label: 'Delete' })]
+      });
+      if (result.button.accept) {
+        try {
+          app.commands.execute('docmanager:delete-file', {
```
fcollonval, line 500:
```suggestion
          await app.commands.execute('docmanager:delete-file', {
```

fcollonval (changes requested):
Almost good to go.

You should wait on the command execution otherwise error may not be catch.

src/commandsAndMenu.tsx
ianhi: Done - thanks!

fcollonval (approved):
Thanks LGTM!

PR #816, opened 2020-10-31
Delete local branch 

### LINKED ISSUE #308: git branch - Delete local branch (Branch Management)
Currently it is not possible to delete a local branch.  It would be useful to be able to do this from the GUI tab.

Proposal - In the git-tab, when a user clicks on a branch from the drop down list of available branches, a prompt box could be presented asking the user if they would like to checkout / delete / rename the branch.

fcollonval:
Fixes #308

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -505,6 +505,8 @@ class Git:
         )
         if code != 0:
             return {"code": code, "command": " ".join(cmd), "message": error}
+        else:
+            return {"code": code}
 
     async def branch_heads(self, current_path):
         """
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -497,6 +497,15 @@ class Git:
             "current_branch": heads["current_branch"],
         }
 
+    async def branch_delete(self, current_path, branch):
+        """Execute 'git branch -D <branchname>'"""
+        cmd = ["git", "branch", "-D", branch]
+        code, _, error = await execute(
+            cmd, cwd=os.path.join(self.root_dir, current_path)
+        )
+        if code != 0:
+            return {"code": code, "command": " ".join(cmd), "message": error}
+
     async def branch_heads(self, current_path):
         """
         Execute 'git for-each-ref' command on refs/heads & return the result.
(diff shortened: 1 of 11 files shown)


jupyterlab_git/handlers.py
```diff
@@ -234,6 +234,31 @@ async def post(self):
...
+        """
+        POST request handler, delete branch in current repository.
+
+        Body: {
+            "current_path": Git repository path relatively to the server root,
+            "branch": Branch name to be deleted
+        }
+        """
+        data = self.get_json_body()
+        result = await self.git.delete_branch(data["current_path"], data["branch"])
```
ianhi, line 253: This line uses `delete_branch` but in `git.py` it is `branch_delete`

ianhi (changes requested):
currently broken due to mismatched function names

jupyterlab_git/git.py
```diff
@@ -497,6 +497,15 @@ async def branch(self, current_path):
...
 
+    async def branch_delete(self, current_path, branch):
+        """Execute 'git branch -D <branchname>'"""
+        cmd = ["git", "branch", "-D", branch]
+        code, _, error = await execute(
+            cmd, cwd=os.path.join(self.root_dir, current_path)
+        )
+        if code != 0:
+            return {"code": code, "command": " ".join(cmd), "message": error}
+
```
ianhi, line 508: I think you also need to return sometime in the case of success. Otherwise the `if result['code']` in the handler will fail
```suggestion
        if code != 0:
            return {"code": code, "command": " ".join(cmd), "message": error}
        else:
            return {"code": 0}

```

ianhi:
Does it make sense to have the tooltip indicate that this only a local deletion? Something like: `Delete this branch locally`

fcollonval:
Thanks @ianhi and sorry for the bugs. I focused on the frontend testing and forgot I haven't tested the backend...

fcollonval:
@ianhi Thanks for the review and suggestions (I push all of them). I also trigger a refresh right after deletion to update the UI as quickly as possible.

ianhi (approved):
This is great - thanks!

PR #818, opened 2020-11-03
Update to Jupyterlab 3.0

### LINKED ISSUE #779: Update for jupyterlab 3
**edit - Install jupyterlab-git beta**: Jupyterlab-git has a beta release which you can install with:
```
pip install --upgrade --pre jupyterlab-git
```

## Description
Jupyterlab version 3 is nearly upon us (rc this week I think?). 
The update to jlab3 may be as simple as running: `python -m jupyterlab.upgrade_extension .` but what should be done w.r.t branches on this repo? Should master track jlab 3 and then create a `0.2x.y` branch for bugfixes for jlab2?

ajbozarth:
Update to support Jupyterlab 3.0 and addresses any issues accosted with that update.

Currently in draft until final release when this will be updated.

This PR is paired with https://github.com/jupyter/nbdime/pull/551 which is require to test this, including a temporary 1 line hard-coded local import of nbdime 

Fixes #779


ajbozarth:
note: CI will fail until nbdime PR is merged and released

jupyterlab_git/_version.py
```diff
@@ -1,7 +1,7 @@
 # Copyright (c) Project Jupyter.
 # Distributed under the terms of the Modified BSD License.
 
-version_info = (0, 22, 3)
+version_info = (0, 23, 0)
```
fcollonval, line 4:
```suggestion
version_info = (0, 30, 0)
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@jupyterlab/git",
-  "version": "0.22.3",
+  "version": "0.23.0",
```
fcollonval, line 3:
```suggestion
  "version": "0.30.0",
```

package.json
```diff
@@ -72,9 +73,9 @@
     "@material-ui/icons": "^4.5.1",
     "@material-ui/lab": "^4.0.0-alpha.54",
     "diff-match-patch": "^1.0.4",
-    "nbdime": "^6.0.0",
-    "react": "~16.9.0",
-    "react-dom": "~16.9.0",
+    "nbdime": "file:../nbdime/packages/nbdime",
```
fcollonval, line 76: Just to highlight this needs to be updated once nbdime has been updated.

fcollonval (commented):
Thanks for starting this @ajbozarth 

I push some suggestions to correct the version number to match the current policy.

ianhi:
I think we should also turn this into a federated extension to eliminate the build step on installation. Thoughts on this?

jupyterlab_git/_version.py
ajbozarth: why jump from 22 to 30?

ajbozarth:
> I think we should also turn this into a federated extension 

I agree, but I believe that initially just updating to 3.0 as a normal extension to be release same day as jupyterlab 3.0 then following up with an update to a federated extension later is the safer option and will cause less delays to users who want to switch to 3.0

jupyterlab_git/_version.py
fcollonval: When the extension was ported to JupyterLab 2, it was decided to jump from 0.10 to 0.20 to left version available for version with backported fix and features. So it will be good to do the same with JLab 3 to left version 0.2x available for backports for JupyterLab 2.

setup.py
```diff
@@ -80,13 +80,13 @@ def runPackLabextension():
         "Programming Language :: Python :: 3.9",
         "Framework :: Jupyter",
     ],
-    install_requires=["notebook", "nbdime ~=2.0", "packaging", "pexpect"],
+    install_requires=["notebook", "nbdime ~=3.0.0dev0", "packaging", "pexpect"],
```
ajbozarth, line 83: also leaving myself a note this will need to be updated when nbdime is updated as well

ajbozarth:
recently update the nbdime PR and requested a rc, once an nbdime rc is released I'll update this PR

ajbozarth:
I've pushed some updates so we can get some review started on this before the nbdime pr is published. Currently the tests are broken and I'm open to others pushing to this PR to address them if you have ideas.

As for checking out and running this PR, you will need to also checkout and build the linked nbdime PR and update the package.json here to the relative location of that project on your machine for this to work.

ajbozarth:
just pushed an update to use the nbdime beta, this PR should now be usable without needing to install a local dev instance of nbdime

fcollonval:
Thanks @ajbozarth 

We are almost there. There is only one fix to figure out why looking into the React tree is not working anymore. Happy to get some help on that one.

setup.py
```diff
@@ -66,7 +66,7 @@
     long_description_content_type="text/markdown",
     cmdclass=cmdclass,
     packages=setuptools.find_packages(),
-    install_requires=["jupyter_server", "nbdime ~=3.0.0b1", "packaging", "pexpect"],
+    install_requires=["jupyter_server", "nbdime~=3.0.0b1", "jupyter_packaging~=0.7.9", "packaging", "pexpect"],
```
fcollonval, line 69:
```suggestion
    install_requires=["jupyter_server", "nbdime~=3.0.0b1", "packaging", "pexpect"],
```

`jupyter_packaging` is not a install dependency but a setup one (only need by dev'). We should add it to the extra requires - and probably rename it as `dev` rather than `test` as there are already other packages like pre-commit or black in there.

meeseeksmachine:

https://discourse.jupyter.org/t/git-extension-for-jupyterlab-3/7569/2

ajbozarth:
@fcollonval @lresende thank you both for helping on this, is this good to merge now? If not what's left blocking it and releasing an rc or beta?

fcollonval:
> @fcollonval @lresende thank you both for helping on this, is this good to merge now? If not what's left blocking it and releasing an rc or beta?

I'm ok to release a beta. For my information, on the `nbdime` front, what is missing to get a release?

fcollonval (approved):

Issue #823, opened 2020-11-12
Fails to discard unstaged changes - cannot read property 'filename' of undefined


lresende:
## Description

![git-fails-to-discard-changes](https://user-images.githubusercontent.com/382917/98893899-6549f600-2458-11eb-8bb7-cd6cab843ccb.gif)

This is using 0.23.0 but I believe I saw similar behavior on 0.22.x as well.

ianhi:
on master I get this error: 
![image](https://user-images.githubusercontent.com/10111092/98898765-52491d00-247c-11eb-954b-c50fa1ca6d4b.png)

I think two things are happening:
1. `options` is being used outside of the check for if it is undefined here:
https://github.com/jupyterlab/jupyterlab-git/blob/582ef5af67b8363b316dc29b92d935ed208ffaad/src/model.ts#L399
2. discard all changes doesn't use the options to specify the files that are being discarded. instead it just run `git checkout`. This will revert the changes but if there are open editors then they won't be updated to reflect the state on disk. (I think this is also a bug)

ianhi:
Changing the revertFile lines to:
```typescript
        } else if (body.filename) {
          this._revertFile(body.filename);
        }
```

fixes the error, but there will still be the issue that open editors aren't updated when doing revert all changes

PR #824, opened 2020-11-12
check for changed files when discarding all changes

ianhi:
Previously when discarding all changed files it was never checked what the names of the changed files were. This meant that open editors were not updated when the files changes. As well this led to a bug where options was referred to even when undefined.

Now we check for the changed files before doing the checkout to all us to pick up the changes in the case of discard all, and will fall back to querying the list of currently modified files if the `body` has no filename or branch name.

Closes: https://github.com/jupyterlab/jupyterlab-git/issues/823

attn: @lresende


fcollonval (approved):
LGTM 
Thanks for the quick fix @ianhi

Issue #828, opened 2020-11-19
Creating a new branch


simonwhite71:
If I try and make a new branch called "test" with the UI it says 
`ambiguous argument 'test': unknown revision or path not in the working tree. Use '--' to separate paths from revisions, like this: 'git <command> [<revision>...] -- [<file>...]'`
Which is the same error I see if on the command line I run 
`git diff test`
rather than
`git checkout -b test`
Am I doing something wrong?

Also if in the UI I try to make a branch with name `-b test` I get the error message
`usage: git diff [<options>] [<commit> [<commit>]] [--] [<path>...]`

ianhi:
Hey @simonwhite71 thanks for the report. You aren't doing anything wrong, this looks like a bug that was introduced by https://github.com/jupyterlab/jupyterlab-git/pull/824

In that PR we started checking for what files will be changed by the git action before we switch branches. But that breaks down if we are opening a new branch. The relevant lines are these:
https://github.com/jupyterlab/jupyterlab-git/blob/1a57cab0b5babb7186e8f12795d284f47146e957/src/model.ts#L385-L405

ianhi:
Also, great job tracking down that it was a git diff error. The reason `git diff` is called when checking out a branches is so that we can keep track of what files are being touched by the git action  to make sure that if they are open in an editor then the editor is kept up to date with the version on disk.

simonwhite71:
Oh good, glad I could help, any idea when it might be resolved (not an issue for now, we can always branch manually)

ianhi:
Fix is here: https://github.com/jupyterlab/jupyterlab-git/pull/829

So waiting on review of that - then we can do a patch release. You can also downgrade to `0.22.x` or maybe even `0.23.0`

simonwhite71:
Thats great, thanks for being so responsive, I will wait for the next patch release 👍

PR #829, opened 2020-11-19
don't check for changed files when checking out a new branch

ianhi:
Checking out a new branch will never cause files to change so we don't need to to check changed files. The attempt to check was failing because we were asking for the git diff between the current branch and one that did not yet exist

Fixes: https://github.com/jupyterlab/jupyterlab-git/issues/828

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -384,15 +384,17 @@ export class GitExtension implements IGitExtension {
       'git:checkout',
       async () => {
         let changes;
-        if (body.checkout_branch) {
-          changes = await this._changedFiles(
-            this._currentBranch.name,
-            body.branchname
-          );
-        } else if (body.filename) {
-          changes = { files: [body.filename] };
-        } else {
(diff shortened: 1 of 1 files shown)


fcollonval:
@ianhi Thanks a lot for the quick fix.

Would you have time to add unit tests on this feature and the one addressed by #824 to reduce potential troubles in the future?

it would make sense to put them in https://github.com/jupyterlab/jupyterlab-git/blob/master/tests/model.spec.tsx#L299. I think you could assert internal calls using spy like https://github.com/jupyterlab/jupyterlab-git/blob/master/tests/model.spec.tsx#L365.

ianhi:
Will do. Thanks for the pointers on how to do it, makes it much more approachable.

fcollonval (approved):
LGTM

I merge this one to cut a patch release. But if you have time to add some tests in another PR, that would be much appreciate.

Issue #842, opened 2020-12-13
Handle not in git repository without error 500 to avoid confusion


jeanmonet:
## Description

Hi. git extension seems to output a rather ugly error message in the bash console when the current working directory is not a git repository, or when switching to such a directory:

With debug mode on, the error seems to be triggered by the following:

```
[D 20:39:53.173 LabApp.jupyterlab_git] Execute ['git', 'rev-parse', '--show-toplevel'] in /opt/files/jyusr.
[D 20:39:53.187 LabApp.jupyterlab_git] Code: 128
    Output: 
    Error: fatal: not a git repository (or any of the parent directories): .git
```

which is followed by the following (showing in the console even without debug mode):

```
[E 20:39:53.187 LabApp] {
      "Host": "127.0.0.1:8888",
      "User-Agent": "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:83.0) Gecko/20100101 Firefox/83.0",
      "Accept": "*/*",
      "Accept-Language": "en-US,en;q=0.5",
      "Accept-Encoding": "gzip, deflate, br",
      "Referer": "https://127.0.0.1:8888/lab?",
      "Content-Type": "text/plain;charset=UTF-8",
      "X-Xsrftoken": "2|0d88b5b9|xxx|1607891023",
      "Origin": "https://127.0.0.1:8888",
      "Content-Length": "24",
      "Connection": "keep-alive",
      "Cookie": "username-127-0-0-1-8888=\"2|1:0|10:1607891028|23:username-127-0-0-1-8888|44:xxx=|xxxx\"; _xsrf=2|0d88b5b9|xxx|1607891023",
      "Pragma": "no-cache",
      "Cache-Control": "no-cache"
    }
[E 20:39:53.188 LabApp] 500 POST /git/show_top_level?1607891993159 (127.0.0.1) 23.65ms referer=https://127.0.0.1:8888/lab?
```

Am I doing anything wrong or is this expected behavior?

## Reproduce

Launch Jupyter Lab. Working directory is not a git repository.

## Expected behavior

Should not show anything even if I'm not in a git repo directory.

## Context

```
jupyter                   1.0.0                      py_2    conda-forge
jupyter-lsp               0.9.3                    pypi_0    pypi
jupyter_client            6.1.7                      py_0    conda-forge
jupyter_console           6.2.0                      py_0    conda-forge
jupyter_core              4.7.0            py38h578d9bd_0    conda-forge
[...]

PR #855, opened 2021-01-09
Avoid 500 error messages when current working directory in Jupyter Lab is not a git repo

jeanmonet:
Discussed: #842

Proposal to avoid long 500 error messages in terminal when current working directory in Jupyter Lab is not a git repository:
```
[E 20:39:53.188 LabApp] 500 POST /git/show_top_level?1607891993159 (127.0.0.1) 23.65ms referer=https://127.0.0.1:8888/lab?
```

Instead, print a message to terminal (`self.log.info`) of the git command and of the error message:

```
[I 21:54:40.644 LabApp] [jupyterlab_git] Failed to execute 'git' command: git rev-parse --show-toplevel, with message: fatal: not a git repository (or any of the parent directories): .git
```

Cheers

### DIFF
--- a/jupyterlab_git/handlers.py
+++ b/jupyterlab_git/handlers.py
@@ -74,7 +74,7 @@ class GitAllHistoryHandler(GitHandler):
         history_count = body["history_count"]
 
         show_top_level = await self.git.show_top_level(current_path)
-        if show_top_level["code"] != 0:
+        if show_top_level.get("top_repo_path") is None:
             self.set_status(500)
             self.finish(json.dumps(show_top_level))
         else:
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -684,7 +684,7 @@ class Git:
         else:
             # Handle special case where cwd not inside a git repo
             lower_error = my_error.lower()
-            if "fatal: not a git repository (or any" in lower_error:
+            if "fatal: not a git repository" in lower_error:
                 return {"code": 0, "top_repo_path": None}
 
             return {
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -682,6 +682,11 @@ class Git:
         if code == 0:
             return {"code": code, "top_repo_path": my_output.strip("\n")}
         else:
+            # Handle special case where cwd not inside a git repo
+            lower_error = my_error.lower()
+            if "fatal: not a git repository (or any" in lower_error:
+                return {"code": 0, "top_repo_path": None}
+
             return {
                 "code": code,
                 "command": " ".join(cmd),
--- a/jupyterlab_git/tests/test_handlers.py
+++ b/jupyterlab_git/tests/test_handlers.py
@@ -28,7 +28,7 @@ def test_mapping_added():
 @patch("jupyterlab_git.handlers.GitAllHistoryHandler.git", spec=Git)
 async def test_all_history_handler_localbranch(mock_git, jp_fetch):
     # Given
-    show_top_level = {"code": 0, "foo": "top_level"}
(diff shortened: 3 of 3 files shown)


jupyterlab_git/handlers.py
```diff
@@ -27,6 +27,36 @@ class GitHandler(APIHandler):
...
+        )
+        lower_error = result.get("message", "").lower()
+        if any([msg in lower_error for msg in error_message]):
+            log_lvl = log_lvl if log_lvl in ["info", "warning", "error"] else "info"
+            getattr(self.log, log_lvl)(
+                "[jupyterlab_git] Failed to execute 'git' command: {!s},"
+                " with message: {!s}".format(result["command"], result["message"].strip())
+            )
+            return True
+        return False
```
fcollonval, line 58: Could you move this after the command execution https://github.com/jupyterlab/jupyterlab-git/blob/90cfb96d5f029badf85e834b5573f9871c85b58f/jupyterlab_git/git.py#L682

Then if the error is the expected, set the returned value to:

```
{"code": code, "top_repo_path": None}
```

Could you send the error message on debug level? Basically we try to be as quiet as possible in default mode.

jupyterlab_git/handlers.py
```diff
@@ -75,7 +105,8 @@ async def post(self):
 
         show_top_level = await self.git.show_top_level(current_path)
         if show_top_level["code"] != 0:
-            self.set_status(500)
+            if not self.git_error_no_500(show_top_level):
+                self.set_status(500)
```
fcollonval, line 109:
```suggestion
            self.set_status(500)
```

jupyterlab_git/handlers.py
```diff
@@ -75,7 +105,8 @@ async def post(self):
 
         show_top_level = await self.git.show_top_level(current_path)
         if show_top_level["code"] != 0:
```
fcollonval, line 107:
```suggestion
        if show_top_level.get("top_repo_path") is None:
```

This handler needs to be removed (it is not called anymore). So we should continue raising an error to be sure the other missing data are not wrongly handled.

fcollonval (changes requested):
Thanks @jeanmonet 

I would prefer to reorganize the code to handle the special case in the `show_top_level` method directly. I provide a few suggestion. It would be great if you can include them.

Regarding `GitAllHistoryHandler`, don't worry too much about it. It is not used and should be removed when I get the time to rewrite the REST API.

jupyterlab_git/handlers.py
jeanmonet: Thanks, I'll revert this week!

jeanmonet: ahh just to say I didn't manage this week and will be away for a few days, but will be able so sit on this next weekend 👍

jupyterlab_git/handlers.py
jeanmonet: Done in new commit 0cdba5b: returns code 0 with top_repo_path None

fcollonval (approved):
Thanks @jeanmonet 

I simplified the check for error as there is only one error message to test. And I took this opportunity to add unit tests too.

jeanmonet:
Awesome!

One final touch to consider (now or later), as I was looking at https://github.com/git/git/search?q=%22Not+a+git+repository%22

Maybe the error string `fatal: not a git repository (or any` should just be `fatal: not a git repository` to avoid having a mismatch in an unforeseen case where git appends something other than `(or any`... at the end of the message (although I'm not sure when that would happen).

fcollonval:
The error is not related - so I will change the error message then merge. Thanks again.

fcollonval:

PR #863, opened 2021-01-27
Backward compatibility for server extension on jupyterhub

SarunasAzna:
Server extension was not working with jupyterlab 3.0 runned on jupyterhub (version1.2.2, to be more exact helm charts version 0.10.5 from here https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/CHANGELOG.md) though it was working fine with exact docker image runned as standalone container.

I cannot point to the exact problem but the fix itself is copied (and used on another internal server extension) from Voila https://github.com/voila-dashboards/voila.


fcollonval (approved):
Thanks @SarunasAzna 

Although I'm sad that this is required, it eases the upgrade for most users. So let's merge it.

meeseeksmachine:

https://discourse.jupyter.org/t/the-server-extension-appears-to-be-missing-when-launching-with-mybinder-org/9910/4

Issue #864, opened 2021-01-29
Git history for a selected file 


adammclaurin:
It's no secret that using GitHub directly isn't great for viewing changes to Jupyter notebooks. jupyterlab-git offers some promise to make that better via integration with nbdime, but unfortunately it only seems to work by going through the history of the whole repo, which might have many changes unrelated to the file you're interested in. It'd be amazing to have a way to select a particular file to explore the history for just that one file, again with nbdime integration to be able to see the diffs clearly.

tcapelle:
For github, you can use the NBReview app, it makes pull requests so much easier.

ianhi:
For future reference the git command for this is:
```
git log --follow -p -- path-to-file
```

(see https://stackoverflow.com/a/5493663/835607)

So this will require:
1. Enabling that log command in the backend
2. Figuring out a good UI on the frontend.

fcollonval:
Thanks @ianhi for the pointer

>     2\. Figuring out a good UI on the frontend.

While avoiding components multiplication 😉

adammclaurin:
If we do implement this feature, it'd be amazing to be able to optionally ignore changes to output cells when diff'ing Jupyter notebooks.

Issue #379, opened 2019-07-10
[UX] Unable to select multiple files to stage at once


dhirschfeld:
It appears to only be possible to select a single file to stage so if you have 10 files to stage you have to click a minimum of 10 times to stage each of them.

![image](https://user-images.githubusercontent.com/881019/60932175-235aee00-a300-11e9-942a-fa40cdb56d43.png)

It should be possible to select multiple files using the standard `Crtl-Click` and `Shift-Click` methods

PR #877, opened 2021-02-20
Implement Git context menu in the file browser

### LINKED ISSUE #774: "git add" from file browser?
## Description

I would like to be able to add a file from the file browser, rather than from the list of untracked filed (it is long for me on some projects, and it is not organised into directories thus it is difficult to find the specific file I want to add). However, I always have this file located in the file browser so it would be of great help if I could just add it from the context menu of the item listing of files in the built-in file browser.

krassowski:
Fixes #774

Changes:
- [x] Added Git context menu in File Browser
- [x] Reworked the context commands to accept multiple files (as multiple files can be selected in the file browser)
- [x] Added icons to the commands (they also display in the context menu of the Git panel widget)
- [x] The old context menu and new context menu share the same commands and the same status assignments (the "open"  command is not shown in the file browser as redundant to the browser capabilities); EDIT: "delete" is no longer shown either (GIF was not updated)
- [x] Fixed a typo (`_getPathRespository`)
- [x] Formalize types:
  - for context commands (as distinct from other commands): `ContextCommandIDs`
  - for context commands' arguments (as those were already being cast in type-unsafe way and I needed to change them all): `CommandArguments`

jupyterlab-git-context-menu

The Git panel context menu got icons (updated):

context-menu-in-git-panel

### DIFF
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -852,7 +852,7 @@ export function addFileBrowserContextMenu(
       const items = getSelectedBrowserItems();
       const statuses = new Set<Git.Status>(
         items
-          .map(item => model.getFileStatus(item.path))
+          .map(item => model.getFile(item.path)?.status)
           .filter(status => typeof status !== 'undefined')
       );
 
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -736,11 +736,7 @@ export function addCommands(
 /**
  * Adds commands and menu items.
  *
- * @private
- * @param app - Jupyter front end
- * @param gitExtension - Git extension instance
- * @param fileBrowser - file browser instance
- * @param settings - extension settings
+ * @param commands - Jupyter App commands registry
  * @returns menu
  */
 export function createGitMenu(commands: CommandRegistry): Menu {
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -505,6 +505,21 @@ export function addCommands(
     icon: diffIcon
   });
 
+  commands.addCommand(ContextCommandIDs.gitFileAdd, {
+    label: 'Add',
+    caption: pluralizedContextLabel(
+      'Stage or track the changes to selected file',
+      'Stage or track the changes of selected files'
+    ),
+    execute: async args => {
+      const { files } = (args as any) as CommandArguments.IGitContextAction;
+      for (const file of files) {
+        await model.add(file.to);
+      }
+    },
+    icon: addIcon
+  });
+
   commands.addCommand(ContextCommandIDs.gitFileStage, {
     label: 'Stage',
     caption: pluralizedContextLabel(
@@ -868,6 +883,13 @@ export function addFileBrowserContextMenu(
               command !== ContextCommandIDs.gitFileOpen &&
               command !== ContextCommandIDs.gitFileDelete &&
               typeof command !== 'undefined'
+          )
+          // replace stage and track with a single "add" operation
+          .map(command =>
+            command === ContextCommandIDs.gitFileStage ||
+            command === ContextCommandIDs.gitFileTrack
+              ? ContextCommandIDs.gitFileAdd
+              : command
           )
       );
 
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -591,8 +591,8 @@ export function addCommands(
         title: 'Delete Files',
         body: (
           <span>
-            Are you sure you want to permanently delete
-            {fileList}? This action cannot be undone.
+            Are you sure you want to permanently delete {fileList}? This action
+            cannot be undone.
           </span>
         ),
         buttons: [Dialog.cancelButton(), Dialog.warnButton({ label: 'Delete' })]
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -15,7 +15,7 @@ import { ISettingRegistry } from '@jupyterlab/settingregistry';
 import { ITerminal } from '@jupyterlab/terminal';
 import { CommandRegistry } from '@lumino/commands';
 import { Menu } from '@lumino/widgets';
-import { toArray, ArrayExt } from '@lumino/algorithm';
+import { ArrayExt, toArray } from '@lumino/algorithm';
 import * as React from 'react';
 import {
   Diff,
@@ -710,6 +710,12 @@ export function addCommands(
       });
     }
   });
+
+  commands.addCommand(ContextCommandIDs.gitNoAction, {
+    label: 'No actions available',
+    isEnabled: () => false,
+    execute: () => void 0
+  });
 }
 
 /**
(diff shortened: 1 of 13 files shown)


krassowski:
Nice binder badge action. I am not sure what rank should the Git menu have (should it be lower or higher?). I just randomly used 5 and it seems like a good position for me but might be sub-optimal.

src/tokens.ts
```diff
@@ -821,3 +833,35 @@ export interface ILogMessage {
    */
   message: string;
 }
+
+/**
+ * The command IDs used in the git context menus.
+ */
+export enum ContextCommandIDs {
```
krassowski, line 840: Enums need to be in a separate file due to a bug in Typescript which causes them to go awry in preference of cyclic imports, birth they do not need to be in tokens if you prefer them elsewhere

krassowski:
This is also a significant step towards #379.

ianhi:
> The Git panel context menu got icons:

:heart: 

> The old context menu and new context menu share the same commands and the same status assignments (the "open" command is not shown in the file browser as redundant to the browser capabilities)

Perhaps the git menu shouldn't have delete as that is also redundant?

Also this made me realize that maybe for `delete` on untracked files we should use the same `x` as the file browser rather than the `-` that is used for unstage.

krassowski:
I changed the icon for delete, I think that it was me who wrongly assigned it and it was not used like that before. I also agree that delete should not be shown in the file browser context menu - changed it now too.

Also now tested that it works with git repository which is not in the JupyterLab root:

works-if-git-not-a-top-repo

Outside of a git repository:

outside-of-git-repo

Or within the git repository but when no action is available (which happens for files that are tracked but not changed, as jupyterlab-git does not implement `git rm` nor any other relevant command):

known-file-no-changes

src/tokens.ts
fcollonval: tokens.ts is fine for me.

fcollonval:
Thanks for this awesome PR @krassowski 

The change for the delete icon is welcomed indeed.

>  I am not sure what rank should the Git menu have (should it be lower or higher?)

I don't have any preferences neither. The current position seems ok for me too.

Some comments:
- In the Git panel context menu, it looks like the icon are not vertically center with the text (they are a bit up). Is it possible to correct that?
- I have an error when opening the diff for multiple files at once. Only one diff is shown correctly - note: I tested this only with plain text files not with notebooks
- If multiple files are selected, some being untracked and some being tracked, two entries appears in the context menu - Stage and Track. But if I click on Stage they all get added and/or staged. I think this is nice to do the action for all files - but it maybe good to have only one entry in the context menu.

krassowski:
> In the Git panel context menu, it looks like the icon are not vertically center with the text (they are a bit up). Is it possible to correct that?

I see that too. I was also looking at the `trashIcon` where the problem is even worse. It seems that the icons were fine-tuned for their use in the GitPanel rather than placed in the center and then fine tuned in the GitPanel via CSS. I do not know which makes more sense:
- to fine-tune position in menu via bindattrs/css
- to correct icons to be all the same size and centered and then adjust their use in GitPanel

> I have an error when opening the diff for multiple files at once. Only one diff is shown correctly - note: I tested this only with plain text files not with notebooks

Yes, I can reproduce that. I believe that it is not an effect of the changes that I made but some limitation of the diff renderer. I was also seeing this bug when I was opening multiple diffs manually in previous releases. My thinking is that either:
- the tab with diff needs to be focused
- the stream from server breaks when another diff gets open

But those are just guesses. Would be great to investigate and fix (possibly in another PR).

> If multiple files are selected, some being untracked and some being tracked, two entries appears in the context menu - Stage and Track. But if I click on Stage they all get added and/or staged. I think this is nice to do the action for all files - but it maybe good to have only one entry in the context menu.

I think that it would be good to have all actions that apply to at least one of the files. Imagine a scenario of working with several icons: you select 10 files of which 9 are not staged and 1 is already staged; you want to stage all but the staged one is in the middle. Instead of doing two selections you can do it once. Similarly for unstage: you have scattered unstaged files among a group of staged files in a specific directory and want to unstage all of the staged one, so you can just select a range without worrying about which ones where staged and which were not.

adammclaurin:
Does this also resolve #864?

krassowski:
No, there is no context command for "show file history" just yet. Another main area widget with history browser would need to be implemented first I think...

But what this PR does is making it super easy to add such a command to the context menu once it gets contributed.

fcollonval:
> Would be great to investigate and fix (possibly in another PR).

Agree I opened #881 

What is your opinion on the third point I raised:

> If multiple files are selected, some being untracked and some being tracked, two entries appears in the context menu - Stage and Track. But if I click on Stage they all get added and/or staged. I think this is nice to do the action for all files - but it maybe good to have only one entry in the context menu.

src/commandsAndMenu.tsx
```diff
@@ -368,168 +387,216 @@ export function addCommands(
...
+  function representFiles(files: Git.IStatusFile[]): JSX.Element {
+    if (files.length > 1) {
+      const elements = files.map(file => (
+        <li key={file.to}>
+          <b>{file.to}</b>
+        </li>
+      ));
+      return <ul>{elements}</ul>;
+    } else {
+      return <b>{files[0].to}</b>;
```
krassowski, line 546: TODO: double check if there is a space between action and file name upon rendering (there was not prior to this PR for delete action).

krassowski:
> What is your opinion on the third point I raised:
> 
> If multiple files are selected, some being untracked and some being tracked, two entries appears in the context menu - Stage and Track. But if I click on Stage they all get added and/or staged. I think this is nice to do the action for all files - but it maybe good to have only one entry in the context menu.

Yes, the distinction between stage and track is artificial. But it appears to me that the current logic is neat and maintainable (aggregate the commands from context menus for specific files and show all unique entries). I can however, add extra logic which replaces the two with a tailored "Add" command.

fcollonval:
> But it appears to me that the current logic is neat and maintainable (aggregate the commands from context menus for specific files and show all unique entries). I can however, add extra logic which replaces the two with a tailored "Add" command.

Ok let keep it simple for now.

krassowski:
I merged then in https://github.com/jupyterlab/jupyterlab-git/commit/3cb610972bd3954c1a084c5cdf8bf695021ccfca

src/model.ts
```diff
@@ -242,6 +242,26 @@ export class GitExtension implements IGitExtension {
...
 
+  /**
+   * Match files status information based on a provided file path.
+   *
+   * If the file is tracked and has no changes, undefined will be returned
+   *
+   * @param path the file path relative to the server root
+   */
+  getFile(path: string): Git.IStatusFile {
+    const matchingFiles = this._status.files.filter(status => {
```
fcollonval, line 253: Would it be more efficient to use find as only the first match is returned?

```suggestion
    const matchingFiles = this._status.files.find(status => {
```

fcollonval (approved):
I have a very small performance suggestion. Otherwise LGTM

ianhi:
I agree, this is a great thing. One thing to consider in the future would be to add an icon for git ignore so the ignore options don't look out of place in the context menu.

telamonian (approved):
This looks great @krassowski!

My only nit is the `track` item. In all cases (ie `track` and `add`/`stage`) what you're really doing is staging the file, so it's confusing to call it anything else. I know I was confused for a second when I saw `track` pop up in your first gif.

krassowski:
`Track`/`Stage` was collapsed into a single `Add` in the browser menu, the top gif was not updated There still is a distinction in the git panel context menus (as it was there before)

fcollonval:
Thanks @krassowski

fcollonval:

meeseeksmachine:

https://discourse.jupyter.org/t/what-does-ifilebrowserfactory-defaultbrowser-do/9702/2

PR #878, opened 2021-02-20
Allow to commit with ctrl + enter

### LINKED ISSUE #773: Ctrl + enter to commit
##  Description

I would like to be able to commit by pressing <kbd>Ctrl</kbd> + <kbd>Enter</kbd> when the focus is on Summary or Description field. Technically, I would propose to reuse the shortcut that is used by the notebook to run the cell (as different operating systems may have different bindings for such actions).

## Context

- Extension version:
@jupyterlab/git v0.21.1  enabled  OK

krassowski:
Fixes #773

Also:
- replaces deprecated `which` with `key`. Some browsers already dropped `which` (see the big red box on [MDN]( https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which)
- Remove incorrect  `@param event - event object` from `_onCommitClick` documentation
- Use appropriate type for `_onSummaryKeyPress` `event` argument

### DIFF
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -87,6 +87,8 @@ export namespace CommandIDs {
   export const gitIgnoreExtension = 'git:context-ignoreExtension';
 }
 
+export const SUBMIT_COMMIT_COMMAND = 'git:submit-commit';
+
 /**
  * Add the commands for the git extension.
  */
@@ -98,6 +100,22 @@ export function addCommands(
   renderMime: IRenderMimeRegistry
 ) {
   const { commands, shell } = app;
+
+  /**
+   * Commit using a keystroke combination when in CommitBox.
+   *
+   * This command is not accessible from the user interface (not visible),
+   * as it is handled by a signal listener in the CommitBox component instead.
+   * The label and caption are given to ensure that the command will
+   * show up in the shortcut editor UI with a nice description.
+   */
+  commands.addCommand(SUBMIT_COMMIT_COMMAND, {
+    label: 'Commit from the Commit Box',
+    caption:
+      'Submit the commit using the summary and description from commit box',
+    execute: () => void 0,
+    isVisible: () => false
+  });
 
   /**
    * Add open terminal in the Git repository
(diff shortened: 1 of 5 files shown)


fcollonval (changes requested):
Thanks @krassowski 

What about adding a note in the summary placeholder like in VSCode to give a hint about the shortcut?

image

krassowski:
The `Summary` placeholder already has a text in parentheses `(required)`. Should it be replaced, or should the hint be added to the `Description` placeholder?

Also, it would require:
- using Accel instead of Control and then
- importing the key representation converter utility from lumino or writing a custom one (so that Mac users do not see Control in their interface which is irrelevant to them).

fcollonval:
> The `Summary` placeholder already has a text in parentheses `(required)`. Should it be replaced, or should the hint be added to the `Description` placeholder?

Good point. What about replacing the _required_ by the helper text and add _optional_ in the description placeholder?

> Also, it would require:
> 
>     * using Accel instead of Control and then
>     * importing the key representation converter utility from lumino or writing a custom one (so that Mac users do not see Control in their interface which is irrelevant to them).

If you can do that with the lumino utility this will be awesome.

src/commandsAndMenu.tsx
```diff
@@ -87,6 +87,8 @@ export namespace CommandIDs {
   export const gitIgnoreExtension = 'git:context-ignoreExtension';
 }
 
+export const SUBMIT_COMMIT_COMMAND = 'git:submit-commit';
```
krassowski, line 90: Happy to add to the enum in #877, did not add to namespace to avoid annoying conflicts.

fcollonval: Yes please do.

fcollonval (approved):
LGTM !

I let you move the new command in the enum in #877

fcollonval:

krassowski:
> I let you move the new command in the enum in #877

Done in https://github.com/jupyterlab/jupyterlab-git/pull/877/commits/f5df35767d9e47bfa7ea8b1cf3a9edd03263a04c (and ready for review).

PR #880, opened 2021-02-22
Force change the repository to a current one

### LINKED ISSUE #876: Extension forgets it is in a git directory when moving to opposite side bar
## Description

Moving the extension from one sidebar to another causes it to forget that the current directory is git-backed, and it will not recover until moving up above the current directory and back in.

## Reproduce

From a git-backed directory:
* left click extension (confirm it shows git-backed directory)
* right click on the extension in left side bar and move to other sidebar
* (extension will show a non-git-backed directory)
* Browsing in file browser deeper into path will not resolve
* Browsing up in filetree until we're out of the git dir, then back into it, will make things work again

## Expected behavior

That extension does not lose track of it being in a git directory when switching side bar side, or at least can quickly recover.

ktaletsk:
Fixes #876.

### DIFF
--- a/src/components/GitPanel.tsx
+++ b/src/components/GitPanel.tsx
@@ -122,7 +122,7 @@ export class GitPanel extends React.Component<IGitPanelProps, IGitPanelState> {
       nCommitsAhead: 0,
       nCommitsBehind: 0,
       pastCommits: [],
-      repository: null,
+      repository: props.model.pathRepository,
       tab: 0
     };
   }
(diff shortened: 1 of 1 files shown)


fcollonval:
Thanks @ktaletsk 

When testing the PR I realized that the current branch and the branch list are also affected. From the state:

```
      branches: [], // Affected
      currentBranch: '', // Affected
      files: [], // Not affected
      nCommitsAhead: 0,  // TBC not affected
      nCommitsBehind: 0, // TBC not affected
      pastCommits: [], // Not affected
      repository: null, // Affected
      tab: 0 // Not affected
```

In light of that, would it work to initialize the state in the constructor with the model values?

ktaletsk:
Hi @fcollonval, that seems to work just fine and fixes the bug. I moved `repository` assignment from `componentDidMount` to constructor and initialized current branch and the branch list from model there as well.

fcollonval (approved):
Awesome thanks a lot @ktaletsk for the quick fix

fcollonval:

Issue #837, opened 2020-11-30
Allow `commit & push`


fcollonval:
To help git beginners (but expert too :wink:), it would be nice to have a setting to transform the commit button in `commit & push` button (if there is a remote...).

adammclaurin:
Definitely a good idea!

ktaletsk:
Hey @fcollonval @adammclaurin , I really like this idea!

I just opened a PR implementing this feature, hope it would help. Seems to work for my limited testing, please take a look too.

fcollonval:
Fixed by #887

PR #887, opened 2021-02-27
Allow `commit & push`

ktaletsk:
Implements #837 

So, here's my take on enabling 'Commit and push' in place of 'Commit' button. I pass branch information along with boolean flag to GitCommitBox so that it can render 'Commit and Push' or display the error hint when applicable. Push action is added on the back of `GitPanel.commitStagedFiles`. There is also a new flag in extension settings to enable/disable this behavior.

Please, check the code and if it doesn't make sense, I am open to rewriting it.


src/components/CommitBox.tsx
```diff
@@ -87,7 +98,11 @@ export class CommitBox extends React.Component<
       ? 'Disabled: No files are staged for commit'
       : !this.state.summary
       ? 'Disabled: No commit message summary'
-      : 'Commit';
+      : !this.props.useCommitAndPush
+      ? 'Commit'
+      : !this.props.branches.some(branch => branch.is_remote_branch) 
+      ? 'Disabled: No remote repository defined'
+      : 'Commit and push';
```
fcollonval, line 105: As the `onCommit` action comes from a property, this component is blind to the actions taking place. So a better design is to add the button label as property of the `CommitBox` and handle the logic in `GitPanel`.

src/components/CommitBox.tsx
```diff
@@ -128,7 +145,10 @@ export class CommitBox extends React.Component<
    * Whether a commit can be performed (files are staged and summary is not empty).
    */
   private _canCommit(): boolean {
-    return !!(this.props.hasFiles && this.state.summary);
```
fcollonval, line 131: This should not be changed as commit should be available on repository without remote even if the settings `useCommitAndPush` is true => it should fallback on commit only.

fcollonval (changes requested):
Thanks @ktaletsk 

I added some suggestion to have a better API.

fcollonval (approved):

fcollonval:

PR #891, opened 2021-03-06
Update README.md:Git panel is not visible

EnisBerk:
Even if the package is installed, it does not work unless it is in the uninstalled packages list. It is caused by a bug in jupyterlab according to this issue https://github.com/jupyterlab/jupyterlab/issues/8122

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,8 @@ If they do not match or one is missing, please [reinstall the package](README.md
     jupyter labextension install @jupyterlab/git
     ```
     
-    If you see `@jupyterlab/git` under the  `Uninstalled core extensions: ` remove jupyter\lab\settings\build_config.json and re-install.
+    If you see `@jupyterlab/git`, your installation may have been corrupted. You can run `jupyter lab clean --all` and
+    reinstall all your extensions.
     
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -137,7 +137,9 @@ If they do not match or one is missing, please [reinstall the package](README.md
     ```bash
     jupyter labextension install @jupyterlab/git
     ```
-
+    
+    If you see `@jupyterlab/git` under the  `Uninstalled core extensions: ` remove jupyter\lab\settings\build_config.json and re-install.
+    
 ## Contributing
 
 If you would like to contribute to the project, please read our [contributor documentation](https://github.com/jupyterlab/jupyterlab/blob/master/CONTRIBUTING.md).


krassowski:
Did you encounter this with JupyterLab 2.x or 3.0?

EnisBerk:
The problem started happening with JupyterLab 2.x, upgrading to 3.0 did not help. But I did not try a clean install of JupyterLab 3.0.

fcollonval:
Hey @EnisBerk did you try running `jupyter lab clean --all` and then `jupyter lab build` ? 

I would rather avoid advertising for quick fix requiring to touch JupyterLab internal building files.

EnisBerk:
Hi @fcollonval, I understand your concern.  I did not try it before, I just tested with 3.0  and `jupyter lab clean --all` also removes the same file. So it solves the problem. I do not know why I did not see that suggestion anywhere else. Maybe you can add that to the ReadMe, it is up to you.

fcollonval:
Thanks for the quick reply.

> I just tested with 3.0 and `jupyter lab clean --all` also removes the same file. So it solves the problem. 

Greate

> I do not know why I did not see that suggestion anywhere else. Maybe you can add that to the ReadMe, it is up to you.

So I'll modify this PR in that sense.

README.md
```diff
@@ -138,7 +138,8 @@ If they do not match or one is missing, please [reinstall the package](README.md
     jupyter labextension install @jupyterlab/git
     ```
     
-    If you see `@jupyterlab/git` under the  `Uninstalled core extensions: ` remove jupyter\lab\settings\build_config.json and re-install.
+    If you see `@jupyterlab/git`, your installation may have been corrupted. You can run `jupyter lab clean --all` and
```
ianhi, line 141: It looks like the first part of the sentence is missing some words now? 

"If you see `@jupyterlab/git` in ____" perhaps?

EnisBerk: Here is the output of the corrupted setup, @jupyterlab/git  is installed but also reported under the `Uninstalled core extensions: `, the sentence can be phrased as  "If you see @jupyterlab/git in Uninstalled core extensions ..."  
JupyterLab v3.0.9
/anaconda3/envs/jupyter38/share/jupyter/labextensions
        @jupyterlab/git v0.30.0-beta.2 enabled OK (python, jupyterlab-git)
        @jupyter-widgets/jupyterlab-manager v3.0.0 enabled OK (python, jupyterlab_widgets)

Uninstalled core extensions:
    @jupyterlab/git

fcollonval: The idea was to get the negation of 135. But definitely it is not clear enough. Should we restrain it to the _uninstalled_ case of @EnisBerk or use `jupyter lab clean --all` as a fallback in abnormal case.

PR #893, opened 2021-03-06
Use animated icon as status indicator

### LINKED ISSUE #803: UX: Bottom Bar Distracting
## Description

Hi,

The bottom toolbar is distracting as it keeps pushing things to the right for a second then back. Below shows Python3 moving over:
idling

Given this happens every couple seconds, I don't think it's a net positive to users to have anything shown when it is working correctly. If people feel strongly on keeping this, can we ensure that it takes up always the same amount of space?

## Context

jupyterlab-git@0.21.1

fcollonval:
Fixes #803


fcollonval:
@mlucool @ianhi Happy to get your feedback on this one.

mlucool:
Thanks for the tag @fcollonval.

I really like the icon for refresh. While this is _much_ better than the previous version, I still find the movement distracting. Can we invert this to only call attention to it when something is wrong or it's doing work triggered by a user? For example, automatically refreshing status is not work but committing a notebook is.

krassowski:
It might be still a bit distracting as discussed above, but aside from that I see a quirk with the icon jumping to the right when switching the panels (I assume it was not intended):

jitter

krassowski:
It seems that the issue that I am seeing is not with jupyterlab-git but with the core logs icon which is hidden (`.jp-LogConsoleStatusItem`) and has its classList populated with some delay after switching panels:

the-problem

krassowski:
I opened an issue upstream.

fcollonval:
> I still find the movement distracting. 

I switch to fill opacity animation to be less catchy.

> Can we invert this to only call attention to it when something is wrong or it's doing work triggered by a user? For example, automatically refreshing status is not work but committing a notebook is.

The idea was to get the status bar reflecting the git actions and the pop-up to inform of the user actions progression. One of the point was to be able to see if for example _refreshing_ was not finishing - this is by essence hard to automate as an arbitrary timeout changing the icon will be unable to take into account poor network for instance.

So I would rather keep the current logic but make it more discrete.

fcollonval:

PR #894, opened 2021-03-06
Fill out details in troubleshooting

ianhi:
https://github.com/jupyterlab/jupyterlab-git/pull/891#discussion_r588900207

@fcollonval @EnisBerk is that what was intended?

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ If they do not match or one is missing, please [reinstall the package](README.md
     jupyter labextension install @jupyterlab/git
     ```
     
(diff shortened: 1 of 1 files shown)


fcollonval:
Duplicating here my comment as I saw the PR after commenting 😞 

The idea was to get the negation of 135. But definitely it is not clear enough. Should we restrain it to the uninstalled case of @EnisBerk or use jupyter lab clean --all as a fallback in abnormal case.

fcollonval (approved):

PR #899, opened 2021-03-11
Restore preview gif

ianhi:
It seems to have been accidentally removed in https://github.com/jupyterlab/jupyterlab-git/pull/818


fcollonval (approved):
Good catch thanks

PR #900, opened 2021-03-14
Improve Diff

### LINKED ISSUE #666: Adding diff provider not working

fcollonval:
Fixes #666

The diff components are back to pure Lumino widget. This suppress the current complexity of putting non-react code (nbdime and codemirror) in react in lumino panel.

The new structure use really the diff provider registry. And implemented file comparison with the new widget should be really easy.

TODO
- [ ] Fixe styles
- [ ] Connect settings and code mirror in plain text diff
- [ ] Add endpoint to compute nbdiff from two notebook content


ianhi:
<3

fcollonval:

PR #902, opened 2021-03-17
Fix pip check in binder

fcollonval:


fcollonval:
Waiting on https://github.com/conda-forge/sqlalchemy-feedstock/pull/64 to remove any unrelated constrains

PR #905, opened 2021-03-18
handle localization of plurals

JessicaBarh:


src/commandsAndMenu.tsx
```diff
@@ -716,11 +716,11 @@ export function addCommands(
       const extensions = files
         .map(file => PathExt.extname(file.to))
         .filter(extension => extension.length > 0);
-      const subject = extensions.length > 1 ? 'extensions' : 'extension';
-      return trans.__(
-        'Ignore %1 %2 (add to .gitignore)',
-        extensions.join(', '),
-        trans.__(subject)
+      return trans._n(
+        'Ignore %2 extension (add to .gitignore)',
```
goanpeca, line 720: This should probably be a `%1` ?

src/commandsAndMenu.tsx
```diff
@@ -716,11 +716,11 @@ export function addCommands(
...
+        'Ignore %2 extensions (add to .gitignore)',
```
goanpeca, line 721: This should probably be a `%1` ?

goanpeca (changes requested):
Thanks @JessicaBarh !

src/commandsAndMenu.tsx
JessicaBarh: using %1 was actually picking up the number arg (extensions.length). I was testing with this: https://github.com/JessicaBarh/jupyterlab-fr_FR/tree/handle-plurals/jupyterlab_language_pack_fr_FR/extensions/jupyterlab-git

goanpeca: You are absolutely right! My bad

fcollonval (approved):
Thanks for the follow-up!

PR #919, opened 2021-04-03
nbdime styles are not imported

fcollonval:
If JupyterLab is not built to integrate nbdime extension, nbdime styles are not loaded and the notebook diff is not pretty.


fcollonval:

PR #930, opened 2021-04-26
Disable 'Open Git Repository in Terminal' if terminals are not available

faucct:
https://github.com/jupyterlab/jupyterlab/blob/f0790b3ffd187b7dc923156b8ae227855abecefb/packages/terminal-extension/src/index.ts#L98


fcollonval (approved):
Thanks @faucct

Issue #958, opened 2021-06-22
Failed to commit, Fail to set your identity. Unhandled error


DarkmatterVale:
## Description

When attempting to commit from the Git UI in Jupyterlab, I receive an error: `Failed to commit, Fail to set your identity. Unhandled error`

## Reproduce

1. Go to the Git extension
2. Stage some changes
3. Fill out the commit name & description
4. Click commit
5. The following error appears:

```
HTTPServerRequest(protocol='http', host='...', method='POST', uri='/user/.../git/test123/config?...', version='HTTP/1.1', remote_ip='...')
    Traceback (most recent call last):
      File "/usr/local/lib64/python3.7/site-packages/tornado/web.py", line 1704, in _execute
        result = await result
      File "/usr/local/lib/python3.7/site-packages/jupyterlab_git/handlers.py", line 650, in post
        options = data.get("options", {})
    AttributeError: 'NoneType' object has no attribute 'get'
```

## Expected behavior

I am able to commit changes to the git repository.

## Context

- Python package version: 0.31.0a0
- Extension version: v0.31.0-alpha.0
- Git version: 2.23.4
- Operating System and its version: Amazon Linux 2

I've traced the error back to https://github.com/jupyterlab/jupyterlab-git/blob/master/jupyterlab_git/handlers.py#L649. As specified at https://github.com/jupyterlab/jupyterlab-git/blob/master/src/model.ts#L507, the body of the `POST` request can be null if options are not set. This in turn causes the `data = self.get_json_body()` object to return `None`, leading to the error on the following line of code.

If I manually find the offending `/config` request in my browser, and inspect the `body`, I can confirm it is empty. If I resend the request with a body of `{}`, the endpoint returns the expected value.

DarkmatterVale:
I'm aware that there is a similar issue (https://github.com/jupyterlab/jupyterlab-git/issues/498) that already exists. However, the reasons for the error appear to be very different, so I created a new issue.

fcollonval:
Thanks @DarkmatterVale for the detailed issue. Indeed the root cause is not the same as of #498.

Would you be willing to propose a PR?

DarkmatterVale:
I'm happy to submit a PR.

Any preference to the expected behavior? It seems most logical to add a check on https://github.com/jupyterlab/jupyterlab-git/blob/master/jupyterlab_git/handlers.py#L650 to handle the case where `data` is `None`.

fcollonval:
[...]

PR #959, opened 2021-06-22
Prevent empty JSON body from causing None error in git config endpoint

DarkmatterVale:
### Notes
- Address #958 

### Tests
- `pytest jupyterlab_git`, verified all tests pass

### DIFF
--- a/jupyterlab_git/handlers.py
+++ b/jupyterlab_git/handlers.py
@@ -646,8 +646,8 @@ class GitConfigHandler(GitHandler):
         """
         POST get (if no options are passed) or set configuration options
         """
-        data = self.get_json_body()
-        options = data.get("options", {}) if data else {}
+        data = self.get_json_body() or {}
+        options = data.get("options", {})
 
         filtered_options = {k: v for k, v in options.items() if k in ALLOWED_OPTIONS}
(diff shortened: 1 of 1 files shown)


tleonhardt (approved):

jupyterlab_git/handlers.py
```diff
@@ -647,7 +647,7 @@ async def post(self, path: str = ""):
         POST get (if no options are passed) or set configuration options
         """
         data = self.get_json_body()
```
fcollonval, line 649: I would set the default here. That will prevent future error if more information are looked for in data.

```suggestion
        data = self.get_json_body() or {}
```

jupyterlab_git/handlers.py
```diff
@@ -647,7 +647,7 @@ async def post(self, path: str = ""):
...
-        options = data.get("options", {})
+        options = data.get("options", {}) if data else {}
```
fcollonval, line 650:
```suggestion
        options = data.get("options", {})
```

fcollonval (changes requested):
Hey @DarkmatterVale thanks for the fix. I would suggest a slightly more robust approach to prevent this to happen in the future. Let me know what you think.

PR #971, opened 2021-07-06
Preserve commit message when switching to the History tab (#931) (#836)

### LINKED ISSUE #836: Don't erase commit message when commit fails
When using git hook (for example such as black), it may happen that the commit is rejected. Unfortunately in case of error the commit message is erased. It would be nice to keep it as the user will want to carry small correction (like add the black format correction) before retrying to commit.

### LINKED ISSUE #931: Don't erase commit message when switching to the History tab
## Description

When switching to the History tab with a commit summary/description entered, the values get discarded.
Similar to #836

## Reproduce

message-discarded

## Expected behavior

The message is preserved when switching to history and back. I sometimes switch to check what was the message of my previous commit (or do `git log` if in terminal) and do not like to have to type the commit message again.

## Context

- Python package version: 0.30.0
- Extension version: v0.30.0
- Git version: 2.31.1
- Operating System and its version: Ubuntu

navn-r:
When switching to the history tab, the commit description and summary are not erased.
When the commit fails (ex: pre-commit checks), the commit description and summary are not erased.

Fixes #931 
Fixes #836

### DIFF
--- a/src/components/CommitBox.tsx
+++ b/src/components/CommitBox.tsx
@@ -61,7 +61,7 @@ export interface ICommitBoxProps {
   /**
    * Callback invoked upon clicking a commit message submit button or otherwise submitting the form.
    */
-  onCommitSubmit: () => void;
+  onCommitSubmit: () => Promise<void>;
 }
 
 /**
--- a/src/components/CommitBox.tsx
+++ b/src/components/CommitBox.tsx
@@ -45,23 +45,25 @@ export interface ICommitBoxProps {
   description: string;
 
   /**
-   * Callback invoked upon updating a commit message summary.
+   * Updates the commit message summary.
    *
-   * @param event - event object
+   * @param summary - commit message summary
    */
-  onSummaryChange: (event: any) => void;
+  setSummary: (summary: string) => void;
 
   /**
-   * Callback invoked upon updating a commit message description.
+   * Updates the commit message description.
    *
-   * @param event - event object
+   * @param description - commit message description
    */
-  onDescriptionChange: (event: any) => void;
+  setDescription: (description: string) => void;
 
   /**
-   * Callback invoked upon clicking a commit message submit button or otherwise submitting the form.
+   * Callback to invoke in order to commit changes.
+   *
+   * @returns a promise which commits changes
    */
-  onCommitSubmit: () => Promise<void>;
+  onCommit: () => Promise<void>;
 }
 
 /**
@@ -116,7 +118,7 @@ export class CommitBox extends React.Component<ICommitBoxProps> {
             'Enter a commit message summary (a single line, preferably less than 50 characters)'
           )}
           value={this.props.summary}
-          onChange={this.props.onSummaryChange}
+          onChange={this._onSummaryChange}
           onKeyPress={this._onSummaryKeyPress}
         />
         <TextareaAutosize
@@ -125,7 +127,7 @@ export class CommitBox extends React.Component<ICommitBoxProps> {
           placeholder={this.props.trans.__('Description (optional)')}
           title={this.props.trans.__('Enter a commit message description')}
           value={this.props.description}
-          onChange={this.props.onDescriptionChange}
+          onChange={this._onDescriptionChange}
         />
         <input
           className={commitButtonClass}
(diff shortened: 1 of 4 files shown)


fcollonval:
> @navn-r I edited your initial comment so GitHub links correctly to the two addressed issues. GitHub parser is not able to capture multiple references after special words like _Fixes_.

tests/test-components/GitPanel.spec.tsx
```diff
@@ -78,10 +78,31 @@ describe('GitPanel', () => {
...
+    });
+
+    it('should set the default commit message summary to an empty string', () => {
+      const panel = new GitPanel(props);
+      expect(panel.state.summary).toEqual('');
+    });
+
+    it('should set the default commit message description to an empty string', () => {
+      const panel = new GitPanel(props);
+      expect(panel.state.summary).toEqual('');
```
fcollonval, line 94: I presume the intention is:

```suggestion
      expect(panel.state.description).toEqual('');
```

tests/test-components/CommitBox.spec.tsx
```diff
@@ -16,50 +16,28 @@ describe('CommitBox', () => {
 
   const trans = nullTranslator.load('jupyterlab-git');
 
+  const defaultProps: ICommitBoxProps = {
```
fcollonval, line 19: Great thanks for this refactoring

fcollonval (changes requested):
This looks great. Thanks @navn-r 

Some comments:
- If you try the binder demo on your PR, you will be able to test committing without your identity set. If you cancel the dialog asking for your identity, the commit summary and description will be erased. It would be nice to change that too. For that I suggest to change `GitPanel._hasIdentity` to not return a boolean but to raise an error instead of returning `false`.
> You probably will need to rework the existing try-catches.
- As `summary` and `description` have been moved in `GitPanel`, I think renaming them in that component `commitSummary` and `commitDescription` will clarify the code.
- [not mandatory] As those new states are not link directly to a view, we may want the `onSummaryChange` and `onDescriptionChange` to be more generic `(value: string) => void` instead of `(event: any) => void`

src/components/GitPanel.tsx
```diff
@@ -205,67 +217,36 @@ export class GitPanel extends React.Component<IGitPanelProps, IGitPanelState> {
...
-   * Commits all staged files.
-   *
-   * @param message - commit message
-   * @returns a promise which commits the files
-   */
-  commitStagedFiles = async (message: string): Promise<void> => {
-    if (!message) {
+    if (!msg) {
       return;
     }
```
navn-r, line 234: As you know, the CommitBox was responsible for the commit message, and one thing I came across was 

This check initially, never actually did anything, since `message` before my changes, will never be empty during this check.
What I mean by that is,
```ts
const summary = "";
const description = ""
const msg = summary + '\n\n' + description + '\n';
console.log(!msg);  // false   
```

This wasn't picked up from a user standpoint since the commit button disables/enables correctly. But yea just fyi I guess

src/components/GitPanel.tsx
```diff
@@ -513,32 +571,30 @@ export class GitPanel extends React.Component<IGitPanelProps, IGitPanelState> {
...
         this._previousRepoPath = path;
       } catch (error) {
+        if (error instanceof Git.GitResponseError) {
+          throw error;
+        }
+
         throw new Error(
           this.props.trans.__('Failed to set your identity. %1', error.message)
         );
       }
```
navn-r, line 596: I wasn't sure if checking for this specific Response Error was important or not. But I figured it was important since has that `traceback` property in the class definition which afaik would have gotten lost if I didn't check.

src/components/CommitBox.tsx
```diff
@@ -56,15 +43,33 @@ export interface ICommitBoxState {
    * Commit message description.
    */
   description: string;
+
+  /**
+   * Updates the commit message summary.
+   *
+   * @param summary - commit message summary
+   */
+  setSummary: (summary: string) => void;
```
fcollonval, line 52: Great thanks for using a more appropriate name

src/components/GitPanel.tsx
fcollonval: Thanks for the detailed explanation and good catch.

fcollonval (approved):
LGTM thanks @navn-r

Issue #695, opened 2020-07-13
git commit --amend


ianhi:
## Description
It would be handy to have the option equivalent to `git commit --amend` - This would need to paired with an option to force push. 

Perhaps this could go into the Git menu similar to vscode:
![image](https://user-images.githubusercontent.com/10111092/87264877-2fd3d280-c48f-11ea-84bb-6d722c5c3ba8.png)

fcollonval:
cc @sinakhalili

fcollonval:
Closed by #973

PR #973, opened 2021-07-09
Git commit amend

SinaKhalili:
Ref #695 

This will add a new option checkbox next to the commit button to amend the current staged commit to the previous one.

This will also add a `Git Push (Force)` option, so that the user can push the amended commit

git_add

In the future we might want to allow the user to change the previous commit message, but for now it just uses the `--no-edit` option.

I haven't written any tests or anything yet because I'd like to see if I'm on the right track so far 👌🏽


src/model.ts
```diff
@@ -487,17 +487,19 @@ export class GitExtension implements IGitExtension {
...
    * @param message - commit message
+   * @param amend - flag to indicate amending commit
    * @returns promise which resolves upon committing file changes
    *
    * @throws {Git.NotInRepository} If the current path is not a Git repository
    * @throws {Git.GitResponseError} If the server response is not ok
    * @throws {ServerConnection.NetworkError} If the request cannot be made
    */
-  async commit(message: string): Promise<void> {
+  async commit(message: string, amend = false): Promise<void> {
```
fcollonval, line 497: It would make sense to set `message` as optional: `message?: string`

fcollonval (changes requested):
Thanks @SinaKhalili 

I think the context menu UI is confusing because if triggered, it will commit all staged files not only the selected one.
I would suggest adding a checkbox next to the commit button:

[ Commit ] 🔳 amend

jupyterlab_git/handlers.py
```diff
@@ -473,7 +473,10 @@ async def post(self, path: str = ""):
         """
         data = self.get_json_body()
         commit_msg = data["commit_msg"]
-        body = await self.git.commit(commit_msg, self.url2localpath(path))
+        if data["amend"]:
+            body = await self.git.commit_ammend(self.url2localpath(path))
```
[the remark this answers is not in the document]
fcollonval: Thinking a second time about it, it will be better to pass an additional argument `amend=True/False`. So we can handle in the future the edition of the commit message.

SinaKhalili:
Hey @fcollonval I've updated this PR, let me know what you think 😄

jupyterlab_git/git.py
```diff
@@ -976,11 +983,15 @@ async def pull(self, path, auth=None, cancel_on_conflict=False):
...
-    async def push(self, remote, branch, path, auth=None, set_upstream=False):
+    async def push(
+        self, remote, branch, path, auth=None, set_upstream=False, force=False
+    ):
         """
         Execute `git push $UPSTREAM $BRANCH`. The choice of upstream and branch is up to the caller.
         """
         command = ["git", "push"]
+        if force:
+            command.append("--force")
```
fcollonval, line 994: We should be less aggressive by using `--force-with-lease`

```suggestion
            command.append("--force-with-lease")
```

See documentation: https://git-scm.com/docs/git-push

jupyterlab_git/handlers.py
```diff
@@ -473,7 +473,8 @@ async def post(self, path: str = ""):
         """
         data = self.get_json_body()
         commit_msg = data["commit_msg"]
-        body = await self.git.commit(commit_msg, self.url2localpath(path))
+        amend = data["amend"]
```
fcollonval, line 476: We could make this one optional to be more forgiven:

```suggestion
        amend = data.get("amend", False)
```

src/components/CommitBox.tsx
```diff
@@ -114,29 +127,52 @@ export class CommitBox extends React.Component<ICommitBoxProps> {
           className={commitSummaryClass}
           type="text"
           placeholder={summaryPlaceholder}
-          title={this.props.trans.__(
-            'Enter a commit message summary (a single line, preferably less than 50 characters)'
-          )}
+          title={
+            this.props.amend
+              ? this.props.trans.__(
+                  'You are amending this commit to the previous one, so the previous commit summary will be used'
```
fcollonval, line 133: Let's make it shorter:

```suggestion
                  'Amending the commit will re-use the previous commit summary'
```

src/components/CommitBox.tsx
```diff
@@ -114,29 +127,52 @@ export class CommitBox extends React.Component<ICommitBoxProps> {
...
         />
         <TextareaAutosize
           className={commitDescriptionClass}
           minRows={5}
           placeholder={this.props.trans.__('Description (optional)')}
-          title={this.props.trans.__('Enter a commit message description')}
+          title={
+            this.props.amend
+              ? this.props.trans.__(
+                  'You are amending this commit to the previous one, so the previous commit message will be used'
```
fcollonval, line 151:
```suggestion
                  'Amending the commit will re-use the previous commit description'
```

src/components/CommitBox.tsx
```diff
@@ -145,6 +181,9 @@ export class CommitBox extends React.Component<ICommitBoxProps> {
    * Whether a commit can be performed (files are staged and summary is not empty).
    */
   private _canCommit(): boolean {
+    if (this.props.amend) {
+      return true;
```
fcollonval, line 185: Should we protect against amending without changes (as we don't allow to edit the commit message)?

```suggestion
      return this.props.hasFiles;
```

src/commandsAndMenu.tsx
```diff
@@ -377,6 +378,41 @@ export function addCommands(
     }
   });
 
+  /** Add git push force command */
+  commands.addCommand(CommandIDs.gitForcePush, {
```
fcollonval, line 382: Let's reuse the existing `gitPush` command - I will explain how in a PR comment.

fcollonval (changes requested):
Hey @SinaKhalili 

This is going to be great. I added some suggestions and here are so more comments:
- Could add test on the python for the commit call with amend? You could take inspiration from https://github.com/jupyterlab/jupyterlab-git/blob/823e68e3c1effce8126f73429ce3d2c6eac9d708/jupyterlab_git/tests/test_handlers.py#L312
- Could you reuse the `CommandIDs.gitPush`?

For the latter, you can pass a JSON object as args to commands.

fcollonval:
> you can pass a JSON object as args to commands.

I forgot to add you an example

You can define the command with args:

```
  commands.addCommand(CommandIDs.gitPush, {
    label: args => args.force ? trans.__('Push to Remote (Force)') : trans.__('Push to Remote'),
    caption: args => trans.__('Push code to remote repository'),
    isEnabled: () => gitModel.pathRepository !== null,
    execute: async (args) => {
      const force = !!args.force;
      // Do something
    }
  });
```

You need to define the `args` object when you add the command to the menu:

```
menu.addItem({ command: 'command-id', args: {force: true} });
```

krassowski:
UI-wise,what do you think about making it like the "Merge" button on GitHub (rather than a checkbox)? Here is a quick adoption of the GitHub's button:

| collapsed | expanded |
|--------|-------|
| Screenshot from 2021-07-17 14-30-06  | Screenshot from 2021-07-17 14-31-59 |

SinaKhalili:
@fcollonval Thanks for the feedback, changes applied 👍🏽 
@krassowski That looks great! Much better than the checkbox lol. I'll try to get this functionality merged and then try and change that in another PR 😄

fcollonval (changes requested):
Thanks @SinaKhalili this looks great.

I open a new issue to track the UX suggestion of @krassowski 
Regarding this PR, one last thing is missing for me: `CommitBox` tests.

In `tests/test-components/CommitBox.spec.tsx` could you add tests (take inspiration of what is already there) to ensure the state of the summary and description element are consistent (dis/enabled) depending of the _amend_ state.

SinaKhalili:
Thanks @fcollonval I've added some test 😄

fcollonval (approved):
Thanks @SinaKhalili and congrats for your first PR :tada:

fcollonval:
@all-contributors please add @SinaKhalili for code

PR #980, opened 2021-07-20
Git history for a selected file (#864)

### LINKED ISSUE #864: Git history for a selected file

navn-r:
### Added history context menu button in file browser

 
 
- Emits a new signal in Git Model
- Navigates to Git Panel, and switches to History Tab
  
### Displays Single File History in History Tab

 
 

- Clicking the `Discard file history` button 'reverts' the history tab back to its usual repo history
- Each commit node is now hoverable, and on click will open the diff
- It will *not* have a hover effect and will disable its hover+click functionality if the file is binary

  

**TODO:**

- [x] Styling: the `FileItem` in the history panel should be styled
- [x] The diff file context menu item should be available for modified file
- [x] Add/Update/Fix tests

Fixes #864


style/icons/clock.svg
```diff
@@ -0,0 +1,14 @@
...
+  aria-hidden="true"
+  class="jp-icon3" 
+  role="img"
+  xmlns="http://www.w3.org/2000/svg"
+  width="16"
+  viewBox="0 0 512 512"
+  fill="none"
+>
+  <path
+    fill="currentColor"
```
fcollonval, line 11:
```suggestion
    class="jp-icon3" 
    fill="currentColor"
```

fcollonval (changes requested):
Thanks a lot @navn-r This is great!

I commented in the code as much as possible. One UI comment - reusing the `FileItem` component to display the file name at the history top is a good idea. It will probably require some styling to highlight it compare to the history list (a different background color maybe).

jupyterlab_git/git.py
```diff
@@ -432,16 +432,19 @@ async def status(self, path):
...
         """
+        is_single_file = follow_path != "."
         cmd = [
             "git",
             "log",
             "--pretty=format:%H%n%an%n%ar%n%s",
             ("-%d" % history_count),
         ]
+        if is_single_file:
+            cmd = cmd + ["--numstat", "-z", "--follow", "--", follow_path]
```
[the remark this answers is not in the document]
navn-r: -z is used for the `is_binary` check for single file history. I figured this approach would require less code changes, versus creating a separate endpoint for `is_binary`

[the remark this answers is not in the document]
navn-r: Wait my apologies lol, yea it can be removed. I thought you meant the `--numstat` option :laughing:

src/components/HistorySideBar.tsx
```diff
@@ -44,17 +50,110 @@ export interface IHistorySideBarProps {
...
+          console.error(`Failed to open diff view for ${filePath}.\n${err}`);
+        }
+      }
+    };
+
+  return (
+    <ol className={historySideBarStyle}>
+      {!!props.model.selectedHistoryFile && (
+        <FileItem
+          style={{ overflowX: 'hidden', flexGrow: 0, flexShrink: 0 }}
```
fcollonval, line 109: Please don't use style directly (this was required by the windowing library but it should be avoided as much as possible except for dealing with positioning). Prefer adding a new props `className` to pass a `className` to be appended in `FileItem` - you can look at `ActionButton` as example.

fcollonval (changes requested):
Thanks for quickly applying the suggestion.

Remaining points I see:
- Styling: the `FileItem` in the history panel should be styled
  - to stand out of the history commit list
  - to not be hidden when the user scrolls on the commits list (maybe making it sticky would work)
  - Not applying style through the style props (see comment in code)
- The diff file context menu item should be available for modified file (it is not currently). For such case it will be good to add a history item that allow to display uncommited changes (you can have a look at how VS Code is doing it in its timeline panel).
- Correct the tests and add new ones

navn-r:
**Progress Update** 

From this recent push... (styled and sticky selected file + Uncommitted changes commit node)

image

Besides adding/fixing tests, these are the only things left, as requested by @adammclaurin: 

- [ ] Handle history edge case for Ignored files 

From @fcollonval:
> It maybe interesting to update the current ignored files when the filebrowser path changes (hmm some polling will probably be needed too as files can be added, renamed or deleted). Maybe listening to fileChanged and pathChanged signals can be enough

- [ ] Select two arbitrary commits and view their diff (`Select for compare` & `Compare with selected` from VSCode)
- Pending discussion on how to handle this

fcollonval:
Thanks for the update.

> Select two arbitrary commits and view their diff (Select for compare & Compare with selected from VSCode)

I would left this one out for a follow-up PR. There brings in already a big change.

For how to do it (in the single file history view), using a context menu would be nice and/or a visual check box on the commit info top right - but this maybe complex because we need to enforce two nodes selection only and there is the notion of what are the reference and challenger.

fcollonval:
> I figured just adding a 'no history found' default to the side bar requires less effort yet delivers the same end-user value, compared to watching for the ignored files then trying to hide the history context button.

:+1:

We can open a new issue with enhancement proposal to deal with ignored files.

src/components/FileItem.tsx
```diff
@@ -182,7 +190,7 @@ export class FileItem extends React.PureComponent<IFileItemProps> {
         }
         onDoubleClick={this.props.onDoubleClick}
         style={this.props.style}
-        title={this.props.trans.__(`%1 ● ${status}`, this.props.file.to)}
+        title={this.props.trans.__(`%1 • ${status}`, this.props.file.to)}
```
fcollonval, line 193: What is the rationale to change that character?

src/components/HistorySideBar.tsx
```diff
@@ -102,11 +106,32 @@ export const HistorySideBar: React.FunctionComponent<IHistorySideBarProps> = (
...
     };
 
+  /**
+   * Commit info for 'Uncommitted Changes' history.
+   */
+  const uncommitted = React.useMemo<Git.ISingleCommitInfo>(() => {
+    return {
+      author: props.trans.__('You'),
+      // INDEX or WORKING special ref
+      commit: `${+(props.model.selectedHistoryFile?.status === 'staged')}`,
```
fcollonval, line 116: That is a nice trick but it makes the code less readable. You should always prefer readability (expect when performance is a prime concerned - but even then what seems like a more efficient written code may not be efficient for the processor)

```suggestion
      commit: `${props.model.selectedHistoryFile?.status === 'staged' ? Git.Diff.SpecialRef.INDEX : Git.Diff.SpecialRef.WORKING }`,
```

src/components/HistorySideBar.tsx
```diff
@@ -102,11 +106,32 @@ export const HistorySideBar: React.FunctionComponent<IHistorySideBarProps> = (
...
+      pre_commit: 'HEAD',
+      is_binary: !!props.commits[0]?.is_binary,
```
fcollonval, line 118: Same here, for clarity let use the nullish coalescing operator `??`

```suggestion
      is_binary: props.commits[0]?.is_binary ?? false,
```

src/components/HistorySideBar.tsx
```diff
@@ -102,11 +106,32 @@ export const HistorySideBar: React.FunctionComponent<IHistorySideBarProps> = (
       }
     };
 
+  /**
+   * Commit info for 'Uncommitted Changes' history.
+   */
+  const uncommitted = React.useMemo<Git.ISingleCommitInfo>(() => {
```
fcollonval, line 112: Nice usage of `useMemo`

fcollonval (changes requested):
Almost there @navn-r 

I have some suggestions to improve readability. You need also to rebase the PR and to fix the tests.

I opened the follow-up issues: #984 and #985

src/components/FileItem.tsx
navn-r: Nothing super important really,

I found that VSCode uses the smaller bullet

image

and I preferred how that looked :sweat_smile:, quite subjective lol

fcollonval (changes requested):
Thanks @navn-r 

Could you please add a Python unit test that check the git command executed and the parsing of the output?

The best starting point for that would be to duplicate `jupyterlab_git/tests/test_detailed_log.py` and apply it to the single file case.

fcollonval (approved):

fcollonval:
Thanks a lot @navn-r

src/components/FileItem.tsx
```diff
@@ -133,7 +137,7 @@ export class FileItem extends React.PureComponent<IFileItemProps> {
     super(props);
   }
   protected _getFileChangedLabel(change: keyof typeof STATUS_CODES): string {
-    return STATUS_CODES[change];
+    return STATUS_CODES[change] || 'Unmodified';
```
krassowski, line 140: Translation missing?

fcollonval: Thanks @krassowski

PR #988, opened 2021-07-29
Fix translation

fcollonval:
- ~~The translation strings are not all captured by introspection~~ => cover by https://github.com/jupyterlab/jupyterlab-translate/pull/6
- Update the domain name to the normalized version

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -76,7 +76,7 @@ async function activate(
   // Get a reference to the default file browser extension
   const filebrowser = factory.defaultBrowser;
   translator = translator || nullTranslator;
-  const trans = translator.load('jupyterlab-git');
+  const trans = translator.load('jupyterlab_git');
 
   // Attempt to load application settings
   try {
--- a/tests/commands.spec.tsx
+++ b/tests/commands.spec.tsx
@@ -37,7 +37,7 @@ describe('git-commands', () => {
       commands,
       shell: null as any
     };
-    const trans = nullTranslator.load('jupyterlab-git');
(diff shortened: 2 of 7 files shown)


krassowski:
We could "just" add `this.props.trans` in `jupyterlab-translate`: https://github.com/jupyterlab/jupyterlab-translate/blob/a4bc91e830200be403cc3bf1bd1c0282b2a52978/jupyterlab_translate/utils.py#L246-L291 (I think the combinations should be generated rather than hardocoded though).

fcollonval:
Thanks for jumping in @krassowski! 

I'm trying to check if it works. Then ja that hardcoded list is of expression is pretty annoying. And there is no way to convert all name idea that developer will come up with... . So I don't know what is the best solution. We could:

- Explain with the documentation what variable names are acceptable
- Allow an extension to define a configuration to be merged with the default one (although I really would like to avoid ending up with really complex scripts in `jupyterlab_translate`
- ... something else

krassowski:
I agree with your assessment. I think that maybe `this.props.trans` should be supported as JupyterLab supports using React widgets in extensions. Another potential addition would be `this.model.trans` as this might be a common pattern (not sure whether good or not). I would not add any more exceptions beyond this too (or probably just add `this.props.trans`), and I agree that making it customizable may not be worth it.

fcollonval:
Ok I just tested the renaming is working (the script extracted two times more strings than before :astonished: )

Ok then I'll push a PR to add `this.props.trans` and `props.trans` (to cover functional component too).

PR #990, opened 2021-07-30
URI -> URL

krassowski:
`git clone` accepts URLs, not any URIs see https://git-scm.com/docs/git-clone#URLS


fcollonval:
Unsure we need to change it as the same page is stating:

> For local repositories, also supported by Git natively, the following syntaxes may be used:
>    /path/to/repo.git/
>    file:///path/to/repo.git/

PR #994, opened 2021-08-03
Improve commit amend selection

### LINKED ISSUE #981: Improve UX for commit amend
UI-wise,what do you think about making it like the "Merge" button on GitHub (rather than a checkbox)? Here is a quick adoption of the GitHub's button:

| collapsed | expanded |
|--------|-------|
| Screenshot from 2021-07-17 14-30-06  | Screenshot from 2021-07-17 14-31-59 |

_Originally posted by @krassowski in https://github.com/jupyterlab/jupyterlab-git/issues/973#issuecomment-881899889_

fcollonval:
Fixes #981

image


fcollonval:
cc @krassowski @SinaKhalili what do you think?

krassowski:
I would use non-thin version of the caret. Should it be the down caret up up caret? Or maybe ⋮?

fcollonval:
> I would consider using non-thin version of the caret. Should it be the down caret up up caret? Or maybe ⋮?

Indeed a ⋮icon will be more appropriate as the pop up will never be displayed below.

fcollonval:
The icon has been updated to ⋮

PR #1001, opened 2021-08-04
Use PageConfig info if git repository is the root folder

### LINKED ISSUE #992: Use PageConfig info if git repository is the root folder
@fcollonval thanks for the fix! Branches work now as expected again!

With https://github.com/jupyterlab/jupyterlab-git/commit/96bef4502efdf6e7afea7bf068d77a9913bb4eb8 I see under `Current Repository` a `/` (see screenshot below).

image

With v0.30.1 I see the repo name (see screenshot below):
image

So just a quick question: Is this is the desired behavior for versions > `v0.30.1`?

_Originally posted by @bbo2adwuff in https://github.com/jupyterlab/jupyterlab-git/issues/989#issuecomment-891690104_

fcollonval:
Restore displaying the repository name if the repository is the Jupyter Server root folder using PageConfig. And use the information to build the local path if possible in the title.

image

Fixes #992

cc @bbo2adwuff is it addressing your question?

### DIFF
--- a/src/components/Toolbar.tsx
+++ b/src/components/Toolbar.tsx
@@ -1,4 +1,4 @@
-import { PathExt } from '@jupyterlab/coreutils';
+import { PageConfig, PathExt } from '@jupyterlab/coreutils';
 import {
   caretDownIcon,
   caretUpIcon,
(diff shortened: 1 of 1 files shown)


bbo2adwuff:
Now it looks like this:
image

Indeed, now it looks exactly like the screenshot that I posted in https://github.com/jupyterlab/jupyterlab-git/issues/989#issuecomment-891690104 for v0.30.1.

Thanks for the quick fixes :rocket:

PR #1019, opened 2021-08-29
Add Integration Tests

navn-r:
Adds `/ui-tests` using Playwright test-runner, currently with one spec for merge conflicts.
Removes `/tests/test-browser`

cc @fcollonval


.github/workflows/build.yml
```diff
@@ -59,7 +59,7 @@ jobs:
             ${{ runner.os }}-yarn-
 
       - name: Install dependencies
-        run: python -m pip install -U jupyter_packaging~=0.7.9 jupyterlab~=3.0 pip wheel
+        run: python -m pip install -U jupyter_packaging~=0.7.9 jupyterlab~=3.0 pip wheel build
```
fcollonval, line 62: Not used

```suggestion
        run: python -m pip install -U jupyter_packaging~=0.7.9 jupyterlab~=3.0 pip wheel
```

ui-tests/docker/prepare.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e
+
+yarn install
```
fcollonval, line 4: Ensure we do not update yarn dependencies

```suggestion
yarn install --frozen-lockfile
```

fcollonval (changes requested):
Thanks a lot @navn-r for pushing this.

Two suggestions and it will be good to go.

fcollonval (approved):
thanks for quickly updating this PR

Issue #998, opened 2021-08-03
Provide a pull-only mode that completely overwrites local files


orome:
## Is your feature request related to a problem? Please describe.
I would like to use `jupyterlab-git` in the context of a shared project (e.g., though Binder), where contributions from users are not supported, and where any pulls should completely overwrite (and prune) local files. I know there are limited existing solutions for similar scenarios, but they create a layer (or more) of abstraction above Git, that results in the loss of information (e.g. `jupyterlab-git` effectively provides notification of remote changes).

## Describe the solution you'd like
It would be great of `jupyterlab-git` provided settings that

1. prevented any pushing, and
2. where all pulls completely overwrote (and pruned) local files

while still providing all of the other file viewing, diffing, and other features it supports

## Describe alternatives you've considered
Other solutions like `nbgitpuller` are very limited.

meeseeksmachine:
This issue has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/what-is-installed-by-default-with-binder/10217/11

fcollonval:
Nice suggestion @orome 

If you want I can provide pointers to help you start a PR for this one.

PR #1020, opened 2021-08-29
Add 'Discard All and Pull' button

### LINKED ISSUE #860: Add 'Discard All and Pull' button
We're trying to enable an operating model where some members of the team are 'creators' of notebooks, and others are 'runners' of those notebooks. The latter group is fairly non-technical, e.g. not familiar with Python code or Git. The model to share notebooks is through Git for a few reasons (esp. compliance), but we're running into frictions because the 'runner' group easily gets confused and frustrated by the details of Git. This is often caused by the simple fact that running a notebook changes the notebook file (since the output gets saved in the file), which requires an explicit discard per file before pulling updates from the Git repo.

Perhaps an easier operating model here would be a simple 'Discard All and Pull' button to allow the 'runner' to have a one-click way to sync their local sandbox with whatever is on the repo (e.g. when the 'creator' pushes an update), without having to worry about what files changed, merging, etc.

What do you think?

Fwiw this is similar to the reason why I created #856 , and why I think #837 and #659 are such good ideas.

navn-r:
This changes introduces a `Pull from remote (Discard all)` option to the Git Menu

image

Closes #860 

---

This may lead into the enhancement of #998, specifically
> where all pulls completely overwrote (and pruned) local files

### DIFF
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -46,6 +46,7 @@ import {
   Level
 } from './tokens';
 import { GitCredentialsForm } from './widgets/CredentialsBox';
+import { discardAllChanges } from './widgets/discardAllChanges';
 import { GitCloneForm } from './widgets/GitCloneForm';
 
 interface IGitCloneArgs {
@@ -66,8 +67,7 @@ enum Operation {
   Clone = 'Clone',
   Pull = 'Pull',
   Push = 'Push',
-  ForcePush = 'ForcePush',
-  DiscardPull = 'discardPull'
+  ForcePush = 'ForcePush'
 }
 
 interface IFileDiffArgument {
@@ -392,22 +392,27 @@ export function addCommands(
   commands.addCommand(CommandIDs.gitPull, {
     label: args =>
       args.discard
-        ? trans.__('Pull from Remote (Discard all)')
+        ? trans.__('Pull from Remote (Force)')
         : trans.__('Pull from Remote'),
     caption: args =>
       args.discard
-        ? trans.__('Discard all and pull from remote repository')
+        ? trans.__(
+            'Discard all current changes and pull from remote repository'
+          )
         : trans.__('Pull latest code from remote repository'),
     isEnabled: () => gitModel.pathRepository !== null,
     execute: async args => {
-      logger.log({
-        level: Level.RUNNING,
-        message: trans.__('Pulling…')
-      });
       try {
+        if (args.discard) {
+          await discardAllChanges(gitModel, trans);
+        }
+        logger.log({
+          level: Level.RUNNING,
+          message: trans.__('Pulling…')
+        });
         const details = await Private.showGitOperationDialog(
           gitModel,
-          args.discard ? Operation.DiscardPull : Operation.Pull,
+          Operation.Pull,
           trans
         );
         logger.log({
@@ -1401,9 +1406,6 @@ namespace Private {
           break;
         case Operation.Pull:
           result = await model.pull(authentication);
-          break;
-        case Operation.DiscardPull:
-          result = await model.pull(authentication, true);
           break;
         case Operation.Push:
           result = await model.push(authentication);
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -391,11 +391,11 @@ export function addCommands(
   /** Add git pull command */
   commands.addCommand(CommandIDs.gitPull, {
     label: args =>
-      args.discard
+      args.force
         ? trans.__('Pull from Remote (Force)')
         : trans.__('Pull from Remote'),
     caption: args =>
-      args.discard
+      args.force
         ? trans.__(
             'Discard all current changes and pull from remote repository'
           )
@@ -403,7 +403,7 @@ export function addCommands(
     isEnabled: () => gitModel.pathRepository !== null,
     execute: async args => {
       try {
-        if (args.discard) {
+        if (args.force) {
           await discardAllChanges(gitModel, trans, args.fallback as boolean);
         }
         logger.log({
@@ -438,7 +438,7 @@ export function addCommands(
             )
         ) {
           await commands.execute(CommandIDs.gitPull, {
-            discard: true,
+            force: true,
             fallback: true
           });
         } else {
@@ -1197,7 +1197,7 @@ export function createGitMenu(
       menu.addItem({ command, args: { force: true } });
     }
     if (command === CommandIDs.gitPull) {
-      menu.addItem({ command, args: { discard: true } });
+      menu.addItem({ command, args: { force: true } });
     }
   });
 
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -443,8 +443,9 @@ export function addCommands(
           });
         } else {
           if ((error as any).cancelled) {
+            // Empty message to hide alert
             logger.log({
-              message: trans.__('Pull cancelled'),
+              message: '',
               level: Level.INFO
             });
           } else {
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -404,7 +404,7 @@ export function addCommands(
     execute: async args => {
       try {
         if (args.discard) {
-          await discardAllChanges(gitModel, trans);
+          await discardAllChanges(gitModel, trans, args.fallback as boolean);
         }
         logger.log({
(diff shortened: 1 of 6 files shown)


src/model.ts
```diff
@@ -738,14 +738,23 @@ export class GitExtension implements IGitExtension {
...
+  async pull(
+    auth?: Git.IAuth,
+    discard = false
+  ): Promise<Git.IResultWithMessage> {
     const path = await this._getPathRepository();
+
+    if (discard) {
+      await this.resetToCommit('HEAD');
+    }
+
```
navn-r, line 757: Right now there is no prompt to warn the user that it'll discard everything. 

I would like it to just call this same dialog found in FileList
https://github.com/jupyterlab/jupyterlab-git/blob/master/src/components/FileList.tsx#L209-L232

Not sure the best way to refactor this dialog so it can be called in either the commandsAndMenu, or the model.

Any ideas?

src/commandsAndMenu.tsx
```diff
@@ -383,18 +384,24 @@ export function addCommands(
 
   /** Add git pull command */
   commands.addCommand(CommandIDs.gitPull, {
-    label: trans.__('Pull from Remote'),
-    caption: trans.__('Pull latest code from remote repository'),
+    label: args =>
+      args.discard
+        ? trans.__('Pull from Remote (Discard all)')
```
fcollonval, line 389: I would use a wording mirroring the push case.

```suggestion
        ? trans.__('Pull from Remote (Force)')
```

src/commandsAndMenu.tsx
```diff
@@ -383,18 +384,24 @@ export function addCommands(
...
+        : trans.__('Pull from Remote'),
+    caption: args =>
+      args.discard
+        ? trans.__('Discard all and pull from remote repository')
```
fcollonval, line 393:
```suggestion
        ? trans.__('Discard all current changes and pull from remote repository')
```

src/model.ts
fcollonval: Hey

It should be called in the command (the model should stay agnostic to the UI).

You can extract the `showDialog` call from `FileList` to its own file (in the `widgets` folder as this is not a react component). The refactored code could look like:

```ts
  async function discardAllChanges(model: IGitExtension, trans: TranslationBundle): Promise<void> {

    const result = await showDialog({
      title: trans.__('Discard all changes'),
      body: trans.__(
        'Are you sure you want to permanently discard changes to all files? This action cannot be undone.'
      ),
      buttons: [
        Dialog.cancelButton({ label: trans.__('Cancel') }),
        Dialog.warnButton({ label: trans.__('Discard') })
      ]
    });
    if (result.button.accept) {
      try {
        await model.resetToCommit();
        return Promise.resolve(true);
      } catch (reason) {
        showErrorMessage(
          this.props.trans.__('Discard all changes failed.'),
          reason
        );
        return Promise.reject(reason);
      }
    } else {
      return Promise.reject('The user refused to discard the changes.');
    }
  };
```

fcollonval (changes requested):
Thanks for starting this @navn-r 

I left some comments

src/commandsAndMenu.tsx
```diff
@@ -383,15 +384,26 @@ export function addCommands(
...
-      });
+    execute: async args => {
       try {
+        if (args.discard) {
+          await discardAllChanges(gitModel, trans);
+        }
+        logger.log({
+          level: Level.RUNNING,
+          message: trans.__('Pulling...')
+        });
```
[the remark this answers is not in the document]
navn-r: Hey @fcollonval sorry for the delayed reply, school just started last week :smiley: 

The recent commit just uses a substring check for the fallback. Is that robust enough? Or do we need to extend `Error` and check that way?

src/tokens.ts
```diff
@@ -1012,9 +1012,7 @@ export enum CommandIDs {
   gitClone = 'git:clone',
   gitOpenGitignore = 'git:open-gitignore',
   gitPush = 'git:push',
-  gitForcePush = 'git:force-push',
   gitPull = 'git:pull',
-  gitDiscardPull = 'git:discard-pull',
```
fcollonval: thanks for cleaning this

src/commandsAndMenu.tsx
```diff
@@ -407,11 +419,24 @@ export function addCommands(
...
+        // Discard changes then retry pull
+        if (
+          (error as string)
+            .toLowerCase()
+            .includes(
+              'your local changes to the following files would be overwritten by merge'
+            )
+        ) {
+          await discardAllChanges(gitModel, trans, true);
+          await commands.execute(CommandIDs.gitPull, {} as any);
```
fcollonval, line 432: This is valid - the default value for `args` is `{}`

```suggestion
          await commands.execute(CommandIDs.gitPull);
```

src/commandsAndMenu.tsx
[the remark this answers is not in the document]
fcollonval: :+1: This sounds good enough.

src/commandsAndMenu.tsx
```diff
@@ -407,11 +419,24 @@ export function addCommands(
...
         );
-        logger.log({
-          message: trans.__('Failed to pull'),
-          level: Level.ERROR,
-          error
-        });
+
+        // Discard changes then retry pull
+        if (
+          (error as string)
```
fcollonval, line 425: Error may be a `Error` object that encapsulated the `message`:

```suggestion
        const errorMsg = typeof error === 'string' ? error : error.message;
        if (
          errorMsg
```

fcollonval (changes requested):
Thanks for pushing this (don't worry about the delay any help at any rate is welcomed).

I left some suggestions and it should be good to go.

Small note: for future contributions, I advice you to use a personal fork rather than the MLH one.

src/commandsAndMenu.tsx
```diff
@@ -407,11 +419,26 @@ export function addCommands(
...
+        // Discard changes then retry pull
+        if (
+          errorMsg
+            .toLowerCase()
+            .includes(
+              'your local changes to the following files would be overwritten by merge'
+            )
+        ) {
+          await discardAllChanges(gitModel, trans, true);
+          await commands.execute(CommandIDs.gitPull);
```
fcollonval, line 434: We need to rework this a bit to handle cancel action.

```suggestion
          await commands.execute(CommandIDs.gitPull, {discard: true, fallback: true});
```

src/commandsAndMenu.tsx
```diff
@@ -383,15 +384,26 @@ export function addCommands(
...
     isEnabled: () => gitModel.pathRepository !== null,
-    execute: async () => {
-      logger.log({
-        level: Level.RUNNING,
-        message: trans.__('Pulling...')
-      });
+    execute: async args => {
       try {
+        if (args.discard) {
+          await discardAllChanges(gitModel, trans);
```
fcollonval, line 401:
```suggestion
          await discardAllChanges(gitModel, trans, args.fallback);
```

src/widgets/discardAllChanges.ts
```diff
@@ -0,0 +1,40 @@
...
+export async function discardAllChanges(
+  model: IGitExtension,
+  trans: TranslationBundle,
+  isFallback?: boolean
+): Promise<void> {
+  const result = await showDialog({
+    title: trans.__('Discard all changes'),
+    body: isFallback
+      ? trans.__(
+          'Do you want to permanently discard changes to all files and pull? This action cannot be undone.'
```
fcollonval, line 19: I rephrase a bit the message

```suggestion
          'Your current changes forbid pulling the latest changes. Do you want to permanently discard those changes? This action cannot be undone.'
```

src/commandsAndMenu.tsx
```diff
@@ -407,11 +419,26 @@ export function addCommands(
...
+          errorMsg
+            .toLowerCase()
+            .includes(
+              'your local changes to the following files would be overwritten by merge'
+            )
+        ) {
+          await discardAllChanges(gitModel, trans, true);
+          await commands.execute(CommandIDs.gitPull);
+        } else {
+          logger.log({
```
fcollonval, line 436: We need to discard the alert widget if `errorMsg` === `'The user refused to discard all changes'` instead of displaying a red alert. You should be able to do that using `logger.log({})`.

fcollonval (changes requested):
Hey @navn-r I tested the latest version, we need to handle better the cases where the user cancel the discard dialog (see the review comments).

PR #1031, opened 2021-09-05
Improve merge widgets

fcollonval:
Follow-up of merge widgets feature

- Correct position Current / Result / Incoming to be coherent for plain text files and notebooks
- Correct styling for raw plain text
  - Position revert chunk button correctly
  - Display scroll lock button
- [ ] Orange text on blue background for selected conflicted file is hard to read
- [ ] Conflict notebook view in dark mode is hard to read
- [ ] Hidden unchanged cells on conflicted notebook is not working
- [ ] Move _Clear all cell outputs_ and _Clear conflicted cell outputs_ checkboxes in the toolbar
- [ ] nbdime working is local / base / remote => text will be expressed using that wording (like _cell added remotely_ or _cell delete locally_). This breaks consistency with a hard to understand wording.
- [ ] It seems that nbdime merge tool allow to move the cell (hence the anchor in the cell header). What do to with that... This interacting in an unexpected way with drag-and-drop animation of tabs in JupyterLab main area.
- [ ] ORIG_HEAD is not base content (it is actuall HEAD in case of merge). The code needs to be reworked to get the real base (requires a round trip to backend using `git ls-files -u` to get object ref?)
- Dealing with additional and deleted files is done directly by git.

### DIFF
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -619,8 +619,8 @@ export function addCommands(
         const diffContext: Git.Diff.IContext =
           status === 'unmerged'
             ? {
-                currentRef: 'HEAD',
-                previousRef: 'MERGE_HEAD',
+                currentRef: 'MERGE_HEAD',
+                previousRef: 'HEAD',
                 baseRef: 'ORIG_HEAD'
               }
             : context ?? {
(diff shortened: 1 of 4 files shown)


fcollonval:
@navn-r would you mind reviewing this one?

navn-r:
Hey @fcollonval, I can take a look and review it. I can't guarantee I'd get to it very soon, so it could be good to add another reviewer if urgent.

Issue #1035, opened 2021-09-14
Tooltip not show on commit button when it is disabled


andrewhli:
Hi, 

Here are my steps so far:

1. Created private GitHub SSH key in ~/.ssh (e.g. ~/.ssh/id_x)
2. Cloned remote GitHub repo via Jupyterlab Git UI (sidebar), using the SSH URL (e.g. git@github.com:user/repo.git)
3. Edited the readme in the repo via Jupyterlab
4. Tried to commit, but the option is greyed out. There are console warnings but no errors.
5. Updated - according to krassowski: files need to be added to "Staged" to enable commits. "Maybe there should be a tooltip showing up when hovering over the greyed out "Commit" button saying "Add at least one file to the staging area to commit"?"

![Screen Shot 2021-09-14 at 12 22 46 PM](https://user-images.githubusercontent.com/5068143/133304709-b9035460-2571-4c76-be85-01756374ed7e.png)

Would appreciate any help on this. Thanks!

krassowski:
Welcome! You need to add a file to the staging area so that the extension knows which is supposed to be committed. You can do this by hovering over the entries on the "Changed" or "Untracked" list (depending on what you would like to add) which should make a plus (+) icon appear - click on it to add a file and you will be able to commit. Alternatively, use context menu in the file browser  (Git submenu).

andrewhli:
Thanks! Adding the files to "Staged" worked perfectly. Any interest to adding this step to the jupyterlab-git Readme? I've been so used to using GitHub desktop, where it only shows the changes and skips over the "Staged" step (not visible to users like me), so I didn't know this was a required step in jupyterlab-git

krassowski:
You may like the `simpleStaging` option (see https://github.com/jupyterlab/jupyterlab-git#ui-settings). This step is not surprising to most Git users, but I see how the GUI is not helping currently.

Maybe there should be a tooltip showing up when hovering over the greyed out "Commit" button saying "Add at least one file to the staging area to commit"?

andrewhli:
The tooltip makes sense. I'll change this into a feature request and rename the title

fcollonval:
Just a note when tackling this: 

- An appropriated title is already set to the button
- It is not displayed due to the style `pointer-events: none;` set by material-ui:

![image](https://user-images.githubusercontent.com/8435071/133386196-4557457d-6fa3-44ae-a929-b2d466e31ac9.png)

[...]

PR #1036, opened 2021-09-15
Fix pointer on CommitBox not displaying

andrewhli:
PR request based on @fcollonval's suggestion from this thread: https://github.com/jupyterlab/jupyterlab-git/issues/1035#issuecomment-919755588

### DIFF
--- a/src/style/CommitBox.ts
+++ b/src/style/CommitBox.ts
@@ -109,5 +109,6 @@ export const activeStyle = style({
 export const disabledStyle = style({
   cursor: 'not-allowed !important',
   color: 'var(--jp-ui-font-color2) !important',
-  backgroundColor: 'var(--jp-layout-color3) !important'
+  backgroundColor: 'var(--jp-layout-color3) !important',
+  pointerEvents: 'auto !important'
 });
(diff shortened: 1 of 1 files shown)


andrewhli:
@fcollonval - Thanks for the tip. Could you please check the syntax? 
`Overload 1 of 2, '(...objects: NestedCSSProperties[]): string', gave the following error.
    Type '"auto !important"' is not assignable to type '"-moz-initial" | "inherit" | "initial" | "revert" | "unset" | "auto" | "none" | "visible" | "all" | "fill" | "stroke" | "painted" | "visibleFill" | "visiblePainted" | "visibleStroke" | PointerEvents[]'.`

src/style/CommitBox.ts
```diff
@@ -109,5 +109,6 @@ export const activeStyle = style({
 export const disabledStyle = style({
   cursor: 'not-allowed !important',
   color: 'var(--jp-ui-font-color2) !important',
-  backgroundColor: 'var(--jp-layout-color3) !important'
+  backgroundColor: 'var(--jp-layout-color3) !important',
+  pointerEvents: 'auto !important'
```
fcollonval, line 113: This is a limitation of the type. It seems it knows about `auto` but not with the flag `!important`. so let's trick ts on this one with a comment:

```suggestion
  // TypeScript does not know about the value with `!important` flag
  pointerEvents: 'auto !important' as any
```

fcollonval (approved):
Thanks @andrewhli

fcollonval:
@meeseeksdev please backport to 0.32.x

PR #1039, opened 2021-09-17
Fix Git history of relocated files

### LINKED ISSUE #1038: Unable to view full Git History of relocated files
## Description

Unable to view full Git History of files that have been relocated

## Reproduce
```
## in demo root repo

# add to file
echo "this is a test" > test.txt
git add test.txt
git commit -m  'add test.txt to root'

# update file 
echo "this is another test" >> test.txt
git add test.txt
git commit -m 'append more to test.txt'

# move file
mkdir -p another_folder
mv test.txt another_folder/test.txt
git add test.txt
git add another_folder/test.txt
git commit -m 'move test.txt to another_folder/test.txt'

git push
```

After the above setup, go into the UI, show history on `another_folder/test.txt` and the first 2 commits will not show the text that has been added and the text that has been appended. Only the last commit will show the diff consisting
```
this is a test
this is another test
```

## Expected behavior

The expected behavior would be the Git History consists of text changes from the first commit and second commit and a mv file in the third commit. 

## Context

```
# cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 10 (buster)"

# pip list | grep jupyterlab-git
jupyterlab-git         0.32.3

# jupyter labextension list
JupyterLab v3.0.17
  @jupyterlab/git v0.32.3 enabled OK (python, jupyterlab-git)

# jupyter server extension list
  jupyterlab_git enabled
  - Validating jupyterlab_git...
     jupyterlab_git 0.32.3 OK

# git --version
git version 2.20.1
```

navn-r:
Fixes #1038

- [x] *If* the rename commit has two different filenames. Want to display the diff like this for renames: 
image


jupyterlab_git/git.py
```diff
@@ -485,13 +486,22 @@ async def log(self, path, history_count=10, follow_path=None):
...
+            parsed_lines = []
+            for line in line_array:
+                if "\0" in line:
+                    split_line = line.replace("\0", "\t").split("\t")
+                    parsed_lines.append("\t".join(split_line[:-1]))
+                    if split_line[-1]:
+                        parsed_lines.append(split_line[-1])
+                else:
+                    parsed_lines.append(line)
+            line_array = parsed_lines
```
navn-r, line 501: Initially I just wanted to use strip_and_split with every separator being `\0`, but this was causing trouble since the number of lines is 4 if its a regular single file commit, but 5 if its a move/relocation one. Doing it this way the line_array is in the form

```python
[
  "Commit",
  "Author",
  "Date"
  "Insertions\tDeletions\tOld_file_path\tNew_file_path",
  "Pre_commit",
  ...
]
```
so now at least the `PREVIOUS_COMMIT_OFFSET` will be constant, like it was before.

Hoping to make this block smaller through list comprehension, but atm I think its fine.

jupyterlab_git/git.py
```diff
@@ -485,13 +486,22 @@ async def log(self, path, history_count=10, follow_path=None):
...
+        if is_single_file:
+            parsed_lines = []
+            for line in line_array:
+                if "\0" in line:
+                    split_line = line.replace("\0", "\t").split("\t")
+                    parsed_lines.append("\t".join(split_line[:-1]))
+                    if split_line[-1]:
+                        parsed_lines.append(split_line[-1])
+                else:
+                    parsed_lines.append(line)
```
fcollonval, line 500: I think we can simplify this. The only case where `split_line[-1]` is empty is the last entry otherwise it will be the previous commit. So

```suggestion
            for line in line_array[:-1]:
                parsed_lines.extend(line.split("\0\0", maxsplit=1))
            parsed_lines.append(line_array[-1].strip("\0"))
```

jupyterlab_git/git.py
```diff
@@ -502,12 +512,14 @@ async def log(self, path, history_count=10, follow_path=None):
 
             if is_single_file:
                 commit["is_binary"] = line_array[i + 4].startswith("-\t-\t")
+                # [insertions, deletions, original_file_path?, current_file_path]
+                commit["file_path"] = line_array[i + 4].split()[-1]
```
fcollonval, line 516: Here you will need to deal with the rename case (aka is there some null character in `line_array[i + 4]`)

fcollonval (changes requested):
Thanks for starting this @navn-r

fcollonval (approved):
Thanks a lot @navn-r 
This looks great.

I have an optional suggestion: it could be nice to update the `label`s passed to the diff widget to display the file names and only the short hash (7 characters). Something like `<filepath> (<short hash>)`.  
This will display clearly the rename.

fcollonval:
Awesome thanks @navn-r

fcollonval (approved):

PR #1075, opened 2022-02-11
Warn the user of unsaved changes when working with staged files

### LINKED ISSUE #853: Better handle case of unsaved changes
I keep running into the same unexpected behavior where I run a notebook to test my final changes, then go to the git extension to commit & push the changes, but then find afterwards that some of the changes didn't get pushed because the file wasn't saved. It'd be nice if the git extension could catch this and either auto-save the file or prompt the user to save.

quachtridat:
## Summary

Show a notice box warning the user to pay attention to unsaved changes as he/she/they are working on files that are staged. In other words, this PR makes JupyterLab-Git shows a warning box in the commit box section when it detects unsaved changes on files that have been staged. This will potentially fix #853.

## Details

**Note: We're only checking dirty state of staged files (not other files even if they belong to the repo) and the warning box only shows up when the user is working with unsaved staged files belonging to a repo recognized by JupyterLab-Git.**

In `src/model.ts`, these are added:
- `_hasDirtyStagedFiles: boolean` indicating whether there are dirty files that are also staged files.
- `_dirtyStagedFilesStatusChanged: Signal<IGitExtension, boolean>`  that emits a signal when `_hasDirtyStageFiles` changes, e.g., was `false` before but now `true` (e.g., due to detection of an unsaved file and that file is also in the staged list)
- `async refreshDirtyStagedStatus()` taking care of the above by basically checking with `docmanager` to see the dirty status of each file in the staged list. This is invoked within `GitExtension.refreshStatus()`.

`src/tokens.ts` is also modifed accordingly to adopt the changes above.

In `src/components/GitPanel.tsx`, these are done:
- Add `hasDirtyStagedFiles: boolean` to `IGitPanelState`
- `componentDidMount()` also connects the `_dirtyStagedFilesStatusChanged` with a callback that syncs `((GitPanel)this).state.hasDirtyStagedFiles`

`src/components/WarningBox.tsx` is a new component that is used for displaying warnings. Its styles are defined in `src/style/CommitBox.ts`. This component is used in `src/components/CommitBox.tsx` to show the warning for dirty state of staged files.

## Demos

image

## Notes

- This PR likely needs more tests.
- `refreshStatus()` is invoked periodically, therefore, as the user makes changes, or discards/saves file changes, the warning box may not show up/disappear immediately.

## Linked issues
#853

### DIFF
--- a/src/components/CommitBox.tsx
+++ b/src/components/CommitBox.tsx
@@ -192,25 +192,21 @@ export class CommitBox extends React.Component<
       'Summary (%1 to commit)',
       shortcutHint
     );
-    const dirtyStagedFilesWarningTitle = this.props.trans.__(
-      'Warning'
-    )
+    const dirtyStagedFilesWarningTitle = this.props.trans.__('Warning');
     const dirtyStagedFilesWarningContent = this.props.trans.__(
-      'Looks like you still have unsaved staged files. ' +
-      'Remember to save and stage all needed changes before committing!'
-    )
+      'Looks like you still have unsaved staged files. Remember to save and stage all needed changes before committing!'
+    );
     return (
       <div className={classes(commitFormClass, 'jp-git-CommitBox')}>
-        { this.props.warnDirtyStagedFiles ? (
+        {this.props.warnDirtyStagedFiles ? (
           <WarningBox
             headerIcon={<WarningRoundedIcon />}
             title={dirtyStagedFilesWarningTitle}
             content={dirtyStagedFilesWarningContent}
           />
         ) : (
-          <React.Fragment>
-          </React.Fragment>
-        ) }
+          <React.Fragment></React.Fragment>
+        )}
         <CommitMessage
           trans={this.props.trans}
           summary={this.props.summary}
--- a/src/components/CommitBox.tsx
+++ b/src/components/CommitBox.tsx
@@ -28,6 +28,8 @@ import {
 } from '../style/NewBranchDialog';
 import { CommandIDs } from '../tokens';
 import { CommitMessage } from './CommitMessage';
+import { WarningBox } from './WarningBox';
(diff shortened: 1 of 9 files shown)


src/components/CommitBox.tsx
```diff
@@ -185,8 +192,21 @@ export class CommitBox extends React.Component<
       'Summary (%1 to commit)',
       shortcutHint
     );
+    const dirtyStagedFilesWarningTitle = this.props.trans.__('Warning');
+    const dirtyStagedFilesWarningContent = this.props.trans.__(
```
fcollonval, line 196: Awesome you already pick up the way of supporting translation!

tests/test-components/CommitBox.spec.tsx
```diff
@@ -135,5 +137,14 @@ describe('CommitBox', () => {
       const prop = node.prop('disabled');
       expect(prop).toEqual(false);
     });
+
+    it('should render a warning box when there are dirty staged files', () => {
```
fcollonval, line 141: And you added a test :1st_place_medal:

src/model.ts
```diff
@@ -1041,6 +1060,38 @@ export class GitExtension implements IGitExtension {
...
+  /**
+   * Determines whether there are unsaved changes on staged files,
+   * e.g., the user has made changes to a file that has been staged,
+   * but has not saved them.
+   * Emits a signal indicating if there are unsaved changes.
+   * @returns promise that resolves upon refreshing the dirty status of staged files
+   */
+  async refreshDirtyStagedStatus(): Promise<void> {
+    // staged files
+    const fileNames = (await this._changedFiles('INDEX', 'HEAD')).files;
```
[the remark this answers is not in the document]
quachtridat: Thank you! This is what I was looking for.

fcollonval (approved):
Thanks @quachtridat

Issue #1070, opened 2022-01-25
Ignore extensions label should only show each extension once


krassowski:
The label for "ignore extensions" could be improved: currently it repeats extension if multiple files of same type are selected:

![Screenshot from 2022-01-25 15-59-41](https://user-images.githubusercontent.com/5832902/151012567-4f99e3bb-72c2-4459-a6c0-6ffdf209d741.png)

Ideally it would only show each unique extension once to avoid blowing up the size of the context menu.

## Additional context

- Python package version: 0.34.1
- Extension version: `@jupyterlab/git v0.34.1 enabled OK (python, jupyterlab-git)`

krassowski:
@meseeksdev label "Good first issue"

krassowski:
@meeseeksdev label "Good first issue"

fcollonval:
Thanks for reporting @krassowski

bsande6:
@krassowski Hey, Could I work on this issue?

krassowski:
Yes, please do open a pull request and feel free to ask questions :)

fcollonval:
Thanks @bsande6 to give you some pointers. The extensions list is built there:

https://github.com/jupyterlab/jupyterlab-git/blob/2f74ac486a6da505a793d79d9fc9527dc47d3155/src/commandsAndMenu.tsx#L1040-L1042

A possibility would be to apply a `.reduce` action to keep unique extensions.

bsande6:
@fcollonval Great, I will take a look.

bsande6:
@krassowski @fcollonval I have started working on the developer setup and am trying to install jupyter lab in a conda environment. When I run the command 
### ``jupyter labextension develop . --overwrite`` ###
I receive the following error ,

### The following error occurred while trying to add or remove files in the
    installation directory:

        [Errno 13] Permission denied: '/usr/lib/python3.8/site-packages'

    The installation directory you specified (via --install-dir, --prefix, or
    the distutils default setting) was:

        /usr/lib/python3.8/site-packages/

    This directory does not currently exist.  Please create it and try again, or
    choose a different installation directory (using the -d or --install-dir
    option).
###
I believe his is occurring because the path to python should go to my conda environment instead of /usr/lib/python3.8/site-packages/ . Do you know how I can solve this problem?

Additionally if there is a separate space to ask questions such as this can you point me to it. Thank you.

fcollonval:
@bsande6 did you activate your environment before running the command?
[...]

PR #1080, opened 2022-02-18
[Fix] Ignore extensions label should only show each extension once #1070

bsande6:
@krassowski Added a filter method to remove duplicates in the array.


fcollonval:
Thanks @bsande6 I push a small patch to fix the linter. Waiting for the CI but this looks good.

fcollonval (approved):

PR #1090, opened 2022-03-15
Refresh open file on pull

### LINKED ISSUE #1086: Automatic reload from disc after git pull
## Description

Automatic reload from disc after git pull in JupyterLab doen't work

## Reproduce

When pulling changes from git in JupyterLab, already opened files don't get updated. This has lead to situations, where developers pulled changes from git with some files already opened and those files didn't get updated with the changes from git. So when developers further edited the files without the changes from git, a warning appears that the files has changed on disk since the last time it was opened or saved with the options to revert the changes made since the git pull or to overwrite the file on disk.
Both of those options are not ideal, because changes in the code get lost anyway, either the ones from git oder the ones made by the developer in the meantime.
A reload from disk from disk immediately after the git pull prevents this problem.

## Expected behavior

Is it possible to automatically reload the files from disk after a git pull so that already opened files in the editor get updated as well? As far as I know, open files already get reloaded from disk for example when switching between branches.

## Context

```
  conda create -n jlab2022 python=3.8  -y
  conda activate jlab2022
  pip install --upgrade jupyterlab jupyterlab-git
```

Got JupyterLab 3.3.0 with jupyterlab-git 0.34.2

1.     Git clone someProject, or run push&pull in an existing clone
2. Open JupyterLab (I used Chrome)
3. Open some existing file mycode.py, e.g. double-click on the file in the jlab file browser. Keep the file open.
4. Modify and commit the file mycode.py outside of the local clone, e.g. edit the file directly in the browser at github.ibm.com
5. JupyterLab detects that the file has changed in the repo
6. Pull from the repo using git extension in jlab
7. Double-click on mycode.py in the jlab file browser again. This brings the editor window of mycode.py to the front. A user might assume that it has the new pulled content. But it still has the old content as the editor was left open before.
8. Make lots of changes in mycode.py, then save. Jlab warns that there are two different changed versions of the file.

@ajbozarth

ajbozarth:
Currently doing a pull does not update the contents of open files
like switching branches does.

This runs the same function to update files as switching branches but instead of diffing the branches it diffs the previous top commit to the new HEAD. 

Fixes #1086

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -840,6 +840,11 @@ export class GitExtension implements IGitExtension {
         );
       }
     );
+    const changes = await this._changedFiles(
+      this._currentBranch.top_commit,
+      'HEAD'
+    );
+    changes?.files?.forEach(file => this._revertFile(file));
     this.refreshBranch(); // Will emit headChanged if required
     return data;
   }
--- a/src/model.ts
+++ b/src/model.ts
@@ -824,7 +824,8 @@ export class GitExtension implements IGitExtension {
    */
   async pull(auth?: Git.IAuth): Promise<Git.IResultWithMessage> {
     const path = await this._getPathRepository();
-    const data = this._taskHandler.execute<Git.IResultWithMessage>(
+    const previousHead = this._currentBranch?.top_commit;
+    const data = await this._taskHandler.execute<Git.IResultWithMessage>(
       'git:pull',
       async () => {
         return await requestAPI<Git.IResultWithMessage>(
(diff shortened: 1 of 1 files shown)


fcollonval (changes requested):
Thanks @ajbozarth 

There is a bug in the existing method. And the call order needs to be set up a bit differently to ensure we get the correct value.

1. Store the top commit hash as we don't control its update - it may happen after the pull and before the changed files are requested.

2. We should wait for the task

```
const data = await this._taskHandler.execute
```

3. Get the changed files and revert

4. await for `this.refreshBranch()`

fcollonval:
Looking at the test failure, we need to check if `_currentBranch` is null or not :wink:

ajbozarth:
@fcollonval I made the updates you suggested. I original had the top commit stored first, but I stream-lined my changes when I found that the `refreshBranch` function call is what updated it. I do believe storing it first is still the safer plan though, even with adding the awaits

fcollonval (approved):
Thanks a lot @ajbozarth

fcollonval:
Release is in progress @ajbozarth

Issue #659, opened 2020-05-29
Cache git credentials via the jupyterlab-git extension


rdelubi:
As a feature request: 

Can jupyterlab-git extension provide a UI to cache git credentials so that user doesn't have to enter the credentials every time he/she needs to interact with remote repo? e.g. wrap the functionality of "git config credential.helper 'cache --timeout=xxx'" in the extension UI

fcollonval:
Tasks:

1. First add a settings in the extension to determine if the user wants to cache its credentials:
   1. Add a that settings in schema/plugin.json
   2. Maybe add a other one for the timeout
2. Pass that settings with the user authentication to the backend
3. If the settings needs to be cached
   1. Check if there is already a credentials helper
   2. if not and setting is true add the _cache helper_
   
That is

PR #1091, opened 2022-03-17
Frontend Cache Credentials for Issue #659

ZeshanFayyaz:
Some frontend exploration on #659

Added new interface `IAuthWithCacheCredentials` that extends `IAuth` that includes a new boolean that is implemented as a checkbox in `CredentialsBox.tsx.` and is used in `commandsAndMenu.tsx` as an additional argument.

When we push and clone we are expecting `authentication.cacheCredentials` (default False).

### DIFF
--- a/src/widgets/CredentialsBox.tsx
+++ b/src/widgets/CredentialsBox.tsx
@@ -24,7 +24,7 @@ export class GitCredentialsForm
     const node = document.createElement('div');
     const label = document.createElement('label');
     const checkBoxLabel = document.createElement('label');
-    const checkBox = document.createElement('input');
+    this._checkboxCacheCredentials = document.createElement('input');
     this._user = document.createElement('input');
     this._password = document.createElement('input');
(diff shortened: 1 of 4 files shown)


src/widgets/CredentialsBox.tsx
```diff
@@ -50,13 +61,17 @@ export class GitCredentialsForm
...
    * Returns the input value.
    */
-  getValue(): Git.IAuth {
+  getValue(): Git.IAuthWithCacheCredentials {
     return {
       username: this._user.value,
-      password: this._password.value
+      password: this._password.value,
+      cacheCredentials: this._checkboxCacheCredentials.checked
     };
```
fcollonval, line 69: I would extend `Git.IAuth` with a new optional parameter `cacheCredentials`. This will allow you to simply pass-by those information through the model and then to the server. One reason for that is the strong bound between that parameter and the credentials; passing it as a separate argument as done currently would be interesting if it had an effect even if the authentication was not defined.

PR #1099, opened 2022-03-23
Add the ability to save `Git` login temporarily

### LINKED ISSUE #659: Cache git credentials via the jupyterlab-git extension

quachtridat:
## Summary

This PR fixes #659 by implementing the ability that lets JupyterLab-Git users *cache* their login credentials. By default, the information is saved temporarily for 1 hour, and this can be configured (the unit of time is *seconds*). This feature is readily available for any one of the following Git operations:
- Clone: tick the checkbox when prompted for Git credentials.
- Push: tick the checkbox when prompted for Git credentials.
- Pull: tick the checkbox when prompted for Git credentials.
- Fetch: when credentials/login information is required, a small notification dot will appear on top of the Git icon in the status bar. Users can click it for the prompt to show up, and within that prompt pop-up, tick the checkbox if they wish to have their credentials cached.

Within the cache period, any of the aforementioned Git operations should not ask for the user's credentials again and should be able to work using the cached credentials.

### Frontend

- The Git credential prompt dialog is extended with a checkbox indicating whether the credentials should be cached or not.
- `Git.IAuth` is extended with an optional `cache_credentials: boolean`.
- The fetch poll will not send requests when Git credentials are required, instead it waits for the credentials to be provided, and gets unblocked when Git credentials are provided and the authentication is successful. A small colored dot appears on top of the Git extension icon in the status bar when this happens.

### Backend

- Endpoints `push`, `pull`, `fetch`, `clone` each accepts a `POST` request that includes `auth` and `auth` includes `cache_credentials` (optional/nullable/`None`-able). These are used to indicate what the user provided from the frontend.
- `git config` is used to make a decision here. If the user wishes to have their credentials temporarily saved, and in `git config`, `credential.helper` is not set, then `credential.helper` will be set to whatever that is configured in the server settings. Otherwise, no action is taken. **After that, in either case, if `credential.helper` is `cache` (i.e., Git's built-in credential caching mechanism), JupyterLab will attempt to spawn a daemon process that takes care of the caching work.**  In a brief amount of time after the daemon has been summoned, credentials given to Git will be picked up by that daemon.
- When the caching mechanism kicks in, a Git credential cache daemon is created in a separated process, and it will go away when the cache period expires. **Right now, the cache mechanism can only work on systems with support for Unix sockets, which means this feature will not work with standard Windows installations**.

## References

https://git-scm.com/docs/git-credential-cache: A `--timeout` argument can be added to specify how long (in seconds) credentials should be cached.

https://git-scm.com/docs/git-credential-cache--daemon

## Demos

image
*An orange dot appears on top of the Git icon in the status bar, indicating that it needs Git credentials.*

image
*Clicking the Git icon in the status bar should make this prompt pop-up show up. This also applies for doing clones, pushes and pulls.*

image
*There should be no colored dot while the cached information is alive and is valid.*

image
*The dot will show up again when the cached information does not exist (anymore) or is no longer valid. Also in this case, doing clones, pushes or pulls would require the user to input their Git credentials again.*

## Notes

This PR might need more tests.

## Linked issues
#659

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -261,7 +261,7 @@ class Git:
 
         return response
 
-    async def clone(self, path, repo_url, auth=None):
+    async def clone(self, path, repo_url, auth=None, cache_credentials=None):
         """
         Execute `git clone`.
         When no auth is provided, disables prompts for the password to avoid the terminal hanging.
@@ -273,6 +273,8 @@ class Git:
         """
         env = os.environ.copy()
         if auth:
+            if cache_credentials:
+                await self.ensure_credential_helper(path)
             env["GIT_TERMINAL_PROMPT"] = "1"
             code, output, error = await execute(
                 ["git", "clone", unquote(repo_url), "-q"],
@@ -296,7 +298,7 @@ class Git:
 
         return response
 
-    async def fetch(self, path):
+    async def fetch(self, path, auth=None, cache_credentials=False):
         """
         Execute git fetch command
         """
@@ -308,8 +310,21 @@ class Git:
             "--all",
             "--prune",
         ]  # Run prune by default to help beginners
-
-        code, _, fetch_error = await execute(cmd, cwd=cwd)
+        env = os.environ.copy()
+        if auth:
+            if cache_credentials:
+                await self.ensure_credential_helper(path)
+            env["GIT_TERMINAL_PROMPT"] = "1"
+            code, _, fetch_error = await execute(
+                cmd,
+                cwd=cwd,
+                username=auth["username"],
+                password=auth["password"],
+                env=env,
+            )
+        else:
+            env["GIT_TERMINAL_PROMPT"] = "0"
+            code, _, fetch_error = await execute(cmd, cwd=cwd, env=env)
 
         result = {
             "code": code,
@@ -998,13 +1013,17 @@ class Git:
             return {"code": code, "command": " ".join(cmd), "message": error}
         return {"code": code}
 
-    async def pull(self, path, auth=None, cancel_on_conflict=False):
+    async def pull(
+        self, path, auth=None, cancel_on_conflict=False, cache_credentials=False
+    ):
         """
         Execute git pull --no-commit.  Disables prompts for the password to avoid the terminal hanging while waiting
         for auth.
         """
         env = os.environ.copy()
         if auth:
+            if cache_credentials:
+                await self.ensure_credential_helper(path)
             env["GIT_TERMINAL_PROMPT"] = "1"
             code, output, error = await execute(
                 ["git", "pull", "--no-commit"],
@@ -1048,7 +1067,14 @@ class Git:
         return response
 
     async def push(
-        self, remote, branch, path, auth=None, set_upstream=False, force=False
+        self,
+        remote,
+        branch,
+        path,
+        auth=None,
+        set_upstream=False,
+        force=False,
+        cache_credentials=False,
     ):
         """
         Execute `git push $UPSTREAM $BRANCH`. The choice of upstream and branch is up to the caller.
@@ -1062,6 +1088,8 @@ class Git:
 
         env = os.environ.copy()
         if auth:
+            if cache_credentials:
+                await self.ensure_credential_helper(path)
             env["GIT_TERMINAL_PROMPT"] = "1"
             code, output, error = await execute(
                 command,
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -8,6 +8,7 @@ import re
 import shlex
 import subprocess
 import traceback
+from typing import Any, Dict, List, Optional
 from urllib.parse import unquote
 
 import nbformat
@@ -173,6 +174,8 @@ class Git:
     """
     A single parent class containing all of the individual git methods in it.
     """
+
+    _GIT_CREDENTIAL_CACHE_DAEMON_PROCESS: subprocess.Popen = None
 
     def __init__(self, config=None):
         self._config = config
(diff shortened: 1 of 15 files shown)


jupyterlab_git/handlers.py
```diff
@@ -23,7 +23,7 @@
 from .log import get_logger
 
 # Git configuration options exposed through the REST API
-ALLOWED_OPTIONS = ["user.name", "user.email"]
+ALLOWED_OPTIONS = ["user.name", "user.email", "credential.helper"]
```
fcollonval, line 26: Don't expose the helper to the frontend.

```suggestion
ALLOWED_OPTIONS = ["user.name", "user.email"]
```

jupyterlab_git/tests/test_clone.py
```diff
@@ -156,3 +158,115 @@ async def test_git_clone_with_auth_auth_failure_from_git():
...
+
+
+@pytest.mark.asyncio
+async def test_git_clone_with_cache_credentials():
+    with patch("jupyterlab_git.git.execute") as mock_execute:
+        # Given
+        test_path = str(Path("/bin") / "test_curr_path")
+        mock_execute.side_effect = [
+            maybe_future((0, "", "")),
+            maybe_future((0, "", "")),
```
fcollonval, line 170: Could you add a comment to tell which call this is addressing?

src/components/StatusWidget.tsx
```diff
@@ -72,6 +115,12 @@ export class StatusWidget extends ReactWidget {
    */
   private _status = '';
 
+  /**
+   * Boolean indicating whether credentials are required.
+   */
+  private _waitingForCredentials: boolean;
+
```
fcollonval, line 122: We have the model, so no need for this

```suggestion
```

src/components/StatusWidget.tsx
```diff
@@ -32,21 +39,57 @@ export class StatusWidget extends ReactWidget {
...
+    return this._waitingForCredentials;
+  }
+
+  /**
+   * Sets the boolean indicating whether credentials are required.
+   */
+  set waitingForCredentials(value: boolean) {
+    this._waitingForCredentials = value;
+  }
+
```
fcollonval, line 55: We will use the model signal directly - the flaw of the current code is that we don't update the widget when setting the value (it seems there is parasite update request that are giving a wrong impression of live udpate).

```suggestion
```

src/components/StatusWidget.tsx
```diff
@@ -32,21 +39,57 @@ export class StatusWidget extends ReactWidget {
...
   render(): JSX.Element {
     return (
-      <div title={`Git: ${this._trans.__(this._status)}`}>
-        <gitIcon.react
-          className={
+      <Badge
```
fcollonval, line 58: We can wrap this to react to the new signal:

```tsx
<UseSignal signal={model.credentialsRequiredSignal} initialArgs={false} >
  {(model, needsCredential) => (<Badge...</Badge>)}
</UseSignal>

File: src/components/StatusWidget.tsx [hunk 6]
```diff
@@ -92,8 +141,14 @@ export function addStatusBarWidget(
...
   const callback = Private.createEventCallback(statusWidget);
   model.taskChanged.connect(callback);
+
+  const credentialsRequiredCallback =
+    Private.createCredentialsRequiredCallback(statusWidget);
+  model.credentialsRequiredSignal.connect(credentialsRequiredCallback);
+
   statusWidget.disposed.connect(() => {
     model.taskChanged.disconnect(callback);
+    model.credentialsRequiredSignal.disconnect(credentialsRequiredCallback);
```
fcollonval, line 151:
```suggestion
  statusWidget.disposed.connect(() => {
    model.taskChanged.disconnect(callback);
```

src/components/StatusWidget.tsx
```diff
@@ -200,5 +255,32 @@ namespace Private {
...
+     * @returns void
+     */
+    function callbackCredentialsRequired(
+      model: IGitExtension,
+      value: boolean
+    ): void {
+      widget.waitingForCredentials = value;
+    }
+    return callbackCredentialsRequired;
+  }
```
fcollonval, line 284:
```suggestion
```

src/model.ts
```diff
@@ -291,6 +291,27 @@ export class GitExtension implements IGitExtension {
...
+    if (this._credentialsRequired !== value) {
+      this._credentialsRequired = value;
+      this._credentialsRequiredSignal.emit(value);
+    }
+  }
+
+  /**
+   * A signal emitted whenever credentials are required, or are not required anymore.
+   */
+  get credentialsRequiredSignal(): ISignal<IGitExtension, boolean> {
```
fcollonval, line 311: Let's use a more Jupyter wording

```suggestion
  get credentialsRequiredChanged(): ISignal<IGitExtension, boolean> {
```

jupyterlab_git/handlers.py
```diff
@@ -73,7 +73,10 @@ async def post(self, path: str = ""):
         """
         data = self.get_json_body()
         response = await self.git.clone(
-            self.url2localpath(path), data["clone_url"], data.get("auth", None)
+            self.url2localpath(path),
+            data["clone_url"],
+            data.get("auth", None),
+            data.get("cache_credentials", False),
```
fcollonval, line 79: Let's propagate the extended auth object from the frontend directly (i.e. `cache_credentials` will be a new key in `auth`)

The reason being it is that the credentials caching without authentication information has no effect.
Could you also propagate that to the Git clone, push, ... methods.

jupyterlab_git/git.py
```diff
@@ -1540,3 +1595,142 @@ async def tag_checkout(self, path, tag):
                 "command": " ".join(command),
                 "message": error,
             }
+
+    async def check_credential_helper(self, path: str) -> Dict[str, Any]:
```
fcollonval, line 1599: as we do not intent to report this back to the frontend - please return straigh type or raise an error.

in this case I would return True, False or None
True => has helper and needs cache daemon
False => has helper and does not need cache daemon
None => No helper

jupyterlab_git/git.py
```diff
@@ -1540,3 +1595,142 @@ async def tag_checkout(self, path, tag):
...
+        """
+        Check whether `git config --list` contains `credential.helper`.
+        If it is not set, then it will be set to the value string for `credential.helper`
+        defined in the server settings.
+
+        path: str
+            Git path repository
+        env: Dict[str, str]
+            Environment variables
+        """
```
fcollonval, line 1631: Same here, don't return.

Be forgiven if it fails, and log a warning message if an error occurred.

To goal to be forgiven is that the git command may succeed with the authentication information even if the credentials mechanism is not working. So the primary goal of the user (aka the git command) will still be executed.

jupyterlab_git/git.py
```diff
@@ -1540,3 +1595,142 @@ async def tag_checkout(self, path, tag):
...
+                await self.config(path, **{"credential.helper": credential_helper})
+            )
+            if GIT_CREDENTIAL_HELPER_CACHE.match(credential_helper.strip()):
+                cache_daemon_required = True
+        else:
+            response["code"] = 0
+
+        # special case: Git credential cache
+        if cache_daemon_required:
+            if not sys.platform.startswith("win32"):
```
fcollonval, line 1657: Are you sure of this? I would rather avoid over constrained code.

jupyterlab_git/git.py
```diff
@@ -1540,3 +1595,142 @@ async def tag_checkout(self, path, tag):
...
+                "A credential cache daemon has been spawned with PID %s",
+                str(self._GIT_CREDENTIAL_CACHE_DAEMON_PROCESS.pid),
+            )
+
+        elif self._GIT_CREDENTIAL_CACHE_DAEMON_PROCESS.poll():
+            return self.ensure_git_credential_cache_daemon(
+                socket, debug, True, cwd, env
+            )
+
+        return self._GIT_CREDENTIAL_CACHE_DAEMON_PROCESS.pid
```
fcollonval, line 1732: What is the goal of returning the PID?

jupyterlab_git/git.py
quachtridat: I am not sure of this. Initially I wanted to only allow spawning the daemon on Linux, but now that you say _over constrained code_, I believe it is good to not have this conditional here as I don't know whether it is ***really*** possible to have Unix socket support on Windows.

jupyterlab_git/git.py
quachtridat: The initial goal was so that we could say _-1_ for when the process is not under `JupyterLab-Git`'s control, and when it is not _-1_ we could use it for debug logging outside of this function. There are no other uses, and I can let this function return `None` instead.

jupyterlab_git/git.py
```diff
@@ -1540,3 +1595,142 @@ async def tag_checkout(self, path, tag):
...
+
+            self._GIT_CREDENTIAL_CACHE_DAEMON_PROCESS = subprocess.Popen(
+                args,
+                cwd=cwd,
+                env=env,
+            )
+
+            get_logger().debug(
+                "A credential cache daemon has been spawned with PID %s",
+                str(self._GIT_CREDENTIAL_CACHE_DAEMON_PROCESS.pid),
```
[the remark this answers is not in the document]
quachtridat: Can you explain why we are using f-strings with loggers? I thought that it is recommended to do the `%d` or `%s` *"format"* way.

jupyterlab_git/git.py
```diff
@@ -279,13 +283,8 @@ async def clone(self, path, repo_url, auth=None, cache_credentials=None):
...
         env = os.environ.copy()
         if auth:
-            if cache_credentials:
-                ensured_response = await self.ensure_credential_helper(path)
-                if ensured_response["code"] != 0:
-                    return {
-                        "code": ensured_response["code"],
-                        "message": ensured_response.get("error", "Unknown error!"),
-                    }
+            if auth.get("cache_credentials", None):
```
fcollonval, line 286: The default fallback value is `None`, so we can simply write:

```suggestion
            if auth.get("cache_credentials"):
```

jupyterlab_git/git.py
```diff
@@ -1630,81 +1621,63 @@ async def ensure_credential_helper(
...
+            if has_credential_helper == False:
+                return
+        except RuntimeError as e:
+            get_logger().error("Error checking credential helper: %s", e, exc_info=True)
+            return
 
-        response = {"code": -1}
+        cache_daemon_required = has_credential_helper == True
 
         if not has_credential_helper:
```
fcollonval, line 1634: Let's be explicit

```suggestion
        if has_credential_helper is None:
```

fcollonval (changes requested):
Thanks @quachtridat for applying the suggestions.

Would you mind fixing the CI?

> I added two minor suggestions more.

quachtridat:
> Thanks @quachtridat for applying the suggestions.
> 
> Would you mind fixing the CI?
> 
> > I added two minor suggestions more.

Yes I am working on fixing the CI!

fcollonval (approved):
Thanks a lot @quachtridat and @ZeshanFayyaz

PR #1108, opened 2022-04-14
Add the feature for comparing two arbitrary commits

### LINKED ISSUE #985: Add ability to display the diff of any two commits in the file history panel
Select two arbitrary commits and view their diff (`Select for compare` & `Compare with selected` from VSCode)

_Originally posted by @navn-r in https://github.com/jupyterlab/jupyterlab-git/issues/980#issuecomment-886121034_

quachtridat:
## Summary

This PR resolves #985 by implementing the necessary components and logic that will allow users to select two commits in the *History* tab and see the Git diff for every changed file.

### Backend

https://github.com/jupyterlab/jupyterlab-git/pull/1108/commits/5cee9bfec6dafe0dfcc826cc83b34e2b61bc3b8b

### Frontend

Each entry in `History` tab now has two buttons, each with an icon of a document containing an arrow pointing right/left. Clicking one of these will make the commit comparison box pop up, and the diff will be evaluated when there are two commits selected for comparison. Stylings are similar to how the `Changed` section for a commit entry in `History` is styled.

## Caveats

The commit comparison box can be collapsed, but cannot be resized for now.

## Demos

image
*The commit that is __selected for compare__ has a little red color on the left side.*

image
*The commit that is to be __compared with selected__ has a little green color on the left side.*

image
*The comparison results are shown in the commit comparison box at the bottom of the `History` tab.*

image
*The diff for the file `README.md` is shown when the diff button for that file is clicked.*

image
*The commit comparison box when it is collapsed.*

## Notes

This PR may need more tests.

## Linked issues

#985

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -261,7 +261,7 @@ class Git:
 
         return response
 
-    async def clone(self, path, repo_url, auth=None):
+    async def clone(self, path, repo_url, auth=None, cache_credentials=None):
         """
         Execute `git clone`.
         When no auth is provided, disables prompts for the password to avoid the terminal hanging.
@@ -273,6 +273,8 @@ class Git:
         """
         env = os.environ.copy()
         if auth:
+            if cache_credentials:
+                await self.ensure_credential_helper(path)
             env["GIT_TERMINAL_PROMPT"] = "1"
             code, output, error = await execute(
                 ["git", "clone", unquote(repo_url), "-q"],
@@ -296,7 +298,7 @@ class Git:
 
         return response
 
-    async def fetch(self, path):
+    async def fetch(self, path, auth=None, cache_credentials=False):
         """
         Execute git fetch command
         """
@@ -308,8 +310,21 @@ class Git:
             "--all",
             "--prune",
         ]  # Run prune by default to help beginners
-
-        code, _, fetch_error = await execute(cmd, cwd=cwd)
+        env = os.environ.copy()
+        if auth:
+            if cache_credentials:
+                await self.ensure_credential_helper(path)
+            env["GIT_TERMINAL_PROMPT"] = "1"
+            code, _, fetch_error = await execute(
+                cmd,
+                cwd=cwd,
+                username=auth["username"],
+                password=auth["password"],
+                env=env,
+            )
+        else:
+            env["GIT_TERMINAL_PROMPT"] = "0"
+            code, _, fetch_error = await execute(cmd, cwd=cwd, env=env)
 
         result = {
             "code": code,
@@ -998,13 +1013,17 @@ class Git:
             return {"code": code, "command": " ".join(cmd), "message": error}
         return {"code": code}
 
-    async def pull(self, path, auth=None, cancel_on_conflict=False):
+    async def pull(
+        self, path, auth=None, cancel_on_conflict=False, cache_credentials=False
+    ):
         """
         Execute git pull --no-commit.  Disables prompts for the password to avoid the terminal hanging while waiting
         for auth.
         """
         env = os.environ.copy()
         if auth:
+            if cache_credentials:
+                await self.ensure_credential_helper(path)
             env["GIT_TERMINAL_PROMPT"] = "1"
             code, output, error = await execute(
                 ["git", "pull", "--no-commit"],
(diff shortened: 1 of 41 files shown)


src/style/PastCommitNode.ts
```diff
@@ -83,3 +83,11 @@ export const singleFileCommitClass = style({
     }
   }
 });
+
+export const commitCompareLhsNodeClass = style({
+  borderLeft: '3px solid darkred'
```
fcollonval, line 88: Could you try with 
```suggestion
  borderLeft: '3px solid var(--jp-git-diff-deleted-color)'
```

src/style/PastCommitNode.ts
```diff
@@ -83,3 +83,11 @@ export const singleFileCommitClass = style({
...
+});
+
+export const commitCompareRhsNodeClass = style({
+  borderLeft: '3px solid darkgreen'
```
fcollonval, line 92:
```suggestion
  borderLeft: '3px solid var(--jp-git-diff-added-color)'
```

fcollonval (changes requested):
Thanks @quachtridat and @ZeshanFayyaz 

I mainly have suggestions for a better wording (using reference/challenger instead of lhs/rhs) and a small rework to avoid storing twice the same information in the panel state.

jupyterlab_git/git.py
```diff
@@ -630,13 +636,14 @@ async def diff(self, path):
...
         for line in line_array:
             linesplit = line.split()
-            result.append(
-                {
-                    "insertions": linesplit[0],
-                    "deletions": linesplit[1],
-                    "filename": linesplit[2],
-                }
-            )
+            if len(linesplit) == 3:
```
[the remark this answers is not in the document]
quachtridat: > Why is this test required? In which case are those information missing?

I have not manually tested without this, but in early development of this feature I got an error without that line (and it was `IndexError` if I remember correctly), and no error with that line.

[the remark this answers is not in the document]
fcollonval: thanks @quachtridat would you mind adding a else condition that log at debug level the _broken_ line?

src/tokens.ts
```diff
@@ -1000,6 +1031,25 @@ export namespace Git {
...
+   */
+  export interface ICommitComparison {
+    /**
+     * The commit to compare against
+     */
+    lhs: ISingleCommitInfo;
+    /**
+     * The commit to compare
+     */
+    rhs: ISingleCommitInfo;
```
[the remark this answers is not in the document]
quachtridat: That means we should just pass `referenceCommit`, `challengerCommit` and `comparison` one-by-one as props for `CommitComparisonBox`?

src/components/CommitComparisonBox.tsx
```diff
@@ -0,0 +1,257 @@
...
+  comparison: Git.ICommitComparison | null;
+  header: string;
+  model: IGitExtension;
+  trans: TranslationBundle;
+  onCancel?: (event: React.MouseEvent<HTMLElement, MouseEvent>) => void;
+  onOpenDiff?: (
+    filePath: string,
+    isText: boolean,
+    previousFilePath?: string
+  ) => (event: React.MouseEvent<HTMLLIElement, MouseEvent>) => void;
```
quachtridat, line 228: TODO: docstring.
@ZeshanFayyaz

quachtridat, line 228: TODO: docstring.
@ZeshanFayyaz

jupyterlab_git/git.py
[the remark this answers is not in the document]
quachtridat: > thanks @quachtridat would you mind adding a else condition that log at debug level the _broken_ line?

Sure!
@ZeshanFayyaz we might as well investigate this.

jupyterlab_git/handlers.py
```diff
@@ -258,7 +258,16 @@ async def post(self, path: str = ""):
         POST request handler, fetches differences between commits & current working
```
fcollonval, line 258:
```suggestion
        POST request handler, fetches differences between two states
```

fcollonval (approved):
Thanks a lot @quachtridat and @ZeshanFayyaz 

This is a great addition

fcollonval:
@all-contributors please add quachtridat for code

fcollonval:
@all-contributors please add @ZeshanFayyaz for code

PR #1122, opened 2022-04-25
fixes #1121 by trimming whitespace from user input of clone URI path

### LINKED ISSUE #1121: Trim whitespace around git clone path
### Summary
If someone cuts and pastes a url that has whitespace in front of it the clone operation fails (extra spaces is pretty common if following along with a blog tutorial).

### Details
There will be a little error box in the bottom right and if you show the details you will see something like:
`Cloning into 'metrobus-repairs-nlp-workshop'... fatal: protocol ' https' is not supported`

### Fix
To fix, I think it's as simple as `trim()` - I think that might go here:
https://github.com/jupyterlab/jupyterlab-git/blob/9df356280b544b8f0e2236ef0757d57a6239ca7c/src/commandsAndMenu.tsx#L310

dudash:
fixes #1121

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ git clone https://github.com/jupyterlab/jupyterlab-git.git
 # Change directory to the jupyterlab-git directory
 cd jupyterlab-git
 # Install package in development mode
-pip install -e .[dev]
+pip install -e ".[dev]"
 pre-commit install
 # Link your development version of the extension with JupyterLab
 jupyter labextension develop . --overwrite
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -207,7 +207,7 @@ git clone https://github.com/jupyterlab/jupyterlab-git.git
 # Change directory to the jupyterlab-git directory
 cd jupyterlab-git
 # Install package in development mode
-pip install -e .[dev]
+pip install -e ".[dev]"
```
dudash, line 210: These quotes were needed to run on the Mac

fcollonval (approved):
Thanks a lot @dudash

Issue #1100, opened 2022-03-24
Git branch name in the status bar


stefano-cherchi:
## Is your feature request related to a problem? Please describe.
While you're working on the notebook you can't see which branch you're working on, unless you select the Git extension tab. I noticed that this can be confusing and lead to errors especially for developers with limited experience on git.

## Describe the solution you'd like
It would be nice to have the name of the current branch in the bottom status bar, next to the git status icon, like in VSCode or other IDEs.

- Python package version:
```
$ pip show jupyterlab-git
Name: jupyterlab-git
Version: 0.34.0
Summary: A JupyterLab extension for version control using git
Home-page: https://github.com/jupyterlab/jupyterlab-git
Author: Jupyter Development Team
Author-email: None
License: BSD-3-Clause
Location: /app/anaconda3-2021.05/lib/python3.8/site-packages
Requires: pexpect, nbformat, jupyter-server, nbdime, packaging
Required-by:
```

- Extension version:
```
$ jupyter labextension list
/app/anaconda3-2021.05/lib/python3.8/site-packages/jupyter_server/transutils.py:13: FutureWarning: The alias `_()` will be deprecated. Use `_i18n()` instead.
  warnings.warn(warn_msg, FutureWarning)
JupyterLab v3.0.14
/app/anaconda3-2021.05/share/jupyter/labextensions
        jupyterlab-sparkmonitor v3.0.1 enabled OK (python, jupyterlab_sparkmonitor)
        nbdime-jupyterlab v2.1.0 enabled OK
        jupyterlab-execute-time v2.0.5 enabled OK (python, jupyterlab_execute_time)
        jupyterlab-plotly v5.6.0 enabled OK
        @jupyter-widgets/jupyterlab-manager v3.0.0 enabled OK (python, jupyterlab_widgets)
        @jupyterlab/git v0.34.0 enabled OK (python, jupyterlab-git)

Other labextensions (built into JupyterLab)
   app dir: /app/anaconda3-2021.05/share/jupyter/lab
```

- Git version:
```
$ git --version
git version 2.32.0
```

- Operating System and its version:
```
$ cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 7.3 (Maipo)
```

fcollonval:
Thanks for the suggestion. I can provide some pointers to implement this. I would only suggest to make it configurable as some users like not too crowded status bar.

PR #1131, opened 2022-06-10
add current branch name on status bar

BoscoCHW:
Fixes https://github.com/jupyterlab/jupyterlab-git/issues/1100

@fcollonval 
Added current branch name on the status bar
I managed to style it but I'm not sure how to center the text vertically.
I should make the parent element a flex box and do align-item: center but I don't know how to select the parent class.
Also I'm not sure how to change the padding of the git icon so that the text could be closer to the icon.

Another issue is when navigate to another repo (change file directory), the branch name stays unless you click on the git extension icon on the left menu bar, which is a bit weird. 

---

Edited to link the related issue.

### DIFF
--- a/src/components/StatusWidget.tsx
+++ b/src/components/StatusWidget.tsx
@@ -10,7 +10,8 @@ import { gitIcon } from '../style/icons';
 import {
   badgeClass,
   statusAnimatedIconClass,
-  statusIconClass
+  statusIconClass,
+  currentBranchClass
 } from '../style/StatusWidget';
 import { toolbarButtonClass } from '../style/Toolbar';
 import { IGitExtension } from '../tokens';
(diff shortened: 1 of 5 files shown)


src/components/StatusWidget.tsx
```diff
@@ -46,32 +47,39 @@ export class StatusWidget extends ReactWidget {
...
+                    ? `Git: ${this._trans.__('credentials required')}`
+                    : `Git: ${this._trans.__(this._status)}`
+                }
+              />
+            </Badge>
+            {this._model.currentBranch && (
+              <span className={currentBranchClass}>
+                {this._model.currentBranch.name}*
+              </span>
+            )}
```
fcollonval, line 81: You should place this outside the `UseSignal` component as it does not need to be refreshed if the credentials status of the extension changes.

But you should wrap it in another `UseSignal` component to listen to `this._model.headChanged` to react when the git head is changed.

fcollonval (changes requested):
Thanks @BoscoCHW 

> Added current branch name on the status bar I managed to style it but I'm not sure how to center the text vertically. I should make the parent element a flex box and do align-item: center but I don't know how to select the parent class. Also I'm not sure how to change the padding of the git icon so that the text could be closer to the icon.

You can add a class in `StatusWidget` (the parent element) within the [constructor](https://github.com/jupyterlab/jupyterlab-git/blob/master/src/components/StatusWidget.tsx#L27) by adding the following line:

```ts
this.addClass('jp-git-StatusWidget')
```

Note: by design it is not possible to select parent of an element with CSS selectors.

> Another issue is when navigate to another repo (change file directory), the branch name stays unless you click on the git extension icon on the left menu bar, which is a bit weird.

The default behavior is to not refresh the Git model status when the sidebar is hidden to avoid useless calls to the backend. You could solve this by forcing refreshing the branch calling [GitExtension.refreshBranch](https://github.com/jupyterlab/jupyterlab-git/blob/master/src/model.ts#L981) when we set a [new repository](https://github.com/jupyterlab/jupyterlab-git/blob/master/src/index.ts#L151).

BoscoCHW:
@fcollonval 

I managed to style the status widget by adding a class like you suggested, but I was unable to reduce the right-padding of the git icon on the status bar because I'm afraid overriding the pre-set would lead to unexpected results.

src/components/StatusWidget.tsx
```diff
@@ -41,39 +44,51 @@ export class StatusWidget extends ReactWidget {
...
+                  needsCredentials
+                    ? `Git: ${this._trans.__('credentials required')}`
+                    : `Git: ${this._trans.__(this._status)}`
+                }
+              />
+            </Badge>
+          )}
+        </UseSignal>
+
+        <UseSignal signal={this._model.headChanged}>
```
fcollonval, line 82: :+1:

src/components/StatusWidget.tsx
```diff
@@ -41,39 +44,51 @@ export class StatusWidget extends ReactWidget {
...
+          {() =>
+            this._model.currentBranch && (
+              <span className={currentBranchNameClass}>
+                {this._model.currentBranch.name}*
```
fcollonval, line 86: Let's remove the star for now.

```suggestion
                {this._model.currentBranch.name}
```

fcollonval (changes requested):
Let's remove the star next to the branch name. And it is good to go.

fcollonval (approved):
Thanks @BoscoCHW

Issue #911, opened 2021-04-02
Discard Changes Should close open diff


ianhi:
## Describe the solution you'd like
It would be nice if when I:
1. Open a diff
2. Discard the changes

The open diff would be closed. Or perhaps there would be a setting enabling this behavior.

## Additional context

vscode does not do this - but github desktop does. So I think either way is reasonable and it would be nice to have a setting for this.

PR #1132, opened 2022-06-11
close diff window when changes to file are discarded

BoscoCHW:
@fcollonval 
Implemented closing of diff window when the changes to file are discarded, but I copied a lot of codes from the gitFileDiff command

Fixes https://github.com/jupyterlab/jupyterlab-git/issues/911

---

Edited to link the issue

### DIFF
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -607,7 +607,7 @@ export function addCommands(
             modelIsLoading.reject(msg);
           }
 
-          const slotRegistered = gitModel.statusChanged.connect((_, status) => {
+          gitModel.statusChanged.connect((_, status) => {
             const targetFile = status.files.find(
               fileStatus => model.filename === fileStatus.from
             );
@@ -615,7 +615,6 @@ export function addCommands(
               mainAreaItem.dispose();
             }
           });
-          console.log(slotRegistered);
         }
 
         return mainAreaItem;
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -606,6 +606,16 @@ export function addCommands(
             })`;
             modelIsLoading.reject(msg);
           }
+
+          const slotRegistered = gitModel.statusChanged.connect((_, status) => {
+            const targetFile = status.files.find(
+              fileStatus => model.filename === fileStatus.from
+            );
+            if (!targetFile || targetFile.status === 'unmodified') {
+              mainAreaItem.dispose();
+            }
+          });
+          console.log(slotRegistered);
         }
 
         return mainAreaItem;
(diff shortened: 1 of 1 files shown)


fcollonval:
Thanks @BoscoCHW 

We may take a slightly different approach to avoid code duplication.

When creating the diff widget at https://github.com/jupyterlab/jupyterlab-git/blob/master/src/commandsAndMenu.tsx#L854, we could add a callback on `gitModel.statusChanged` just after that. If the `file` is not changed any more (you will need to introspect `gitModel.status.files`), we will call `widget.dispose()` to close it.

BoscoCHW:
@fcollonval 
I implemented it by adding a callback when in the gitShowDiff command, though, I'm not sure if the implementation is good.
I have two questions:
1. What is the purpose for the "unmodified" status? Based on my observation, discarding changes will remove the file from `status.files`
2. Do we need to disconnect the slot/callback when the widget is closed? I'm not sure how to do it.

fcollonval:
> I implemented it by adding a callback when in the gitShowDiff command. 

The implementation looks good, thanks for changing it.

> I have two questions: 
>     1. What is the purpose for the "unmodified" status? Based on my observation, discarding changes will remove the file from `gitModel.status.files`

The _unmodified_ status is set [there](https://github.com/jupyterlab/jupyterlab-git/blob/master/src/model.ts#L386). It is set only when triggering the [file browser context menu](https://github.com/jupyterlab/jupyterlab-git/blob/master/src/commandsAndMenu.tsx#L1295) because we want to display the menu to open the git history for a versioned file that is not currently modified.

But in this case, this is as you figure out. If the file is not listed in `gitModel.status.files`, it is unchanged. Hence you can close the diff widget.

>     2. Do we need to disconnect the slot/callback when the widget is closed? I'm not sure how to do it.

Theoretically, it is a good practice to do so. But the widget itself has a clean up code when disposed to remove any slot.

src/commandsAndMenu.tsx
```diff
@@ -606,6 +606,16 @@ export function addCommands(
...
+
+          const slotRegistered = gitModel.statusChanged.connect((_, status) => {
+            const targetFile = status.files.find(
+              fileStatus => model.filename === fileStatus.from
+            );
+            if (!targetFile || targetFile.status === 'unmodified') {
+              mainAreaItem.dispose();
+            }
+          });
+          console.log(slotRegistered);
```
fcollonval, line 618: Perfect :clap: 

I presume the `console.log` comes from your test to understand the code.

```suggestion
          gitModel.statusChanged.connect((_, status) => {
            const targetFile = status.files.find(
              fileStatus => model.filename === fileStatus.from
            );
            if (!targetFile || targetFile.status === 'unmodified') {
              mainAreaItem.dispose();
            }
          });
```

fcollonval (approved):
Thanks a lot @BoscoCHW

PR #1134, opened 2022-06-15
Feature 866 - Display Commit Description

### LINKED ISSUE #866: Where can I find the commit description
Hi there,

For committing, a commit message could be input including a title and description. After committing, I can see the commit title, but cannot find the description. Just wondering it is lost or just hidden somewhere? If hidden, where can I find this description because usually put lots of update information there so it is easy for other people to get the update idea at the first glance.

thanks

iflinda:
Enhancement #866 

Modified the `detailed_log` function to return the entirety of the commit body when a user views additional details in the history. It will then split at the first `\n`, and if a title exists then it will return the last index of the message only. 

I also adding minor padding for styling.

Example:

image

## Issues

1. The current iteration of this enhancement does not format the commit description in a readable way. Additional care should be taken to have better readability.

    Example: 
    image

2. I still have to create a test case to check the modified `detailed_log` as there are two execute functions. Currently the test will fail in `test_detailed_log` due to this.


src/components/SinglePastCommitInfo.tsx
```diff
@@ -159,47 +166,50 @@ export class SinglePastCommitInfo extends React.Component<
...
-        }
-        numFiles={this.state.numFiles}
-        insertions={this.state.insertions}
-        deletions={this.state.deletions}
-        files={this.state.modifiedFiles}
-        onOpenDiff={this.props.onOpenDiff}
-        trans={this.props.trans}
-      ></CommitDiff>
+      <div>
+        <p className={commitBodyClass}>{this.state.commitBody}</p>
```
fcollonval, line 170: I find it nice to display the detailed only when expanding the detailed. This keeps the first view as compact as possible.

For the format, you probably need to replace the `\n` with `<br />`. Probably something like:

```ts
const body = [];
this.state.commitBody.split('\n').forEach(line => { body.push(line, <br />) });

// ...

<p className={commitBodyClass}>{body}</p>
```

jupyterlab_git/git.py
```diff
@@ -555,14 +570,26 @@ async def detailed_log(self, selected_hash, path):
...
+        commit_cmd = [
+            "git",
+            "log",
+            "-1",
+            "--pretty=format:%B",
+            selected_hash,
+        ]
+
+        code, my_output = await execute_log_command(cmd, path)
+        my_commit_output = await execute_log_command(commit_cmd, path)
```
fcollonval, line 592: Playing a bit with the command, using the following single command:

`git log -1 --numstat --pretty=format:%B%x00 -z <hash>` will give you all required information at once. The first item will be the subject + body.

> Shouldn't you use only `%b` to get the body and not the subject?

jupyterlab_git/git.py
```diff
@@ -609,6 +636,7 @@ async def detailed_log(self, selected_hash, path):
 
         return {
             "code": code,
+            "commit_body": strip_and_split(my_commit_output.split("\n", 1)[-1]),
```
fcollonval, line 639: Why are you keeping only the last item?

If this is to drop the commit summary, is it not enough to use `%b` instead of `%B` in the format.

fcollonval (changes requested):
Thanks @iflinda 

I have a couple of questions/suggestions.

> When you are opening a WIP PR, I advice you to use the recent feature of `Draft PR`. This makes it clearer in GitHub :wink:

jupyterlab_git/git.py
iflinda: I ended up using `git log -1 --numstat --pretty=format:%b%x00 -z <hash>`.

iflinda (commented):
Hi @fcollonval, I made some changes based on your recommendations. 😊 Thank you for the feedback.

jupyterlab_git/git.py
```diff
@@ -555,19 +555,30 @@ async def detailed_log(self, selected_hash, path):
...
             return {"code": code, "command": " ".join(cmd), "message": my_error}
 
         total_insertions = 0
         total_deletions = 0
         result = []
         line_iterable = iter(strip_and_split(my_output)[1:])
+
+        commit_body = my_output.split("\x00", 1)[0].strip()
+
         for line in line_iterable:
```
fcollonval, line 582: As the output is split by `strip_and_split`, we can simplify further the code.

```suggestion
        line_iterable = strip_and_split(my_output)

        commit_body = line_iterable[0]

        for line in line_iterable[1:]:
```

jupyterlab_git/tests/test_detailed_log.py
```diff
@@ -13,7 +13,7 @@ async def test_detailed_log():
     with patch("jupyterlab_git.git.execute") as mock_execute:
         # Given
         process_output = [
-            "f29660a (HEAD, origin/feature) Commit message",
+            "    Test Description with leading and trailing spaces    ",
             "10\t3\tnotebook_without_spaces.ipynb",
```
fcollonval, line 17: Let's add a return line character as it is present in the git output.

```suggestion
            "\n10\t3\tnotebook_without_spaces.ipynb",
```

fcollonval (changes requested):
Thanks a lot @iflinda

I have two minor suggestions and we should be good to go.

jupyterlab_git/git.py
iflinda: Hey @fcollonval,

I tried using this suggestion, but when I test it in the JupyterLab extension--the output looks like the following:

image

It does not seem to work for commits descriptions. It works fine for commits with a summary and description.

fcollonval: Thanks Linda for trying. We are hit by the initial strip of `\x00` in front the string when there is no description.

Could you try the following a first split as you implemented; but using the second part to list the files:

```py
      first_split = my_output.split("\x00", 1)
      commit_body = first_split[0].strip()
      line_iterable = strip_and_split(first_split[1]))
      for line in line_iterable:
```

jupyterlab_git/tests/test_detailed_log.py
iflinda: Hey @fcollonval,

I noticed that when we add the `"\n10\t3\tnotebook_without_spaces.ipynb"` that the return does not actually strip the `\n` at this point in time; thus, failing the test.

Would I want to modify the `strip_and_split` function to also strip out new lines?

Before:
`return s.strip("\x00").split("\x00")`
After:
`return s.strip("\x00").strip("\n").split("\x00")`

jupyterlab_git/git.py
iflinda: Hi @fcollonval this worked, and I've pushed this change with this update.

jupyterlab_git/tests/test_detailed_log.py
fcollonval: Good suggestion @iflinda let's do that.

iflinda (commented):
Made recommended changes based on feedback.

fcollonval (approved):
Thanks a lot @iflinda 

This looks good.

PR #1135, opened 2022-06-15
Add 'open file' command to context menu on diff tab

### LINKED ISSUE #941: Open the diffed file from the diff tab context menu and/or toolbar
## Is your feature request related to a problem? Please describe.
I would like to be able to go to the file I am diffing from the diff easily. It is usually open in a tab, but I have many tabs open and sometimes getting lost.

## Describe the solution you'd like

I would like to have a button "Open the diffed file" (which would usually just bring it up to focus rather than open) on the toolbar and/or similar option in the context menu when clicking on the diff tab:

Screenshot from 2021-05-26 15-08-09

## Additional context

- Python package version: 0.30.1
- Extension version: 0.30.1

BoscoCHW:
Fixes #941

### DIFF
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -496,7 +496,7 @@ export function addCommands(
         getDiffProvider(fullPath) ?? (isText && createPlainTextDiff);
 
       if (buildDiffWidget) {
-        const id = `diff-${fullPath}-${model.reference.label}-${model.challenger.label}`;
+        const id = `git-diff-${fullPath}-${model.reference.label}-${model.challenger.label}`;
         const mainAreaItems = shell.widgets('main');
         let mainAreaItem = mainAreaItems.next();
         while (mainAreaItem) {
@@ -757,7 +757,7 @@ export function addCommands(
     execute: async _ => {
       const widget = app.contextMenuHitTest((node: HTMLElement) => {
         const nodeId = node.dataset.id;
-        return nodeId && nodeId.substring(0, 4) === 'diff';
+        return nodeId && nodeId.substring(0, 8) === 'git-diff';
       });
       if (!widget) {
         return;
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -755,25 +755,48 @@ export function addCommands(
     label: trans.__('Open File'),
     caption: trans.__('Open file from its diff view'),
     execute: async _ => {
-      const widget = app.contextMenuHitTest((node: HTMLElement) => {
+      const domNode = app.contextMenuHitTest((node: HTMLElement) => {
         const nodeId = node.dataset.id;
         return nodeId && nodeId.substring(0, 8) === 'git-diff';
       });
-      if (!widget) {
+      if (!domNode) {
         return;
       }
 
-      const filename = widget.title;
-      const file = gitModel.status.files.find(
-        fileStatus => `${gitModel.pathRepository}/${fileStatus.to}` === filename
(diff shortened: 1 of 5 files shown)


BoscoCHW:
@fcollonval 

I successfully implemented opening a diff-ed file from the context menu. However, since the tab element of the diff window does not have a custom class, I cannot select only the diff tabs to add the command. I am unable to add a class to the diff tab also. Please let me know what I can do. Thank you!

fcollonval:
@BoscoCHW I looked for a solution, you can add a class to the diff widget title that will allow to set a more specific context menu selector.

You can add a class using `diffWidget.title.className = '...';` at

https://github.com/jupyterlab/jupyterlab-git/blob/708d19fc022825d9748f24ce7f94fe4839b738fc/src/commandsAndMenu.tsx#L521

BoscoCHW:
@fcollonval 
Hi, I successfully implemented a more specific context menu selector. The 'Open File' command only show up on diff tabs. Please let me know if there are any other changes to make on this PR.

src/commandsAndMenu.tsx
```diff
@@ -732,6 +733,31 @@ export function addCommands(
     icon: openIcon.bindprops({ stylesheet: 'menuItem' })
   });
 
+  commands.addCommand(ContextCommandIDs.openFileFromDiff, {
+    label: trans.__('Open File'),
+    execute: async _ => {
+      const widget = app.contextMenuHitTest((node: HTMLElement) => {
+        const nodeId = node.dataset.id;
+        return nodeId && nodeId.substring(0, 4) === 'diff';
```
fcollonval, line 741: Would you mind adding `git-` as prefix to `id` at line 499. I fear that testing for `diff` may not be sufficiently robust.

```suggestion
        return nodeId && nodeId.substring(0, 4) === 'git-diff';
```

src/commandsAndMenu.tsx
```diff
@@ -732,6 +733,31 @@ export function addCommands(
     icon: openIcon.bindprops({ stylesheet: 'menuItem' })
   });
 
+  commands.addCommand(ContextCommandIDs.openFileFromDiff, {
+    label: trans.__('Open File'),
+    execute: async _ => {
```
fcollonval, line 738: Let's add a caption bringing more information on the command purpose:

```suggestion
    caption: trans.__('Open file from its diff view'),
    execute: async _ => {
```

fcollonval (changes requested):
Thanks @BoscoCHW 

Two minor suggestions and we are good to go.

src/commandsAndMenu.tsx
BoscoCHW: @fcollonval 
But the first 4 characters of the nodeId is just `diff`, I don't think changing it to 'git-diff' would work.

fcollonval: @BoscoCHW you need indeed to add `git-` to line 499 where the id is defined for that suggestion to work (I can't make suggestion on unchanged lines :wink: )

BoscoCHW: @fcollonval That makes sense. I was just afraid it might break other codes if they also relied on parsing the NodeId to do other things.

fcollonval: It should not be the case - as this is something to avoid as much as possible. But thanks for raising the questions, better to discuss it than moving forward blindly.

BoscoCHW: @fcollonval Should I move forward with the suggestion then?

fcollonval: Yes please

PR #1146, opened 2022-07-15
Implement ui to support multiple remotes

BoscoCHW:
@fcollonval 

I implemented a dialogue to manage remote repositories.


jupyterlab_git/handlers.py
```diff
@@ -391,6 +391,39 @@ async def post(self, path: str = ""):
...
+class GitRemoteDetailsShowHandler(GitHandler):
+    """Handler for 'git remote -v'."""
+
+    @tornado.web.authenticated
+    async def get(self, path: str = ""):
+        """GET request handler to retrieve existing remotes."""
+        local_path = self.url2localpath(path)
+        output = await self.git.remote_show(local_path, verbose=True)
+        if output["code"] == 0:
+            self.set_status(201)
```
fcollonval, line 403: GET should return 200 (OK); 201 (CREATED) is adequate for POST request (see [wikipedia](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#2xx_success))

```suggestion
            self.set_status(200)
```

jupyterlab_git/handlers.py
```diff
@@ -391,6 +391,39 @@ async def post(self, path: str = ""):
...
+        local_path = self.url2localpath(path)
+        data = self.get_json_body()
+        name = data.get("name", None)
+
+        output = await self.git.remote_remove(local_path, name)
+        if output["code"] == 0:
+            self.set_status(201)
+        else:
+            self.set_status(500)
+        self.finish(json.dumps(output))
```
fcollonval, line 424: Let's use delete as this is more appropriate. This means you cannot pass a body but you should include the _name_ in the endpoint `/git/<path>/remote/<name>`.

The status code should be 204.

```suggestion


class GitRemoteRemoveHandler(GitHandler):
    """Handler for 'git remote remote <name>'."""
    
    @tornado.web.authenticated
    async def delete(self, path: str = "", name: str = ""):
        """DELETE request handler to remove a remote."""
        local_path = self.url2localpath(path)

        output = await self.git.remote_remove(local_path, name)
        if output["code"] == 0:
            self.set_status(204)
        else:
            self.set_status(500)
        self.finish(json.dumps(output))
```

jupyterlab_git/handlers.py
```diff
@@ -871,6 +904,8 @@ def setup_handlers(web_app):
         ("/push", GitPushHandler),
         ("/remote/add", GitRemoteAddHandler),
         ("/remote/fetch", GitFetchHandler),
+        ("/remote/show", GitRemoteDetailsShowHandler),
+        ("/remote/remove", GitRemoteRemoveHandler),
```
fcollonval, line 908: It will required a specific element to receive the name as parameter. So you should not place it in this list but add to `git_handlers` something like:

```py
git_handlers = # ... \
    + url_path_join(base_url, NAMESPACE + path_regex + r"/remote/%s" % r"(?P<name>[^/]*)" , GitRemoteRemoveHandler),
```

The handler argument is extracted from a named group in the regex.

src/components/AddRemoteDialogue.tsx
```diff
@@ -0,0 +1,191 @@
...
+                  newRemote: {
+                    ...this.state.newRemote,
+                    name: event.target.value
+                  }
+                })
+              }
+            />
+            <input
+              type="text"
+              placeholder="Remote Git repository URL"
```
fcollonval, line 122: That should be translated too.

```suggestion
              placeholder={this.props.trans.__('Remote Git repository URL')}
```

src/components/AddRemoteDialogue.tsx
```diff
@@ -0,0 +1,191 @@
...
+        <div className={contentWrapperClass}>
+          <label className={remoteDialogInputClass}>
+            <span>
+              {this.props.trans.__(
+                'Enter a new remote repository name and url'
+              )}
+            </span>
+            <input
+              type="text"
+              placeholder="name"
```
fcollonval, line 110: The placeholder will be displayed. So we should translate it too.

```suggestion
              placeholder={this.props.trans.__('name')}
```

src/components/AddRemoteDialogue.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+          {this.props.warningContent && (
+            <div className="jp-RemoteDialog-warning">
+              {this.props.warningContent}
+            </div>
+          )}
+
+          {this.state.existingRemotes.length > 0 && (
+            <ul className={existingRemoteListClass}>
+              <p>Existing Remotes:</p>
```
fcollonval, line 142:
```suggestion
              <p>{this.props.trans.__('Existing Remotes:')}</p>
```

fcollonval (changes requested):
Thanks for this one. This looks pretty good.

I added some suggestions. As discussed at the meeting, I let you propose something to display the dialog only if the branch has no remote or when triggered from the menu.

fcollonval (changes requested):
Thanks @BoscoCHW 

Nice work of getting a pure React and a widget dialogs within this PR.

On the code side, would you mind adding tests (in particular Python tests for the new handlers).

On the design side, 

What about renaming the menu entry `Manage Remote Repositories` instead of `Add Remote Repository`?
About the remote dialog:
image

What about moving the `Add` button below the input fields and remove the footer (with the two buttons)?
As it takes some time to retrieve the remote list, what about differentiating `state.existingRemotes === null` and `state.existingRemotes.length === 0`. So you can inform the user the remotes list is being retrieve when the state is null. And that there is no remote if the list is empty?

For the advanced push dialog, could you also inform the user that the remote list is being retrieved to reduce the popping of the list?

src/components/AddRemoteDialogue.tsx
```diff
@@ -0,0 +1,198 @@
...
+              className={classes(buttonClass, cancelButtonClass)}
+              type="button"
+              title={this.props.trans.__(
+                'Close this dialog without adding a remote repo'
+              )}
+              value={this.props.trans.__('Cancel')}
+              onClick={() => {
+                this.props.onClose();
+              }}
+            />
```
fcollonval, line 151: Let's remove the cancel button as it feels awkward.

```suggestion
```

src/commandsAndMenu.tsx
```diff
@@ -254,8 +255,8 @@ export function addCommands(
 
   /** Command to add a remote Git repository */
   commands.addCommand(CommandIDs.gitAddRemote, {
```
fcollonval, line 257: Let's refactor this a bit. Would you mind renaming the `gitAddRemote` command to be `gitManageRemote`.

We then should remove the ability to pass args to the `execute` method. Then in the dialog we should do the addition without closing the dialog. So the dialog is respecting its contract of management not of single action.

fcollonval (changes requested):
Thanks @BoscoCHW 

I propose to do  a bit more refactoring of the command. But every is there.

We can discuss this PR more tomorrow.

src/components/AddRemoteDialogue.tsx
BoscoCHW: Removed the cancel button

src/commandsAndMenu.tsx
BoscoCHW: Renamed the command to gitManageRemote and disabled arguments for the command. 
To-do:
Implement adding a new remote without closing the dialog

fcollonval (approved):
Thanks for clean-up the tests and adding new one @BoscoCHW 

I think is time to get this in :wink:

Issue #1109, opened 2022-04-16
Support branch visualization similar to `git log --graph`


joelostblom:
## Is your feature request related to a problem? Please describe.

When working with multiple branching, it can be quite helpful to visualize when they branched of from the trunk of the repo and also when they are merged back in. Particularly for teaching purposes it is valuable to visualize these relationships. From the command line it is possible to run `git log --oneline --all --graph` and display this information:

![image](https://user-images.githubusercontent.com/4560057/163680102-13d4db42-df0f-474e-9057-84011bbd8261.png)

Many IDEs include something similar, for example VSCode which is where the next screenshot is from

![image](https://user-images.githubusercontent.com/4560057/163680116-fe40c93e-80ff-4be1-8605-d6f05cd7725c.png)

## Describe the solution you'd like

I saw that this was included in the mock suggestions for improving the the history panel in https://github.com/jupyterlab/jupyterlab-git/issues/245, but it seems like it wasn't implemented in the PR that closed that issue (https://github.com/jupyterlab/jupyterlab-git/pull/270) (or I am missing where it is).

I like the mock pictures included there for how this could look in jupyterlab-git:

![image](https://user-images.githubusercontent.com/4560057/163680393-799b4061-2f91-46ce-a013-d2eb03d655c9.png)

## Describe alternatives you've considered

I currently use either command line or VScode for this purpose instead as a workaround solution. Not sure what an alternative in jupyterlab git would look like

## Additional context

- Python package version:
 ```
jupyterlab                3.3.4             
jupyterlab-git            0.36.0  
```
- Extension version:
```
JupyterLab v3.3.4
/home/joel/.local/share/jupyter/labextensions
        @jupyter-widgets/jupyterlab-manager v3.0.1 enabled OK (python, jupyterlab_widgets)

/home/joel/miniconda3/envs/jlgit/share/jupyter/labextensions
        jupyterlab_pygments v0.2.2 enabled OK (python, jupyterlab_pygments)
        nbdime-jupyterlab v2.1.1 enabled OK
        @jupyterlab/git v0.36.0 enabled OK (python, jupyterlab-git)

Other labextensions (built into JupyterLab)
   app dir: /home/joel/miniconda3/envs/jlgit/share/jupyter/lab
```
- Git version:
```
git version 2.32.0
```
- Operating System and its version:
```
Pop!OS 21.10
5.16.15-76051615-generic
```

fcollonval:
[...]

PR #1156, opened 2022-07-29
Git commit graph visualization on history sidebar

BoscoCHW:
Displayed commit graph on history sidebar that is responsive to commit history expansion

Fixes https://github.com/jupyterlab/jupyterlab-git/issues/1109

Issues: 
Codes are currently very messy. Will have to clean up code after the functionality is finished.
The graph became unresponsive to history expansion after changing repo / path. Somehow the useEffect function was not run after navigating to another repo, but if you switch to the "change" panel and switch back to the "history" panel, it works as expected.


jupyterlab_git/tests/test_single_file_log.py
```diff
@@ -42,37 +37,18 @@ async def test_single_file_log():
...
-                    "commit_msg": "add test.txt to root",
-                    "pre_commit": "",
+                    "author": "Lazy Junior Developer",
+                    "date": "5 hours ago",
+                    "commit_msg": "Something More",
+                    "pre_commits": ["263f762e0aad329c3c01bbd9a28f66403e6cfa5f"],
                     "is_binary": False,
-                    "file_path": "test.txt",
+                    "file_path": "folder/test.txt",
                 },
```
fcollonval, line 52: Would you mind adding a case were we don't have _precommits_ in the test?

src/components/GitCommitGraph.tsx
```diff
@@ -0,0 +1,220 @@
...
+  '#fef2c0',
+  '#bfe5bf',
+  '#c7def8',
+  '#bfdadc',
+  '#bfd4f2',
+  '#d4c5f9',
+  '#cccccc',
+  '#84b6eb',
+  '#e6e6e6',
+  '#ffffff',
```
fcollonval, line 29: We probably want to remove white
```suggestion
```

src/components/GitCommitGraph.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+  return maxBranch + 1;
+};
+
+export interface IGitCommitGraphProps {
+  commits: ICommit[];
+  getNodeHeight: (sha: string) => number;
+  dotRadius?: number;
+  lineWidth?: number;
+}
```
fcollonval, line 59: Would you mind adding some doc string describing those attributes?

src/components/HistorySideBar.tsx
```diff
@@ -112,81 +114,133 @@ export const HistorySideBar: React.FunctionComponent<IHistorySideBarProps> = (
...
+    }
+  });
+  React.useEffect(() => {
+    props.commits.forEach(commit =>
+      resizeObserver.observe(nodes.current[commit.commit], {
+        box: 'border-box'
+      })
+    );
+    return () => resizeObserver.disconnect();
+  }, [props.commits]);
```
fcollonval, line 140: This looks great :clap:

src/generateGraphData.ts
```diff
@@ -0,0 +1,133 @@
+export interface ICommit {
```
fcollonval, line 1: Would you mind documenting the interfaces?

src/generateGraphData.ts
```diff
@@ -0,0 +1,133 @@
...
+  sha,
+  dot: { lateralOffset: offset, branch },
+  routes,
+  yOffset
+});
+
+const remove = function (list: number[], item: number) {
+  list.splice(list.indexOf(item), 1);
+  return list;
+};
```
fcollonval, line 33: Let's name the function directly.

```suggestion
function remove(list: number[], item: number): number[] {
  list.splice(list.indexOf(item), 1);
  return list;
};
```

src/generateGraphData.ts
```diff
@@ -0,0 +1,133 @@
...
+
+/*
+  Generate preformatted data of commits graph.
+*/
+export const generateGraphData = function (
```
fcollonval, line 38:
```suggestion
export function generateGraphData(
```

src/generateGraphData.ts
```diff
@@ -0,0 +1,133 @@
...
+      [
+      sha,
+      [offset, branch], //dot
+      [
+      [from, to, branch],  // route 1
+      [from, to, branch],  // route 2
+      [from, to, branch],
+      ]  // routes
+      ],  // node
+  */
```
fcollonval, line 57: You could use that as docstring

src/generateGraphData.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  const nodes: INode[] = [];
+  const branchIndex = [0];
+  const reserve: number[] = [];
+  const branches: { [sha: string]: number } = {};
+
+  const getBranch = function (sha: string) {
```
fcollonval, line 64:
```suggestion
  function getBranch(sha: string) {
```

src/svgPathData.ts
```diff
@@ -0,0 +1,30 @@
...
+    cp1y: number,
+    cp2x: number,
+    cp2y: number,
+    x: number,
+    y: number
+  ): void {
+    this.commands.push(`C ${cp1x}, ${cp1y}, ${cp2x}, ${cp2y}, ${x}, ${y}`);
+  }
+
+  commands: string[];
```
fcollonval, line 29: We could rename this to be more consistent with its purpose and it should be private to avoid any tempering outside of the API:

```suggestion
  private _SVGPath: string[];
```

fcollonval (changes requested):
This is looking great.

I added a couple of suggestion and I noticed that the `package-lock.json` file slipped in. Would you mind removing it?

fcollonval (approved):
Big thanks @BoscoCHW and @iflinda for this one

Issue #1162, opened 2022-08-30
Prevent `git rev-parse --show-origin` from running in some dirs


divyansshhh:
## Is your feature request related to a problem? Please describe.

We have a contents manager that prefixes **some** paths for some special handling (these aren't actual file system paths). When navigating into these "prefixed dirs", jupyterlab-git tries to run `git rev-parse --show-origin` which causes it to fail and litter the console. 

Log - 
```
[20220830 11:19:29:378 git.py:158 WARNING] Fail to execute ['git', 'rev-parse', '--show-prefix']
[20220830 11:19:29:378 git.py:158 WARNING] Traceback (most recent call last):
[20220830 11:19:29:378 git.py:158 WARNING]   File "/usr/local/python/python-3.10/std/lib64/python3.10/site-packages/jupyterlab_git/git.py", line 144, in execute
[20220830 11:19:29:378 git.py:158 WARNING]     code, output, error = await current_loop.run_in_executor(
[20220830 11:19:29:378 git.py:158 WARNING]   File "/opt/python/python-3.10/lib64/python3.10/asyncio/futures.py", line 284, in __await__
[20220830 11:19:29:378 git.py:158 WARNING]     yield self  # This tells Task to wait for completion.
[20220830 11:19:29:378 git.py:158 WARNING]   File "/opt/python/python-3.10/lib64/python3.10/asyncio/tasks.py", line 304, in __wakeup
[20220830 11:19:29:378 git.py:158 WARNING]     future.result()
[20220830 11:19:29:378 git.py:158 WARNING]   File "/opt/python/python-3.10/lib64/python3.10/asyncio/futures.py", line 201, in result
[20220830 11:19:29:378 git.py:158 WARNING]     raise self._exception
[20220830 11:19:29:378 git.py:158 WARNING]   File "/opt/python/python-3.10/lib64/python3.10/concurrent/futures/thread.py", line 58, in run
[20220830 11:19:29:378 git.py:158 WARNING]     result = self.fn(*self.args, **self.kwargs)
[20220830 11:19:29:378 git.py:158 WARNING]   File "/usr/local/python/python-3.10/std/lib64/python3.10/site-packages/jupyterlab_git/git.py", line 110, in call_subprocess
[20220830 11:19:29:378 git.py:158 WARNING]     process = subprocess.Popen(
[...]

PR #1163, opened 2022-09-01
Add configuration to exclude paths

divyansshhh:
Resolves jupyterlab/jupyterlab-git#1162

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,7 @@ Once installed, extension behavior can be modified via the following settings wh
   It is possible to provide a list of commands to be executed in a folder after it is initialized as Git repository.
 - `JupyterLabGit.credential_helper`: Git credential helper to set to cache the credentials.
   The default value is `cache --timeout=3600` to cache the credentials for an hour. If you want to cache them for 10 hours, set `cache --timeout=36000`.
+- `JupyterLabGit.excluded_paths`: Set paths to exclude from this extension.
 
 <details>
 <summary><b>How to set server settings?</b></summary>
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ Once installed, extension behavior can be modified via the following settings wh
   It is possible to provide a list of commands to be executed in a folder after it is initialized as Git repository.
 - `JupyterLabGit.credential_helper`: Git credential helper to set to cache the credentials.
   The default value is `cache --timeout=3600` to cache the credentials for an hour. If you want to cache them for 10 hours, set `cache --timeout=36000`.
-- `JupyterLabGit.excluded_paths`: Set paths to exclude from this extension.
+- `JupyterLabGit.excluded_paths`: Set path patterns to exclude from this extension. You can use wildcard and interrogation mark for respectively everything or any single character in the pattern.
 
 <details>
 <summary><b>How to set server settings?</b></summary>
--- a/conftest.py
+++ b/conftest.py
@@ -7,5 +7,5 @@ pytest_plugins = ["jupyter_server.pytest_plugin"]
 def jp_server_config(jp_server_config, jp_root_dir):
     return {
         "ServerApp": {"jpserver_extensions": {"jupyterlab_git": True}},
-        "JupyterLabGit": {"excluded_paths": [f"/test-path/*"]},
+        "JupyterLabGit": {"excluded_paths": ["/ignored-path/*"]},
     }
--- a/conftest.py
+++ b/conftest.py
@@ -4,5 +4,8 @@ pytest_plugins = ["jupyter_server.pytest_plugin"]
 
 
 @pytest.fixture
-def jp_server_config(jp_server_config):
-    return {"ServerApp": {"jpserver_extensions": {"jupyterlab_git": True}}}
+def jp_server_config(jp_server_config, jp_root_dir):
+    return {
+        "ServerApp": {"jpserver_extensions": {"jupyterlab_git": True}},
+        "JupyterLabGit": {"excluded_paths": [f"/test-path/*"]},
+    }
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -1750,3 +1750,7 @@ class Git:
 
         elif self._GIT_CREDENTIAL_CACHE_DAEMON_PROCESS.poll():
             self.ensure_git_credential_cache_daemon(socket, debug, True, cwd, env)
+
+    @property
+    def excluded_paths(self):
+        return self._config.excluded_paths
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -1754,7 +1754,7 @@ class Git:
     @property
     def excluded_paths(self) -> List[str]:
         """Wildcard-style path patterns that do not support git commands.
-        
+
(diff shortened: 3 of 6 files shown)


divyansshhh:
@fcollonval , a few of the tests have failed and it seems because I'm accessing the settings through `self.git._config.excluded_paths`.

Do you have any other recommendations for this?

> The configuration will be accessible from self.git.config.excluded_paths

I tried this as well but this doesn't work because `self.git.config` is a function and also it returns the `git` config not `JupyterLabGit` config.

krassowski:
One option could be to expose `excluded_paths` as a property of `GitHandler`

```diff
diff --git a/jupyterlab_git/git.py b/jupyterlab_git/git.py
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -1750,3 +1750,7 @@ class Git:
 
         elif self._GIT_CREDENTIAL_CACHE_DAEMON_PROCESS.poll():
             self.ensure_git_credential_cache_daemon(socket, debug, True, cwd, env)
+
+    @property
+    def excluded_paths(self):
+        return self._config.excluded_paths
diff --git a/jupyterlab_git/handlers.py b/jupyterlab_git/handlers.py
--- a/jupyterlab_git/handlers.py
+++ b/jupyterlab_git/handlers.py
@@ -43,7 +43,7 @@ class GitHandler(APIHandler):
         super().prepare()
         path = self.path_kwargs.get("path")
         if path is not None:
-            excluded_paths = self.git._config.excluded_paths
+            excluded_paths = self.git.excluded_paths
             for excluded_path in excluded_paths:
                 if re.match(excluded_path, path):
                     self.set_status(404)
```

divyansshhh:
I was planning to write a test for this in `jupyterlab_git/test/test_handlers.py` but I can't figure out how I should pass the excluded paths setting.

Also, another question I have is - 
Consider - 
https://github.com/jupyterlab/jupyterlab-git/blob/a08a929188b3f05dbd169f4b71668c534c12e87c/jupyterlab_git/tests/test_handlers.py#L24

 Where do `jp_web_app` and `jp_root_dir` parameters come from?

fcollonval:
You can override the default server configuration by overriding the `jp_server_config` in the test file as done in https://github.com/jupyterlab/jupyterlab-git/blob/a08a929188b3f05dbd169f4b71668c534c12e87c/conftest.py#L7

The fixture requested are provided by the jupyter Server pytest Plugin (loaded in the file linked above).

This makes it quite convenient to create pytest for Jupyter server.

conftest.py
```diff
@@ -4,5 +4,8 @@
 
 
 @pytest.fixture
-def jp_server_config(jp_server_config):
-    return {"ServerApp": {"jpserver_extensions": {"jupyterlab_git": True}}}
+def jp_server_config(jp_server_config, jp_root_dir):
+    return {
+        "ServerApp": {"jpserver_extensions": {"jupyterlab_git": True}},
+        "JupyterLabGit": {"excluded_paths": [f"/test-path/*"]},
```
fcollonval, line 10: Unneeded f-string and could we use more precise rule so we avoid hard-to-find bug if somebody uses `test-path` in tests :wink: 

```suggestion
        "JupyterLabGit": {"excluded_paths": ["/ignored-path/*"]},
```

jupyterlab_git/git.py
```diff
@@ -1750,3 +1750,7 @@ def ensure_git_credential_cache_daemon(
 
         elif self._GIT_CREDENTIAL_CACHE_DAEMON_PROCESS.poll():
             self.ensure_git_credential_cache_daemon(socket, debug, True, cwd, env)
+
+    @property
+    def excluded_paths(self):
+        return self._config.excluded_paths
```
fcollonval, line 1756: Let's add some documentation:

```suggestion
    def excluded_paths(self) -> List[str]:
        """Wildcard-style path patterns that do not support git commands.
        
        You can use ``*`` to match everything or ``?`` to match any single character.
        """
        return self._config.excluded_paths
```

jupyterlab_git/handlers.py
```diff
@@ -37,6 +38,17 @@ class GitHandler(APIHandler):
...
+    def prepare(self):
+        """Check if the path should be skipped"""
+        super().prepare()
+        path = self.path_kwargs.get("path")
+        if path is not None:
+            excluded_paths = self.git.excluded_paths
+            for excluded_path in excluded_paths:
+                if re.match(excluded_path, path):
+                    self.set_status(404)
+                    self.finish()
```
fcollonval, line 50: We should use [fnmatchcase](https://docs.python.org/3/library/fnmatch.html) instead of regex as it is more commonly expected - as proven by your test `/test-path/*` in regex means `/test-path` followed by any number of `/`; not `/test-path/` followed by anything (as I suppose you meant).

```suggestion
            for excluded_path in excluded_paths:
                if fnmatch.fnmatchcase(path, excluded_path):
                    self.set_status(404)
                    self.finish()
                    break
```

jupyterlab_git/tests/test_handlers.py
```diff
@@ -106,6 +107,23 @@ async def test_git_show_prefix(mock_execute, jp_fetch, jp_root_dir):
     )
 
 
+async def test_git_show_prefix_for_excluded_path(
+    jp_fetch, jp_server_config, jp_root_dir
+):
+    local_path = jp_root_dir / "test-path"
```
fcollonval, line 113:
```suggestion
    local_path = jp_root_dir / "ignored-path"
```

README.md
```diff
@@ -112,6 +112,7 @@ Once installed, extension behavior can be modified via the following settings wh
   It is possible to provide a list of commands to be executed in a folder after it is initialized as Git repository.
 - `JupyterLabGit.credential_helper`: Git credential helper to set to cache the credentials.
   The default value is `cache --timeout=3600` to cache the credentials for an hour. If you want to cache them for 10 hours, set `cache --timeout=36000`.
+- `JupyterLabGit.excluded_paths`: Set paths to exclude from this extension.
```
fcollonval, line 115:
```suggestion
- `JupyterLabGit.excluded_paths`: Set path patterns to exclude from this extension. You can use wildcard and interrogation mark for respectively everything or any single character in the pattern.
```

jupyterlab_git/handlers.py
```diff
@@ -12,6 +12,7 @@
 from jupyter_server.services.contents.manager import ContentsManager
 from jupyter_server.utils import url2path, url_path_join
 from packaging.version import parse
+import re
```
fcollonval, line 15:
```suggestion
import fnmatch
```

fcollonval (changes requested):
Thanks @divyansshhh 

I made some suggestions to improve the code and add some more documentation.

conftest.py
divyansshhh: Ah, my bad, the f string was left over from something I was trying.

jupyterlab_git/handlers.py
divyansshhh: Got it, I was just looking into better ways of doing the pattern match.

fcollonval (approved):
Thanks a lot @divyansshhh

Issue #889, opened 2021-03-03
jupyterlab-git shouldn't show failure message when a user cancels an action


adammclaurin:
Right now jupyterlab-git will show a failure dialog if the user cancels a commit/push. It really shouldn't show anything at all in that case.

Damans227:
Hi @fcollonval I am Daman (The new fellow at MLH for RBC). I was trying to reproduce this issue, but from what I saw, the error message is already being displayed when you hit cancel in the dialog box asking for user credentials. Please look at this video I've recorded:

![testcase](https://user-images.githubusercontent.com/61474540/192659274-38faff71-497f-45c4-b42c-8b5e660388bb.gif)

Is there a different error message we are expecting. Lemme know what's the ask here.

Damans227:
> Hi @fcollonval I am Daman (The new fellow at MLH for RBC). I was trying to reproduce this issue, but from what I saw, the error message is already being displayed when you hit cancel in the dialog box asking for user credentials. Please look at this video I've recorded:
> 
> ![testcase](https://user-images.githubusercontent.com/61474540/192659274-38faff71-497f-45c4-b42c-8b5e660388bb.gif) [ ![testcase](https://user-images.githubusercontent.com/61474540/192659274-38faff71-497f-45c4-b42c-8b5e660388bb.gif) ](https://user-images.githubusercontent.com/61474540/192659274-38faff71-497f-45c4-b42c-8b5e660388bb.gif) [ ](https://user-images.githubusercontent.com/61474540/192659274-38faff71-497f-45c4-b42c-8b5e660388bb.gif)
> 
> Is there a different error message we are expecting. Lemme know what's the ask here.

@fcollonval you can ignore my last comment. I misread the issue description. 

Can you please assign this issue to me ? 

Thanks!

fcollonval:
Thanks @Damans227 for starting this.

PR #1170, opened 2022-09-28
Return without an error message when user cancels a push/pull

Damans227:
Provides a fix for https://github.com/jupyterlab/jupyterlab-git/issues/889

Push/Pull cancellation behaviour After Changes: 

fixforcancel

### DIFF
--- a/src/cancelledError.ts
+++ b/src/cancelledError.ts
@@ -1,0 +1,6 @@
+export class CancelledError extends Error {
+  constructor(...params: any) {
+    super(...params);
+    this.name = 'CancelledError';
+  }
+}
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -91,6 +91,9 @@ export namespace CommandArguments {
   }
 }
 
+const errorMessageOnCancel =
+  /fatal: could not read Username for 'https:\/\/github.com': terminal prompts disabled/;
+
 function pluralizedContextLabel(singular: string, plural: string) {
   return (args: any) => {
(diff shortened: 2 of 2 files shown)


Damans227:
> thanks @Damans227
> 
> Sorry my explanation yesterday was not clear.
> 
> As we want to be sure to react to cancel being pressed by the user, I was thinking of creating a new Error class and throw that in such case. Something like:
> 
> ```ts
> class CancelledError extends Error {
>   constructor(...paras) {
>     super(...params);
>     this.name = 'CancelledError';
>   }
> }
> 
> function showGitOperationDialog(...) {
>   ...
>   const result = await showDialog(...)
> 
>   if (result.button.accept) {
>     // ...
>   } else {
>     throw new CancelledError();
>   }
> }
> ...
> 
> try {
>   const r = showGitOperationDialog(...);
> } catch (error) {
>   if (error.name !== 'CancelledError') {
>     // Display error notification
>   }
>   
> }
> ```
> 
> Documentation: [MDN custom error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#custom_error_types)

Thanks for your feedback. I will try to get those changes in shortly.

Damans227:
Hi @fcollonval Could you please review the recent changes? I have manually tested this PR again after adding the `CancelledError` class, and everything seems to work.

fcollonval (approved):
This looks good!

Thanks @Damans227

fcollonval:
@all-contributors please add @Damans227 for code

Issue #856, opened 2021-01-18
Enable downloading repository without git metadata 


adammclaurin:
For some lesser technical Jupyter users, all they want to do is grab a notebook created by someone else and just run it. The problem is that running a notebook changes the notebook file, which means that the lesser technical person will need to understand how to discard changes in order to obtain an updated notebook from the creator. It would be nice to support a "download" functionality that copies the git repo without the metadata to reduce how much the lesser technical person needs to understand. This is analogous to how the GitHub web interface has a "Clone or download" button.

fcollonval:
Xref: https://github.com/jupyterlab/jupyterlab-git/issues/998

The two issues are trying to address a similar use case. In #998, it goes a bit further by keeping the git link in order to get the notification that the upstream did change.

The main question here is how/where to store a parameter that will tell the extension it is in downloading mode.

fcollonval:
A good solution is the one propose [there](https://stackoverflow.com/questions/11497457/git-clone-without-git-directory); aka

1. Cloning the repository with depth option equals to 1
2. Remove the `.git` repository

PR #1177, opened 2022-10-11
Support downloading a git repository snapshot

sheezaaziz:
Resolves https://github.com/jupyterlab/jupyterlab-git/issues/856

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -287,12 +287,16 @@ class Git:
                 await self.ensure_credential_helper(path)
             env["GIT_TERMINAL_PROMPT"] = "1"
             code, output, error = await execute(
-                ["git", "clone", unquote(repo_url), "-q"],
+                ["git", "clone", "--depth=1", unquote(repo_url), "-q"],
                 username=auth["username"],
                 password=auth["password"],
                 cwd=path,
                 env=env,
             )
+
+            # if cloned successfully
+            # rm -rf ./dirformynewrepo/.git
+
         else:
             env["GIT_TERMINAL_PROMPT"] = "0"
             code, output, error = await execute(
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -280,6 +280,7 @@ class Git:
         :param path: the directory where the clone will be performed.
         :param repo_url: the URL of the repository to be cloned.
         :param auth: OPTIONAL dictionary with 'username' and 'password' fields
+        :param versioning: OPTIONAL whether to clone or download a snapshot of the remote repository; default clone
         :return: response with status code and error message.
         """
         env = os.environ.copy()
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -272,7 +272,7 @@ class Git:
 
         return response
 
-    async def clone(self, path, repo_url, meta_data_checked=False, auth=None):
+    async def clone(self, path, repo_url, versioning=False, auth=None):
         """
         Execute `git clone`.
         When no auth is provided, disables prompts for the password to avoid the terminal hanging.
@@ -288,7 +288,7 @@ class Git:
                 await self.ensure_credential_helper(path)
             env["GIT_TERMINAL_PROMPT"] = "1"
 
-            if meta_data_checked:
+            if versioning:
                 current_content = set(os.listdir(path))
                 code, output, error = await execute(
                     ["git", "clone", "--depth=1", unquote(repo_url), "-q"],
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -272,7 +272,7 @@ class Git:
 
         return response
 
-    async def clone(self, path, repo_url, versioning=True, auth=None):
+    async def clone(self, path, repo_url, versioning=False, auth=None):
         """
         Execute `git clone`.
         When no auth is provided, disables prompts for the password to avoid the terminal hanging.
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -272,7 +272,7 @@ class Git:
 
         return response
 
-    async def clone(self, path, repo_url, not_versioning=False, auth=None):
+    async def clone(self, path, repo_url, versioning=True, auth=None):
         """
         Execute `git clone`.
         When no auth is provided, disables prompts for the password to avoid the terminal hanging.
@@ -284,7 +284,7 @@ class Git:
         """
         env = os.environ.copy()
         cmd = ["git", "clone"]
-        if not_versioning:
+        if not versioning:
             cmd.append("--depth=1")
             current_content = set(os.listdir(path))
         cmd.append(unquote(repo_url))
@@ -309,7 +309,7 @@ class Git:
                 env=env,
             )
 
-        if not_versioning:
+        if not versioning:
             new_content = set(os.listdir(path))
             directory = (new_content - current_content).pop()
             shutil.rmtree(f"{directory}/.git")
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -287,21 +287,28 @@ class Git:
             if auth.get("cache_credentials"):
                 await self.ensure_credential_helper(path)
             env["GIT_TERMINAL_PROMPT"] = "1"
-            current_content = set(os.listdir(path))
-            code, output, error = await execute(
-                ["git", "clone", "--depth=1", unquote(repo_url), "-q"],
-                username=auth["username"],
-                password=auth["password"],
-                cwd=path,
-                env=env,
-            )
 
             if meta_data_checked:
+                current_content = set(os.listdir(path))
+                code, output, error = await execute(
+                    ["git", "clone", "--depth=1", unquote(repo_url), "-q"],
+                    username=auth["username"],
+                    password=auth["password"],
+                    cwd=path,
+                    env=env,
+                )
                 new_content = set(os.listdir(path))
                 directory = (new_content - current_content).pop()
                 # Check that the directory contains a '.git' folder.
                 shutil.rmtree(f"{directory}/.git")
 
+            code, output, error = await execute(
+                ["git", "clone", unquote(repo_url), "-q"],
+                username=auth["username"],
+                password=auth["password"],
+                cwd=path,
+                env=env,
+            )
         else:
             env["GIT_TERMINAL_PROMPT"] = "0"
             code, output, error = await execute(
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -313,7 +313,7 @@ class Git:
         if not versioning:
             new_content = set(os.listdir(path))
             directory = (new_content - current_content).pop()
-            shutil.rmtree(f"{directory}/.git")
+            shutil.rmtree(f"{path}/{directory}/.git")
 
         response = {"code": code, "message": output.strip()}
 
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -272,7 +272,7 @@ class Git:
 
         return response
 
-    async def clone(self, path, repo_url, auth=None):
+    async def clone(self, path, repo_url, meta_data_checked=False, auth=None):
         """
         Execute `git clone`.
         When no auth is provided, disables prompts for the password to avoid the terminal hanging.
@@ -295,10 +295,12 @@ class Git:
                 cwd=path,
                 env=env,
             )
-            new_content = set(os.listdir(path))
-            directory = (new_content - current_content).pop()
-            # Check that the directory contains a '.git' folder.
-            shutil.rmtree(f"{directory}/.git")
+
+            if meta_data_checked:
+                new_content = set(os.listdir(path))
+                directory = (new_content - current_content).pop()
+                # Check that the directory contains a '.git' folder.
+                shutil.rmtree(f"{directory}/.git")
 
         else:
             env["GIT_TERMINAL_PROMPT"] = "0"
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -272,7 +272,7 @@ class Git:
 
         return response
 
-    async def clone(self, path, repo_url, versioning=False, auth=None):
+    async def clone(self, path, repo_url, auth=None, versioning=True):
         """
         Execute `git clone`.
         When no auth is provided, disables prompts for the password to avoid the terminal hanging.
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -283,27 +283,19 @@ class Git:
         :return: response with status code and error message.
         """
         env = os.environ.copy()
+        cmd = ["git", "clone"]
+        if versioning:
+            cmd.append("--depth=1")
+        cmd.append(unquote(repo_url))
+
         if auth:
             if auth.get("cache_credentials"):
                 await self.ensure_credential_helper(path)
             env["GIT_TERMINAL_PROMPT"] = "1"
-
-            if versioning:
-                current_content = set(os.listdir(path))
-                code, output, error = await execute(
-                    ["git", "clone", "--depth=1", unquote(repo_url), "-q"],
-                    username=auth["username"],
-                    password=auth["password"],
-                    cwd=path,
-                    env=env,
-                )
-                new_content = set(os.listdir(path))
-                directory = (new_content - current_content).pop()
-                # Check that the directory contains a '.git' folder.
-                shutil.rmtree(f"{directory}/.git")
-
+            cmd.append("-q")
+            current_content = set(os.listdir(path))
             code, output, error = await execute(
-                ["git", "clone", unquote(repo_url), "-q"],
+                cmd,
                 username=auth["username"],
                 password=auth["password"],
                 cwd=path,
@@ -312,10 +304,15 @@ class Git:
         else:
             env["GIT_TERMINAL_PROMPT"] = "0"
             code, output, error = await execute(
-                ["git", "clone", unquote(repo_url)],
+                cmd,
                 cwd=path,
                 env=env,
             )
+
+        if versioning:
+            new_content = set(os.listdir(path))
+            directory = (new_content - current_content).pop()
+            shutil.rmtree(f"{directory}/.git")
 
         response = {"code": code, "message": output.strip()}
 
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -285,9 +285,9 @@ class Git:
         env = os.environ.copy()
         cmd = ["git", "clone"]
         if versioning:
-            cmd.append("--depth=1")
             cmd.append("-q")
             current_content = set(os.listdir(path))
+            cmd.append("--depth=1")
         cmd.append(unquote(repo_url))
 
         if auth:
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -285,15 +285,15 @@ class Git:
         env = os.environ.copy()
         cmd = ["git", "clone"]
         if versioning:
-            cmd.append("-q")
-            current_content = set(os.listdir(path))
             cmd.append("--depth=1")
+            current_content = set(os.listdir(path))
         cmd.append(unquote(repo_url))
 
         if auth:
             if auth.get("cache_credentials"):
                 await self.ensure_credential_helper(path)
             env["GIT_TERMINAL_PROMPT"] = "1"
+            cmd.append("-q")
             code, output, error = await execute(
                 cmd,
                 username=auth["username"],
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -272,7 +272,7 @@ class Git:
 
         return response
 
-    async def clone(self, path, repo_url, versioning=False, auth=None):
+    async def clone(self, path, repo_url, not_versioning=False, auth=None):
         """
         Execute `git clone`.
         When no auth is provided, disables prompts for the password to avoid the terminal hanging.
@@ -284,7 +284,7 @@ class Git:
         """
         env = os.environ.copy()
         cmd = ["git", "clone"]
-        if versioning:
+        if not_versioning:
             cmd.append("--depth=1")
             current_content = set(os.listdir(path))
         cmd.append(unquote(repo_url))
@@ -309,7 +309,7 @@ class Git:
                 env=env,
             )
 
-        if versioning:
+        if not_versioning:
             new_content = set(os.listdir(path))
             directory = (new_content - current_content).pop()
             shutil.rmtree(f"{directory}/.git")
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -286,14 +286,14 @@ class Git:
         cmd = ["git", "clone"]
         if versioning:
             cmd.append("--depth=1")
+            cmd.append("-q")
+            current_content = set(os.listdir(path))
         cmd.append(unquote(repo_url))
 
         if auth:
             if auth.get("cache_credentials"):
                 await self.ensure_credential_helper(path)
             env["GIT_TERMINAL_PROMPT"] = "1"
-            cmd.append("-q")
-            current_content = set(os.listdir(path))
             code, output, error = await execute(
                 cmd,
                 username=auth["username"],
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -6,6 +6,7 @@ import os
 import pathlib
 import re
 import shlex
+import shutil
 import subprocess
 import traceback
 from typing import Dict, List, Optional
@@ -286,6 +287,7 @@ class Git:
             if auth.get("cache_credentials"):
                 await self.ensure_credential_helper(path)
             env["GIT_TERMINAL_PROMPT"] = "1"
+            current_content = set(os.listdir(path))
             code, output, error = await execute(
                 ["git", "clone", "--depth=1", unquote(repo_url), "-q"],
                 username=auth["username"],
@@ -293,9 +295,10 @@ class Git:
                 cwd=path,
                 env=env,
             )
-
-            # if cloned successfully
-            # rm -rf ./dirformynewrepo/.git
+            new_content = set(os.listdir(path))
+            directory = (new_content - current_content).pop()
+            # Check that the directory contains a '.git' folder.
+            shutil.rmtree(f"{directory}/.git")
 
         else:
             env["GIT_TERMINAL_PROMPT"] = "0"
--- a/jupyterlab_git/handlers.py
+++ b/jupyterlab_git/handlers.py
@@ -87,6 +87,7 @@ class GitCloneHandler(GitHandler):
         response = await self.git.clone(
             self.url2localpath(path),
             data["clone_url"],
+            data["metaDataChecked"],
             data.get("auth", None),
         )
 
--- a/jupyterlab_git/handlers.py
+++ b/jupyterlab_git/handlers.py
@@ -87,7 +87,7 @@ class GitCloneHandler(GitHandler):
         response = await self.git.clone(
             self.url2localpath(path),
             data["clone_url"],
-            data["metaDataChecked"],
+            data["versioning"],
             data.get("auth", None),
         )
 
--- a/jupyterlab_git/handlers.py
+++ b/jupyterlab_git/handlers.py
@@ -87,7 +87,7 @@ class GitCloneHandler(GitHandler):
         response = await self.git.clone(
             self.url2localpath(path),
             data["clone_url"],
-            data["versioning"],
+            data["not_versioning"],
             data.get("auth", None),
         )
 
--- a/jupyterlab_git/handlers.py
+++ b/jupyterlab_git/handlers.py
@@ -87,7 +87,7 @@ class GitCloneHandler(GitHandler):
         response = await self.git.clone(
             self.url2localpath(path),
             data["clone_url"],
-            data["not_versioning"],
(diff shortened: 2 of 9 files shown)


jupyterlab_git/git.py
```diff
@@ -287,12 +287,16 @@ async def clone(self, path, repo_url, auth=None):
...
-                ["git", "clone", unquote(repo_url), "-q"],
+                ["git", "clone", "--depth=1", unquote(repo_url), "-q"],
                 username=auth["username"],
                 password=auth["password"],
                 cwd=path,
                 env=env,
             )
+
+            # if cloned successfully
+            # rm -rf ./dirformynewrepo/.git
```
fcollonval, line 298: To remove a folder recursively in Python you can use `shutil.rmtree` (see the [doc](https://docs.python.org/3/library/shutil.html#shutil.rmtree)).

The though part will be to figure out the name of the folder. So I would propose to list the current path before and after the clone command. Something like:

```python
           current_content = set(os.listdir(path))
           # ... execute clone
           new_content = set(os.listdir(path))
           directory = (new_content - current_content).pop()
           # Check that the directory contains a '.git' folder.
           shutil.rmtree(f"{directory}/.git")
```

> Using `set` allows to quickly capture the content variation.

fcollonval (commented):
Thanks @sheezaaziz 

On the backend side, you will need to:

- Get a new flag to activate this feature from the request JSON body in the `GitCloneHandler.post` method
- Pass that flag to the `clone` method
- Modify or not the clone behavior depending on the flag value

On the frontend, the entry point for the user is the git clone command defines in `src/cloneCommand.ts`. It displays a dialog to the user. Then calls the model `clone` method (defined in `src/model.ts` L627). So the changes I imagine are:

- Extend `GitCloneForm` in `src/widgets/GitCloneForm.ts` to display a checkbox to the user
- Extend `GitCloneForm.getValue` to return a object {url, flag} with the repository url and the checkbox state instead of a single string.
- Pass the flag value to the model
- Add the flag value to the JSON body in `model.ts`

If you want to create a Python unit test, the best would be to duplicate `test_git_clone_success` in `jupyterlab_git/tests/test_clone.py`.

The new code will look like:

```py
@pytest.mark.asyncio
async def test_git_clone_success():
    with patch("os.environ", {"TEST": "test"}):
        with patch("jupyterlab_git.git.execute") as mock_execute:
            # Given
            output = "output"
            mock_execute.return_value = maybe_future((0, output, "error"))
            
            # Create a temporary directory and fake the content of it to be a directory containing a `.git` folder (that should be removed)
            # I advice you to use the pytest fixture tmp_path; see https://docs.pytest.org/en/6.2.x/reference.html#std-fixture-tmp_path

            # When
            actual_response = await Git().clone(
                path="",  # Here set the path to the temporary directory
                repo_url="ghjkhjkl",
                flag=True
            )

            # Then
            mock_execute.assert_called_once_with(
                ["git", "clone", "ghjkhjkl"], # to be update
                cwd=str(Path("/bin") / "test_curr_path"), # to be udpated
                env={"TEST": "test", "GIT_TERMINAL_PROMPT": "0"},
            )
            assert {"code": 0, "message": output} == actual_response
            
            # Check the `.git` folder has been removed.
```

jupyterlab_git/git.py
```diff
@@ -286,6 +287,21 @@ async def clone(self, path, repo_url, auth=None):
...
+                    password=auth["password"],
+                    cwd=path,
+                    env=env,
+                )
+                new_content = set(os.listdir(path))
+                directory = (new_content - current_content).pop()
+                # Check that the directory contains a '.git' folder.
+                shutil.rmtree(f"{directory}/.git")
+
             code, output, error = await execute(
```
fcollonval, line 305: This needs to be under a `else` condition that will be executed only if `meta_data_checked` is `False`. I guess this is the reason you are still having a .git folder as you clone twice the repo.

jupyterlab_git/git.py
```diff
@@ -271,7 +272,7 @@ async def changed_files(self, path, base=None, remote=None, single_commit=None):
 
         return response
 
-    async def clone(self, path, repo_url, auth=None):
+    async def clone(self, path, repo_url, meta_data_checked=False, auth=None):
```
fcollonval, line 275: Let's rename the variable to match better its purpose:

```suggestion
    async def clone(self, path, repo_url, versioning=False, auth=None):
```

jupyterlab_git/git.py
```diff
@@ -286,6 +287,21 @@ async def clone(self, path, repo_url, auth=None):
             if auth.get("cache_credentials"):
                 await self.ensure_credential_helper(path)
             env["GIT_TERMINAL_PROMPT"] = "1"
+
+            if meta_data_checked:
```
fcollonval, line 291: As there is already a condition on `auth` (aka are credentials provided or not?), I would do the test on the new flag first. And depending on its value, I would pick the correct command list.

Then after cloning (either with or without authentication), I will do a test against the flag again to know if the `.git` repository needs to be removed.

src/widgets/GitCloneForm.ts
```diff
@@ -16,23 +16,48 @@ export class GitCloneForm extends Widget {
...
+    linkText.textContent = trans.__('Enter the Clone URI of the repository');
+    inputLink.placeholder = 'https://host.com/org/repo.git';
+    checkboxText.textContent = 'Clone without metadata';
+    checkbox.setAttribute('type', 'checkbox');
+
+    inputLinkLabel.appendChild(linkText);
+    inputLinkLabel.appendChild(inputLink);
+
+    checkboxLabel.appendChild(checkboxText);
+    checkboxLabel.appendChild(checkbox);
```
fcollonval, line 56: I would advice you to invert the insertion order here.
Then for the look and feel, you probably needs to play with CSS styles.

You can look at the credential box that also uses a checkbox:
https://github.com/jupyterlab/jupyterlab-git/blob/master/style/credentials-box.css

src/cloneCommand.ts
```diff
@@ -46,7 +46,7 @@ export const gitCloneCommandPlugin: JupyterFrontEndPlugin<void> = {
           ]
         });
 
-        if (result.button.accept && result.value) {
+        if (result.button.accept && result.value.url) {
```
fcollonval, line 49: :+1:

src/commandsAndMenu.tsx
```diff
@@ -60,6 +60,10 @@ export interface IGitCloneArgs {
    * Git repository url
    */
   url: string;
+  /**
+   * Git metadata checked
+   */
+  metaDataChecked: boolean;
```
fcollonval, line 66: Let's rename this one `versioning` too

```suggestion
   * Whether to activate git versioning in the clone or not.
   * If false, this will remove the .git folder after cloning.
   */
  versioning: boolean;
```

fcollonval (commented):
The logic is looking good @sheezaaziz :clap:

src/widgets/GitCloneForm.ts
```diff
@@ -16,23 +16,54 @@ export class GitCloneForm extends Widget {
...
+    node.className = 'jp-CredentialsBox';
+    inputWrapper.className = 'jp-RedirectForm';
+    checkboxWrapper.className = 'jp-CredentialsBox-wrapper';
+    checkboxLabel.className = 'jp-CredentialsBox-label-checkbox';
+    checkbox.id = 'checkbox';
+    inputLink.id = 'input-link';
+
+    linkText.textContent = trans.__('Enter the Clone URI of the repository');
+    inputLink.placeholder = 'https://host.com/org/repo.git';
+    checkboxLabel.textContent = 'Clone without metadata';
```
fcollonval, line 53: The extension supports translation. For that you need to wrap the string to be translated with special `trans.__` function:

```suggestion
    checkboxLabel.textContent = trans.__('Clone without metadata');
```

fcollonval (commented):
This is looking good @sheezaaziz 

Do you need help to upgrade the tests?

jupyterlab_git/tests/test_clone.py
```diff
@@ -32,6 +32,41 @@ async def test_git_clone_success():
...
+
+            # Then
+            mock_execute.assert_called_once_with(
+                ["git", "clone", "ghjkhjkl"],  # to be update
+                cwd=str(Path("/bin") / "test_curr_path"),  # to be udpated
+                env={"TEST": "test", "GIT_TERMINAL_PROMPT": "0"},
+            )
+
+            # Check the `.git` folder has been removed.
+            os.path.isdir(str(Path("/git")))
```
sheezaaziz, line 65: should this be outputted in assert

jupyterlab_git/tests/test_clone.py
```diff
@@ -32,6 +32,41 @@ async def test_git_clone_success():
...
+        with patch("jupyterlab_git.git.execute") as mock_execute:
+            # Given
+            output = "output"
+            mock_execute.return_value = maybe_future((0, output, "error"))
+
+            # Create a temporary directory and fake the content of it to be a directory containing a `.git` folder (that should be removed)
+            # I advice you to use the pytest fixture tmp_path; see https://docs.pytest.org/en/6.2.x/reference.html#std-fixture-tmp_path
+            path = tmp_path_factory.mktemp("check-symlinks")
+            tmpdir = tempfile.mkdtemp()
+            temppath = tmp_path()
```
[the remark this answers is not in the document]
fcollonval: The easiest to use is `tmp_path` - that you can request directly as input argument (see comment above).

You will then need to create a `.git` folder with the temporary folder to check that it is removed.

jupyterlab_git/tests/test_clone.py
```diff
@@ -32,6 +32,41 @@ async def test_git_clone_success():
...
+
+            # When
+            actual_response = await Git().clone(
+                path=path,  # Here set the path to the temporary directory
```
fcollonval, line 52: To pass the temporary folder to the clone method, you will probably need to convert it to a string

```suggestion
                path=str(temppath),  # Here set the path to the temporary directory
```

jupyterlab_git/tests/test_clone.py
fcollonval, line 65: Yes something like

```py
assert (tmp_path / ".git").exists()
```

jupyterlab_git/tests/test_clone.py
```diff
@@ -32,6 +32,41 @@ async def test_git_clone_success():
...
+            mock_execute.assert_called_once_with(
+                ["git", "clone", "ghjkhjkl"],  # to be update
+                cwd=str(Path("/bin") / "test_curr_path"),  # to be udpated
+                env={"TEST": "test", "GIT_TERMINAL_PROMPT": "0"},
+            )
+
+            # Check the `.git` folder has been removed.
+            os.path.isdir(str(Path("/git")))
+
+            assert {"code": 0, "message": output} == actual_response
```
fcollonval, line 67: As discuss at the meeting, you can drop this test.

```suggestion
```

fcollonval (commented):
Thanks @sheezaaziz for looking at adding a test

jupyterlab_git/git.py
```diff
@@ -282,12 +283,19 @@ async def clone(self, path, repo_url, auth=None):
...
+        if versioning:
+            cmd.append("--depth=1")
+        cmd.append(unquote(repo_url))
+
         if auth:
             if auth.get("cache_credentials"):
                 await self.ensure_credential_helper(path)
             env["GIT_TERMINAL_PROMPT"] = "1"
+            cmd.append("-q")
+            current_content = set(os.listdir(path))
```
fcollonval, line 296: This should be called in the `if` condition L287

This should solves the remaining test failures.

fcollonval (changes requested):

jupyterlab_git/git.py
```diff
@@ -271,7 +272,7 @@ async def changed_files(self, path, base=None, remote=None, single_commit=None):
 
         return response
 
-    async def clone(self, path, repo_url, auth=None):
+    async def clone(self, path, repo_url, not_versioning=False, auth=None):
```
fcollonval, line 275:
```suggestion
    async def clone(self, path, repo_url, versioning=True, auth=None):
```

jupyterlab_git/git.py
```diff
@@ -282,12 +283,19 @@ async def clone(self, path, repo_url, auth=None):
         :return: response with status code and error message.
         """
         env = os.environ.copy()
+        cmd = ["git", "clone"]
+        if not_versioning:
```
fcollonval, line 287:
```suggestion
        if not versioning:
```

jupyterlab_git/git.py
```diff
@@ -296,11 +304,16 @@ async def clone(self, path, repo_url, auth=None):
         else:
             env["GIT_TERMINAL_PROMPT"] = "0"
             code, output, error = await execute(
-                ["git", "clone", unquote(repo_url)],
+                cmd,
                 cwd=path,
                 env=env,
             )
 
+        if not_versioning:
```
fcollonval, line 312:
```suggestion
        if not versioning:
```

jupyterlab_git/handlers.py
```diff
@@ -87,6 +87,7 @@ async def post(self, path: str = ""):
         response = await self.git.clone(
             self.url2localpath(path),
             data["clone_url"],
+            data["not_versioning"],
```
fcollonval, line 90:
```suggestion
            data["versioning"],
```

jupyterlab_git/tests/test_clone.py
```diff
@@ -32,6 +32,32 @@ async def test_git_clone_success():
...
+
+            # Then
+            mock_execute.assert_called_once_with(
+                ["git", "clone", "ghjkhjkl"],  # to be update
+                cwd=str(Path("/bin") / "test_curr_path"),  # to be udpated
+                env={"TEST": "test", "GIT_TERMINAL_PROMPT": "0"},
+            )
+
+            # Check the `.git` folder has been removed.
+            assert (str(tmp_path) / ".git").exists()
```
fcollonval, line 58:
```suggestion
            assert not (str(tmp_path) / ".git").exists()
```

jupyterlab_git/tests/test_clone.py
```diff
@@ -32,6 +32,32 @@ async def test_git_clone_success():
...
+async def test_git_clone_success(tmp_path):
+    with patch("os.environ", {"TEST": "test"}):
+        with patch("jupyterlab_git.git.execute") as mock_execute:
+            # Given
+            output = "output"
+            mock_execute.return_value = maybe_future((0, output, "error"))
+
+            # When
+            await Git().clone(
+                path=str(Path("/bin/test_curr_path")), repo_url="ghjkhjkl"
```
fcollonval, line 47:
```suggestion
                path=str(Path("/bin/test_curr_path")),
                repo_url="ghjkhjkl",
                versioning=False,
```

jupyterlab_git/git.py
```diff
@@ -282,12 +283,19 @@ async def clone(self, path, repo_url, auth=None):
         :return: response with status code and error message.
```
fcollonval, line 283:
```suggestion
        :param versioning: OPTIONAL whether to clone or download a snapshot of the remote repository; default clone
        :return: response with status code and error message.
```

jupyterlab_git/tests/test_clone.py
```diff
@@ -32,6 +32,32 @@ async def test_git_clone_success():
             assert {"code": 0, "message": output} == actual_response
 
 
+# Create a temporary directory and fake the content of it to be a directory containing a `.git` folder (that should be removed)
+@pytest.mark.asyncio
+@pytest.fixture
+async def test_git_clone_success(tmp_path):
```
fcollonval, line 38:
```suggestion
async def test_git_download_success(tmp_path):
```

src/cloneCommand.ts
```diff
@@ -56,7 +56,11 @@ export const gitCloneCommandPlugin: JupyterFrontEndPlugin<void> = {
               gitModel as GitExtension,
               Operation.Clone,
               trans,
-              { path: fileBrowserModel.path, url: result.value }
+              {
+                path: fileBrowserModel.path,
+                url: result.value.url,
+                not_versioning: result.value.not_versioning
```
fcollonval, line 62: This is a shortcut syntax that tells the compiler to copy all (key, value) from `result.value` into the new object:

```suggestion
                ...result.value
```

src/commandsAndMenu.tsx
```diff
@@ -60,6 +60,11 @@ export interface IGitCloneArgs {
    * Git repository url
    */
   url: string;
+  /**
+   * Whether to activate git not_versioning in the clone or not.
+   * If false, this will remove the .git folder after cloning.
+   */
+  not_versioning: boolean;
```
fcollonval, line 67:
```suggestion
   * Whether to activate git versioning in the clone or not.
   * If false, this will remove the .git folder after cloning.
   */
  versioning: boolean;
```

src/commandsAndMenu.tsx
```diff
@@ -1538,8 +1543,8 @@ export async function showGitOperationDialog<T>(
     switch (operation) {
       case Operation.Clone:
         // eslint-disable-next-line no-case-declarations
-        const { path, url } = args as any as IGitCloneArgs;
-        result = await model.clone(path, url, authentication);
+        const { path, url, not_versioning } = args as any as IGitCloneArgs;
+        result = await model.clone(path, url, not_versioning, authentication);
```
fcollonval, line 1547:
```suggestion
        const { path, url, versioning } = args as any as IGitCloneArgs;
        result = await model.clone(path, url, versioning, authentication);
```

src/model.ts
```diff
@@ -618,6 +618,7 @@ export class GitExtension implements IGitExtension {
    *
    * @param path - local path into which the repository will be cloned
    * @param url - Git repository URL
+   * @param not_versioning - boolean flag of Git metadata
```
fcollonval, line 621:
```suggestion
   * @param versioning - Whether to clone or download the Git repository
```

src/model.ts
```diff
@@ -627,6 +628,7 @@ export class GitExtension implements IGitExtension {
   async clone(
     path: string,
     url: string,
+    not_versioning: boolean,
```
fcollonval, line 631:
```suggestion
    versioning: boolean,
```

src/model.ts
```diff
@@ -637,6 +639,7 @@ export class GitExtension implements IGitExtension {
           'POST',
           {
             clone_url: url,
+            not_versioning: not_versioning,
```
fcollonval, line 642: This is a shortcut, when the key is identical to the variable name, you can just provide the variable:

```suggestion
            versioning,
```

src/tokens.ts
```diff
@@ -231,6 +231,7 @@ export interface IGitExtension extends IDisposable {
    *
    * @param path - local path into which the repository will be cloned
    * @param url - Git repository URL
+   * @param versioning - boolean flag of Git metadata
```
fcollonval, line 234:
```suggestion
   * @param versioning - Whether to clone or download the Git repository
```

src/tokens.ts
```diff
@@ -240,6 +241,7 @@ export interface IGitExtension extends IDisposable {
   clone(
     path: string,
     url: string,
+    not_versioning: boolean,
```
fcollonval, line 244:
```suggestion
    versioning: boolean,
```

src/widgets/GitCloneForm.ts
```diff
@@ -16,23 +16,54 @@ export class GitCloneForm extends Widget {
...
-  getValue(): string {
-    return encodeURIComponent(this.node.querySelector('input').value.trim());
+  getValue(): { url: string; not_versioning: boolean } {
+    return {
+      url: encodeURIComponent(
+        (
+          this.node.querySelector('#input-link') as HTMLInputElement
+        ).value.trim()
+      ),
+      not_versioning: Boolean(
```
fcollonval, line 26:
```suggestion
      versioning: !Boolean(
```

src/widgets/GitCloneForm.ts
```diff
@@ -16,23 +16,54 @@ export class GitCloneForm extends Widget {
...
+    node.className = 'jp-CredentialsBox';
+    inputWrapper.className = 'jp-RedirectForm';
+    checkboxWrapper.className = 'jp-CredentialsBox-wrapper';
+    checkboxLabel.className = 'jp-CredentialsBox-label-checkbox';
+    checkbox.id = 'checkbox';
+    inputLink.id = 'input-link';
+
+    linkText.textContent = trans.__('Enter the Clone URI of the repository');
+    inputLink.placeholder = 'https://host.com/org/repo.git';
+    checkboxLabel.textContent = trans.__('Clone without metadata');
```
fcollonval, line 53:
```suggestion
    checkboxLabel.textContent = trans.__('Download the repository');
    checkboxLabel.title = trans.__('If checked, the remote repository default branch will be downloaded instead of cloned');
```

src/widgets/GitCloneForm.ts
```diff
@@ -16,23 +16,54 @@ export class GitCloneForm extends Widget {
...
-    text.textContent = trans.__('Enter the Clone URI of the repository');
-    input.placeholder = 'https://host.com/org/repo.git';
+    node.className = 'jp-CredentialsBox';
+    inputWrapper.className = 'jp-RedirectForm';
+    checkboxWrapper.className = 'jp-CredentialsBox-wrapper';
+    checkboxLabel.className = 'jp-CredentialsBox-label-checkbox';
+    checkbox.id = 'checkbox';
+    inputLink.id = 'input-link';
+
+    linkText.textContent = trans.__('Enter the Clone URI of the repository');
```
fcollonval, line 51:
```suggestion
    linkText.textContent = trans.__('Enter the URI of the remote Git repository');
```

jupyterlab_git/tests/test_clone.py
```diff
@@ -32,6 +32,32 @@ async def test_git_clone_success():
             assert {"code": 0, "message": output} == actual_response
 
 
+# Create a temporary directory and fake the content of it to be a directory containing a `.git` folder (that should be removed)
+@pytest.mark.asyncio
+@pytest.fixture
```
fcollonval, line 37: This decorator is to mention a function is a test fixture. If you want to use a test fixture, you only need to request them as input argument of the test function

```suggestion
```

> The consequence is that your test was actually not executed. Otherwise you would have got some name clash with the previous test.

fcollonval (changes requested):
Thanks @sheezaaziz 

I switch back to `versioning` correcting its value because it makes the code easier to read. I also modify the comments to be more consistent. And finally I fixed the test, because actually we were not testing the case of downloading the remote repository but of cloning it; we were checking that `.git` is existing.

jupyterlab_git/tests/test_clone.py
```diff
@@ -32,6 +32,30 @@ async def test_git_clone_success():
...
 
 
+@pytest.mark.asyncio
+async def test_git_clone_success(tmp_path):
+    with patch("os.environ", {"TEST": "test"}):
+        with patch("jupyterlab_git.git.execute") as mock_execute:
+            # Given
+            output = "output"
+            mock_execute.return_value = maybe_future((0, output, "error"))
+
```
fcollonval, line 42:
```suggestion

            dummy_repo = tmp_path / "test-repo" / ".git"
            dummy_repo.mkdir(parents=True)            
```

src/model.ts
```diff
@@ -627,6 +628,7 @@ export class GitExtension implements IGitExtension {
   async clone(
     path: string,
     url: string,
+    versioning: boolean,
     auth?: Git.IAuth
```
fcollonval, line 632:
```suggestion
    auth?: Git.IAuth,
    versioning = true
```

src/model.ts
```diff
@@ -618,6 +618,7 @@ export class GitExtension implements IGitExtension {
    *
    * @param path - local path into which the repository will be cloned
    * @param url - Git repository URL
+   * @param versioning - boolean flag of Git metadata
    * @param auth - remote repository authentication information
```
fcollonval, line 622:
```suggestion
   * @param auth - remote repository authentication information
   * @param versioning - boolean flag of Git metadata (default true)
```

src/commandsAndMenu.tsx
```diff
@@ -1538,8 +1543,8 @@ export async function showGitOperationDialog<T>(
     switch (operation) {
       case Operation.Clone:
         // eslint-disable-next-line no-case-declarations
-        const { path, url } = args as any as IGitCloneArgs;
-        result = await model.clone(path, url, authentication);
+        const { path, url, versioning } = args as any as IGitCloneArgs;
+        result = await model.clone(path, url, versioning, authentication);
```
fcollonval, line 1547:
```suggestion
        result = await model.clone(path, url, authentication, versioning);
```

src/commandsAndMenu.tsx
```diff
@@ -60,6 +60,11 @@ export interface IGitCloneArgs {
    * Git repository url
    */
   url: string;
+  /**
+   * Whether to activate git versioning in the clone or not.
+   * If false, this will remove the .git folder after cloning.
+   */
+  versioning: boolean;
```
fcollonval, line 67:
```suggestion
  versioning?: boolean;
```

src/tokens.ts
```diff
@@ -231,6 +231,7 @@ export interface IGitExtension extends IDisposable {
    *
    * @param path - local path into which the repository will be cloned
    * @param url - Git repository URL
+   * @param versioning - boolean flag of Git metadata
    * @param auth - remote repository authentication information
```
fcollonval, line 235:
```suggestion
   * @param auth - remote repository authentication information
   * @param versioning - Whether to clone or download the Git repository
```

src/tokens.ts
```diff
@@ -240,6 +241,7 @@ export interface IGitExtension extends IDisposable {
   clone(
     path: string,
     url: string,
+    versioning: boolean,
     auth?: Git.IAuth
```
fcollonval, line 245:
```suggestion
    auth?: Git.IAuth,
    versioning?: boolean,
```

src/commandsAndMenu.tsx
```diff
@@ -1538,8 +1543,8 @@ export async function showGitOperationDialog<T>(
     switch (operation) {
       case Operation.Clone:
         // eslint-disable-next-line no-case-declarations
-        const { path, url } = args as any as IGitCloneArgs;
-        result = await model.clone(path, url, authentication);
+        const { path, url, versioning } = args as any as IGitCloneArgs;
+        result = await model.clone(path, url, authentication, versioning ?? true);
```
fcollonval, line 1547:
```suggestion
        result = await model.clone(
          path,
          url,
          authentication,
          versioning ?? true
        );
```

src/tokens.ts
```diff
@@ -240,7 +241,8 @@ export interface IGitExtension extends IDisposable {
   clone(
     path: string,
     url: string,
-    auth?: Git.IAuth
+    auth?: Git.IAuth,
+    versioning?: boolean,
```
fcollonval, line 245:
```suggestion
    versioning?: boolean
```

jupyterlab_git/tests/test_clone.py
```diff
@@ -32,6 +32,30 @@ async def test_git_clone_success():
...
+            # Then
+            mock_execute.assert_called_once_with(
+                ["git", "clone", "ghjkhjkl"],  # to be update
+                cwd=str(Path("/bin") / "test_curr_path"),  # to be udpated
+                env={"TEST": "test", "GIT_TERMINAL_PROMPT": "0"},
+            )
+
+            # Check the `.git` folder has been removed.
+            assert not (str(tmp_path) / ".git").exists()
```
fcollonval, line 56: @sheezaaziz I forgot that we need to create the folder `repo/.git` when `execute` is called. So this test is gonna be a bit more advance as we need to use the `side_effect` feature of `patch` instead of `return_value` (see [documentation](https://docs.python.org/3/library/unittest.mock.html?highlight=patch#unittest.mock.Mock.side_effect))

Something like that

```suggestion
            git_folder = tmp_path / "dummy-repo" / ".git"
            def create_fake_git_repo(*args, **kwargs):
                git_folder.mkdir(parents=True)
                return maybe_future((0, output, "error"))
            mock_execute.side_effect = create_fake_git_repo

            # When
            await Git().clone(
                path=str(tmp_path), repo_url="ghjkhjkl", None, False)
            )

            # Then
            mock_execute.assert_called_once_with(
                ["git", "clone", "--depth=1", "ghjkhjkl"],  # to be update
                cwd=str(tmp_path),  # to be udpated
                env={"TEST": "test", "GIT_TERMINAL_PROMPT": "0"},
            )

            # Check the repository folder has been created
            assert git_folder.parent.exists()
            # Check the `.git` folder has been removed.
            assert not git_folder.exists()
```

jupyterlab_git/handlers.py
```diff
@@ -87,6 +87,7 @@ async def post(self, path: str = ""):
         response = await self.git.clone(
             self.url2localpath(path),
             data["clone_url"],
+            data["versioning"],
             data.get("auth", None),
```
fcollonval, line 91:
```suggestion
            data.get("auth", None),
            data["versioning"],
```

jupyterlab_git/git.py
```diff
@@ -271,23 +272,31 @@ async def changed_files(self, path, base=None, remote=None, single_commit=None):
 
         return response
 
-    async def clone(self, path, repo_url, auth=None):
+    async def clone(self, path, repo_url, versioning=False, auth=None):
```
fcollonval, line 275:
```suggestion
    async def clone(self, path, repo_url, auth=None, versioning=True):
```

fcollonval (approved):
Thanks a lot @sheezaaziz Let's merge this one

Issue #664, opened 2020-06-02
Allow diffs of images


ianhi:
Currently standalone diffs of images are not supported. However they are supported inside of notebook diffs. I propose having the standalone present diffs of images the same way they are handled by nbdime: showing the old version and new version side by side:

**notebook diff:**
![image](https://user-images.githubusercontent.com/10111092/83477061-4e8e8480-a460-11ea-8d4c-a93726e5d221.png)
**vscode:**
![image](https://user-images.githubusercontent.com/10111092/83477210-c197fb00-a460-11ea-8f12-5985b01072bf.png)
**Github desktop:**
Provides several options beyond side by side
![image](https://user-images.githubusercontent.com/10111092/83477195-bb098380-a460-11ea-967d-c986c2d1bf6b.png)

## Approach
I think that this can be accomplished by using `git show HEAD:./img.png` to get the image as bytes. At least I found that  `git show HEAD:./img.png > test.png` produced a working image that was indeed from the previous commit. So the flow would be:
- Request diff of image
    - One of the handlers realizes this is an image and swaps request to be `git show`
- Return bytes to frontend
- convert to blob
- generate `ObjectUrl` for use in an `img` tag

fcollonval:
Nice idea.

One comment:

You should be able to display it more simply by injecting the bytes as base64 encoded string. See for example the image viewer in JupyterLab:
https://github.com/jupyterlab/jupyterlab/blob/e7317f21b48183423e375dc266476a4289eb0fa3/packages/imageviewer/src/widget.ts#L175

> It may be worth to reuse that widget actually

ianhi:
Typescript library with a react component for doing this: https://octoframes.github.io/compare_view/

Which is used in this jupyter magic: https://github.com/Octoframes/jupyter_compare_view#jupyter-compare_view

fcollonval:
Closing as resolved by #1223

PR #1179, opened 2022-10-27
Adding support for standalone diffs of images 

Damans227:
This PR intends to resolve: https://github.com/jupyterlab/jupyterlab-git/issues/664 

*Note: For now, this is just a draft pr raised for collaboration. Therefore, it's not ready to be merged yet.*

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -26,6 +26,7 @@ import { Git, IGitExtension } from './tokens';
 import { addCloneButton } from './widgets/gitClone';
 import { GitWidget } from './widgets/GitWidget';
 import { gitCloneCommandPlugin } from './cloneCommand';
+import { createImageDiff } from './components/diff/ImageDiff';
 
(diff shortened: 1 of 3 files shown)


Damans227 (commented):
@fcollonval Hi Frederic, What should be the name of the image diff that gets passed here? Is this value arbitrary?

src/index.ts
```diff
@@ -218,5 +219,12 @@ async function activate(
       createNotebookDiff(model, renderMime, toolbar, translator)
   );
 
+  gitExtension.registerDiffProvider(
+    'ImageDiff',
```
Damans227, line 223: @fcollonval What should be the name of the image diff that gets passed here? Is this value arbitrary?

fcollonval: Yes

fcollonval:
Closed by #1223

Issue #620, opened 2020-04-24
add a setting for only using a single tab for diffs


ianhi:
## Description
Currently every diff is opened in a new jupyter tab. I propose a setting that would enable behavior similar to github desktop where there is only one tab for diffs, and opening a new diff would change what diff is displayed but not create a new tab.
![single_tab](https://user-images.githubusercontent.com/10111092/80234557-5d467800-8626-11ea-872b-a99b9258ea8a.gif)

ianhi:
Contrast with the tab explosion when viewing the same commits in jupyterlab-git:
![tab_explosion](https://user-images.githubusercontent.com/10111092/80235007-1016d600-8627-11ea-8f05-6418859a9878.gif)

Often this multitab behavior is pretty nice, but I'd like to be able to toggle it on and off.

fcollonval:
That is a nice idea. But a better approach instead of all or none would be to set a maximum number of diff tab to be kept opened using a LIFO (Last In First Out) queue. If the setting is 0, then the behavior would be the actual. And if it is one, it will be the one you propose. And if it is something else, it would probably make more sense like looking at all changes in a commit (especially as diffing big file can be quite slow) but keeping the number of opened tabs _acceptable_.

jaipreet-s:
VSCode  shows each file diff as a separate tab as well.
![Screen Shot 2020-04-27 at 11 22 08 AM](https://user-images.githubusercontent.com/43826141/80406714-5a8f9100-8879-11ea-8c81-8fc928e464be.png)


I like this idea as a  setting for the opinionated user, but its important to not over-index on GitHub Desktop or a particular Git GUI.

dhirschfeld:
In VSCode when you click on a file it shows the diff in a separate diff tab. If you click on another file it shows the diff in the *same* diff tab. A *new* diff tab is only opened if you *double-click* on a file (or if no diff tab is already open).

I would love for `jupyterlab-git` to replicate this UX as I think it's pretty perfect:
https://github.com/jupyterlab/jupyterlab-git/issues/252#issuecomment-509865304

telamonian:
@dhirschfeld As you said in the other issue, what you're describing is part of a more complex feature in vscode called (preview mode)[https://code.visualstudio.com/docs/getstarted/userinterface#_preview-mode]. Implementation of our own preview mode would probably best be first tackled in core before we implement our own flavor of it here in the git extension.

Some thoughts:
[...]

PR #1184, opened 2022-11-15
Use preview tab for diff file views

sheezaaziz:
Fixes https://github.com/jupyterlab/jupyterlab-git/issues/620

### DIFF
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -15,11 +15,11 @@ import { ISettingRegistry } from '@jupyterlab/settingregistry';
 import { ITerminal } from '@jupyterlab/terminal';
 import { ITranslator, TranslationBundle } from '@jupyterlab/translation';
 import { closeIcon, ContextMenuSvg } from '@jupyterlab/ui-components';
-import { ArrayExt, toArray } from '@lumino/algorithm';
+import { ArrayExt, toArray, find } from '@lumino/algorithm';
 import { CommandRegistry } from '@lumino/commands';
 import { PromiseDelegate } from '@lumino/coreutils';
 import { Message } from '@lumino/messaging';
-import { ContextMenu, Menu, Panel, Title, Widget } from '@lumino/widgets';
+import { ContextMenu, DockPanel, Menu, Panel, Widget } from '@lumino/widgets';
 import * as React from 'react';
 import { DiffModel } from './components/diff/model';
 import { createPlainTextDiff } from './components/diff/PlainTextDiff';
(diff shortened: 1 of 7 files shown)


src/commandsAndMenu.tsx
```diff
@@ -558,6 +562,34 @@ export function addCommands(
...
+                : null;
+            const tabTitle = tab.querySelector<HTMLElement>(
+              '.lm-TabBar-tabLabel'
+            );
+            tabTitle.style.fontStyle = 'italic';
+            tabTitle.addEventListener('mouseup', () => {
+              tabTitle.style.fontStyle = 'normal';
+              if (PreviewMainAreaWidget.previewWidget === diffWidget) {
+                PreviewMainAreaWidget.previewWidget = null;
+              }
```
fcollonval, line 589: Could you create a static method in `PreviewMainAreaWidget` like `pin(widget: Widget)` that will do that?

src/components/diff/PreviewMainAreaWidget.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Message } from '@lumino/messaging';
+import { Widget } from '@lumino/widgets';
+
+export class PreviewMainAreaWidget<
+  T extends Widget = Widget
+> extends MainAreaWidget {
+  /**
+   * Handle on the preview widget
+   */
+  static previewWidget: PreviewMainAreaWidget | null = null;
```
fcollonval, line 11: Let's make this protected

```suggestion
  protected static previewWidget: PreviewMainAreaWidget | null = null;
```

src/commandsAndMenu.tsx
```diff
@@ -558,6 +562,34 @@ export function addCommands(
...
+
+          let tabPosition = -1;
+          const tabBar = find(dockPanel.tabBars(), bar => {
+            tabPosition = bar.titles.indexOf(diffWidget.title);
+            console.log('2', tabPosition);
+            return tabPosition !== -1;
+          });
+
+          // Get the tab
+          setTimeout(() => {
```
fcollonval, line 576: Let's add some explanation about the need for `setTimeout`

```suggestion
          // We need to wait for the tab node to be inserted in the DOM
          setTimeout(() => {
```

fcollonval (changes requested):
Thanks @sheezaaziz I have a few suggestions.

style/diff-nb.css
```diff
@@ -143,6 +143,10 @@
 	border: none;
 }
 
+.jp-git-tab-mod-preview {
+  font-style: italic;
+}
+
```
fcollonval, line 149: Would you mind placing this block at the end of `style/base.css`.

src/components/diff/PreviewMainAreaWidget.ts
```diff
@@ -0,0 +1,78 @@
...
+        tabPosition >= 0 ? tabBar.contentNode.children[tabPosition] : null;
+      const tabTitle = tab.querySelector<HTMLElement>('.lm-TabBar-tabLabel');
+
+      tabTitle.classList.add('jp-git-tab-mod-preview');
+      tabTitle.onclick = () => {
+        tabTitle.classList.remove('jp-git-tab-mod-preview');
+        if (PreviewMainAreaWidget.previewWidget === diffWidget) {
+          PreviewMainAreaWidget.previewWidget = null;
+        }
+      };
```
fcollonval, line 52: Almost correct, we also want to remove the click listener once it has been triggered or when the diff view is disposed:

```suggestion
      const onClick = () => {
        tabTitle.classList.remove('jp-git-tab-mod-preview');
        tabTitle.removeEventListener('click', onClick, true);
        if (PreviewMainAreaWidget.previewWidget === diffWidget) {
          PreviewMainAreaWidget.previewWidget = null;
        }
      };
      
      tabTitle.addEventListener('click', onClick, true);
      diffWidget.disposed.connect(() => {
        tabTitle.removeEventListener('click', onClick, true);
      });
```

fcollonval (changes requested):
Thank @sheezaaziz 

Two small things and we are good to go.

fcollonval (commented):
Thanks a lot @sheezaaziz 

I tested and look again at the code. This is great. I rerun the tests failing hoping it is a random error.

fcollonval (approved):

fcollonval:
https://github.com/all-contributors please add @sheezaaziz for code

fcollonval:
@all-contributors please add @sheezaaziz for code

krassowski:
Great work here, it is a nice feature to have!
- Could we add a setting to disable this behaviour? It seems that https://github.com/jupyterlab/jupyterlab-git/issues/620 proposed adding this as an opt-in setting.
- When I move the diff to another spit panel (say to the right):
   - the italic style in tab bar disappears (probably the class does not stick after reattaching in another panel)
   - if I open another diff the old diff in right panel gets closed and reopened in the left panel. I find this annoying, I guess it was not intended.
- I had no idea what italic style means before looking at the docs; it would be helpful if there was an explanation in the title, or maybe even a pin icon; a pin icon could be implemented as unicode prefix (📌) added to title

PR #1188, opened 2022-11-25
Add submodules clone checkbox

### LINKED ISSUE #1187: Clone git submodules
Feature request
----------------
At the moment, `jupyterlab-git` plugin is able to clone git repository in a JupyterLab environment in a very convenient way.
Unfortunately for git repo that have [submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules) only the parent is cloned and you get empty folders on the submodules.

Would it possible to add the `--recurse-submodules` flag by default when the project is cloned:
```
git clone --recurse-submodules my_repo_url
```
(or to add an optional recursive check box on the clone window)

Current workaround
--------------------
It is possible to pull the submodules after the clone with this command:
```
git submodule update --init --recursive
```
But it requires an additional step from the user (and do it at the right location…)

seignovert:
Resolves #1187

- Add an optional `submodules` bolean argument to `Git.clone()` to append `--recurse-submodules` flag (default `False`)
- Add `test_git_submodules_success` test
- Add `data["submodules"]` to `GitCloneHandler`
- Add a new submodules checkbox in `GitCloneForm` (checked by default)
- Renamed `checkbox` input to `download` in `GitCloneForm`

The clone dialog looks like this:

image

---
Edited to add a new clone dialog image

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -272,7 +272,7 @@ class Git:
 
         return response
 
-    async def clone(self, path, repo_url, auth=None, versioning=True):
+    async def clone(self, path, repo_url, auth=None, versioning=True, submodules=False):
         """
         Execute `git clone`.
         When no auth is provided, disables prompts for the password to avoid the terminal hanging.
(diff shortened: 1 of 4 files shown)


fcollonval (approved):
Thanks a lot @seignovert 

This is awesome with doc, test and new UI :rocket:

PR #1191, opened 2022-11-28
Fix submodules unhandler

seignovert:
Unfortunately it seems that #1188 introduced a `Unhandled error` 😞

```
Uncaught exception POST /git/clone?1669638844222 (127.0.0.1)
    HTTPServerRequest(protocol='http', host='localhost:8888', method='POST', uri='/git/clone?1669638844222', version='HTTP/1.1', remote_ip='127.0.0.1')
    Traceback (most recent call last):
      File ".../site-packages/tornado/web.py", line 1713, in _execute
        result = await result
      File ".../site-packages/jupyterlab_git/handlers.py", line 92, in post
        data["submodules"],
    KeyError: 'submodules'
```

This PR should implements a hot fix with the correct clone command handler (I have mainly duplicated the `versioning` attribute in the TS command model).

You might also need to write a test for the clone TS component (I don't really know how to do it).

Feel free to double check and edit this PR if needed. My apologies for not spotting this earlier.

### DIFF
--- a/src/cloneCommand.ts
+++ b/src/cloneCommand.ts
@@ -59,7 +59,8 @@ export const gitCloneCommandPlugin: JupyterFrontEndPlugin<void> = {
               {
                 path: fileBrowserModel.path,
                 url: result.value.url,
-                versioning: result.value.versioning
+                versioning: result.value.versioning,
+                submodules: result.value.submodules
               }
             );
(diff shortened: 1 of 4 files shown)


fcollonval (approved):
Thanks for quickly fixing this

PR #1196, opened 2022-12-05
Update commandsAndMenu.tsx to replace "`" in file path for Open Git Repository since it leads to Command Injection Vulnerability

rahrad123:
When a repo is created with the backtick character around it and Initialized as a repo and then opened in Terminal, the linux command is resolved or executed on a running instance. For example if a folder with the name "whoami" is created, initialized as a repo and then opened in terminal using 'Open Git repository in Terminal' you will see that whoami is resolved to the current user which is a vector of command injection.

### DIFF
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -160,7 +160,11 @@ export function addCommands(
           terminal.session.send({
             type: 'stdin',
             content: [
-              `cd "${gitModel.pathRepository.split('"').join('\\"')}"\n`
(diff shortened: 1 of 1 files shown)


fcollonval (approved):
Thanks @rahrad123

rahrad123:
@fcollonval AWS Sagemaker currently uses version 0.11.0 and 0.24.0 for JupyterLab 1 version currently. Could you merge these changes to the jlab-2 branch and the 0.11.x branch please?

PR #1200, opened 2022-12-08
.

sheezaaziz:


jupyterlab_git/git.py
```diff
@@ -222,6 +222,10 @@ async def config(self, path, **kwargs):
 
         return response
 
+    def check_ignore(self, path):
+        # in git.py Git, add a new method check_ignore that execute git check-ignore -z <path> and parse the result to return the list of ignored files; changed_files method is a good example of a similar method.
+        return
```
fcollonval, line 227: What you would like to do is to execute the git `check-ignore` command in the repository to know if it is ignored or not. Something like this should do:

```suggestion
    def check_ignore(self, repository_path, path):
       """
       Check if a path is ignored or not.
       
       Args:
         repository_path: Path to the Git repository
         path: Relative path to check
       """
        # in git.py Git, add a new method check_ignore that execute git check-ignore -z <path> and parse the result to return the list of ignored files; changed_files method is a good example of a similar method.

        cmd = ["git", "check-ignore", "-z", path]

        response = {}
        try:
            code, output, error = await execute(cmd, cwd=repository_path)
        except subprocess.CalledProcessError as e:
            response["code"] = e.returncode
            response["message"] = e.output.decode("utf-8")
        else:
            response["code"] = code
            if code != 0:
                response["command"] = " ".join(cmd)
                response["message"] = error
            else:
                response["files"] = strip_and_split(output)

        return response
```

jupyterlab_git/handlers.py
```diff
@@ -822,6 +822,10 @@ async def post(self, path: str = ""):
             self.set_status(500)
         self.finish(json.dumps(body))
 
+    @tornado.web.authenticated
+    def check_ignore(self, path):
+        self.git.check_ignore(self, path)
```
fcollonval, line 827: Here you want this method to be executed on an `GET` request. So the method should be named `get`. Then you will obtain the repository path as `path` input argument and you could for example get the relative path to check through a query argument like this:

```suggestion
    async def get(self, path):
        path_to_check = self.get_query_argument("path", "")
        body = await self.git.check_ignore(self, self.url2localpath(path), path_to_check)

        if body["code"] != 0:
            self.set_status(500)
        self.finish(json.dumps(body))
```

src/model.ts
```diff
@@ -80,6 +80,13 @@ export class GitExtension implements IGitExtension {
     }
   }
 
+  /**
+   * Requests the new GET endpoint.
+   */
+  get isIgnored(): Git.IBranch {
+    return;
+  }
```
fcollonval, line 88: Then to call the backend with the proper argument, you will need something like that

```suggestion
  async isIgnored(pathToCheck: string): Promise<{files: string[]}> {
    const path = await this._getPathRepository();
    return requestAPI<{files: string[]}>(
      URLExt.join(path, 'ignore') +  URLExt.objectToQueryString({path: pathToCheck}),
      'GET'
    );
  }
```

PR #1209, opened 2023-01-17
Improve HTTP status codes for the some remote git operations

MGSousa:
Improve HTTP status codes by switch some of the Server Error (5xx) to Client Error codes (4xx), since it is related to client specific actions.
Also this allows to not break some security rules for 5xx codes in NGINX and other reverse proxies, maintaining the feature as is.

### DIFF
--- a/jupyterlab_git/handlers.py
+++ b/jupyterlab_git/handlers.py
@@ -98,7 +98,7 @@ class GitCloneHandler(GitHandler):
         )
 
         if response["code"] != 0:
-            self.set_status(500)
+            self.set_status(401)
         self.finish(json.dumps(response))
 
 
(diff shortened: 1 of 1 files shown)


fcollonval (changes requested):
Thanks for the suggestion @MGSousa 

Although I don't agree with switching any code error to **401**. The git commands may fail for other reasons than an authentication error. We could raise an specific error for authentication failure and set the code to 401 in that specific case. But this should not be the default code.

The usage of **404** seems acceptable.

lufisaal:
Hello @fcollonval :)
Is there any news regarding this topic?

fcollonval (changes requested):
Thanks for the ping @lufisaal and sorry for not getting back at you @MGSousa 

I have one comment that prevent to get it merge as is.

PR #1212, opened 2023-02-02
Set file list overflow in both directions to hidden

### LINKED ISSUE #1210: The "add" (stage changes) button of last entry runs away
When you try to click the + button on the last entry with mouse it sometimes shifts away because scrollbar appears which makes clicking it difficult:

cannot-click

ardislu:
Otherwise a vertical scrollbar will appear when a button within the file list is clicked.

Closes #1210

### DIFF
--- a/src/components/GitStage.tsx
+++ b/src/components/GitStage.tsx
@@ -87,7 +87,7 @@ export const GitStage: React.FunctionComponent<IGitStageProps> = (
           itemData={props.files}
           itemKey={(index, data) => data[index].to}
           itemSize={ITEM_HEIGHT}
-          style={{ overflowX: 'hidden' }}
+          style={{ overflow: 'hidden' }}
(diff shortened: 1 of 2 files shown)


fcollonval (changes requested):
Thanks @ardislu 

Unfortunately, this will prevent scrolling long list of files. In the example below I get 44 files untracked. The list is not vertically scrollable any longer with your change. So I cannot look at all the files:

image

ardislu:
Thank you @fcollonval for this feedback, I didn't check the case where there's a lot of files.

I've moved the `overflow: hidden` to the lowest `<div>` that wraps the buttons. Now a long list will still be scrollable, but the original issue should be fixed.

fcollonval:
Kicking the CI

fcollonval (approved):
Thanks a lot @ardislu

PR #1214, opened 2023-02-04
Re-encode cwd if path does not exist

### LINKED ISSUE #1211: Git repositories with spaces in their name are not recognized by the extension
## Description

The git side panel does not recognize a git repo exists when you open a freshly-cloned repo that has spaces in its name.

For example: a repo with the name `repo with spaces in its name` is cloned to a folder named `repo%20with%20spaces%20in%20its%20name`. All the files inside the folder are cloned as expected. However, opening the folder does not trigger the git side panel to recognize any git repo.

## Reproduce

1. Create a new git repo that contains spaces in its name (NOTE: GitHub does not allow this, but other hosts such as [Azure DevOps](https://dev.azure.com/) do).
2. In Jupyter, clone the repo.
3. Open the folder and try to use the git side panel.

## Expected behavior

The repo is cloned successfully and the git side panel works as expected.

## Actual behavior

The repo is cloned successfully with the spaces URL encoded in the folder name (i.e. "%20" instead of spaces). However, the git side panel does not detect any git repo inside the folder (it shows the default "You are not currently in a Git repository" page).

## Workarounds

Workaround 1: Manually rename the folder to replace the "%20" encoding with spaces. After renaming the folder, the side panel works as expected.
Workaround 2: Open a new terminal and manually use the git CLI.

## Context

- Python package version: 0.41.0
- Extension version: 0.41.0
- Git version: 2.34.1

ardislu:
Fixes the case where a folder name is cloned with url-encoded spaces.

Closes #1211

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -11,7 +11,7 @@ import subprocess
 import traceback
 from typing import Dict, List, Optional
 from unittest.mock import NonCallableMock
-from urllib.parse import unquote
+from urllib.parse import quote, unquote
 
 import nbformat
 import pexpect
@@ -73,6 +73,9 @@ async def execute(
         env: "Optional[Dict[str, str]]" = None,
     ) -> "Tuple[int, str, str]":
         try:
+            if not os.path.exists(cwd):
+                cwd = quote(cwd, safe=":/\\")
+
             p = pexpect.spawn(
                 cmdline[0],
                 cmdline[1:],
@@ -112,6 +115,9 @@ async def execute(
         cwd: "Optional[str]" = None,
         env: "Optional[Dict[str, str]]" = None,
     ) -> "Tuple[int, str, str]":
+        if not os.path.exists(cwd):
+            cwd = quote(cwd, safe=":/\\")
+
         process = subprocess.Popen(
             cmdline, stdout=subprocess.PIPE, stderr=subprocess.PIPE, cwd=cwd, env=env
         )


fcollonval (commented):
I don't understand how this can fix the issue. Would you mind providing more explanation?

ardislu:
If I have a folder with the literal name:

```
repo%20with%20spaces%20in%20its%20name
```

jupyterlab-git does not recognize that any git repository exists in this folder (it shows the "You are not currently in a Git repository" page).

This folder is encoded into the Jupyter URL as:

```
http://localhost:8888/lab/tree/repo%2520with%2520spaces%2520in%2520its%2520name
```

I believe the problem is that this URL is somehow getting decoded twice (or more) by the time it reaches the `git.py` file.

The desired value after exactly 1 decode:

```
repo%20with%20spaces%20in%20its%20name
```

The actual value (2 or more decodes happening):

```
repo with spaces in its name
```

Since the folder `/repo with spaces in its name` does not exist, jupyterlab-git does not recognize any git repository exists.

I'm unable to identify where/how exactly the URL is getting decoded twice, to fix the root cause. So for this PR I suggest checking if the path exists:

```
/repo with spaces in its name
```

And if this path does not exist, try again but with the spaces encoded:

```
/repo%20with%20spaces%20in%20its%20name
```

This fix will correctly identify that there is a git repository inside a folder literally named: `/repo%20with%20spaces%20in%20its%20name`.

Hope that clarifies the problem and the fix.

fcollonval:
Thanks for the additional explanation @ardislu 

So it means that with this PR if the repository is named `repo with spaces in its name` it will actually be cloned within a folder named `repo%20with%20spaces%20in%20its%20name`. Is that correct?

fcollonval:
And if you use the `git clone` command in the terminal, what is happening? Is the folder name with space escaped or not?

ardislu:
> So it means that with this PR if the repository is named `repo with spaces in its name` it will actually be cloned within a folder named `repo%20with%20spaces%20in%20its%20name`. Is that correct?

Correct.

> And if you use the `git clone` command in the terminal, what is happening? Is the folder name with space escaped or not?

`git clone` also clones a folder literally named `repo%20with%20spaces%20in%20its%20name`.

---

Full end to end example:

1. Here's a repo named `repo with spaces in its name` in Azure DevOps (GitHub does not allow spaces in repo names):

How a repo named "repo with spaces in its name" appears in the Azure DevOps UI

2. Azure DevOps provides the following URL to clone this repo:

```
https://ardislu@dev.azure.com/ardislu/ardis.lu/_git/repo%20with%20spaces%20in%20its%20name
```

3. Calling:

```
git clone https://ardislu@dev.azure.com/ardislu/ardis.lu/_git/repo%20with%20spaces%20in%20its%20name
```

Gives me a folder literally named `repo%20with%20spaces%20in%20its%20name`. Using the Jupyter UI to clone also gives the same result.

fcollonval:
Thanks a lot for the reply.

My fear with the current fix is that it may escape unwanted space. For example if your repository is located at `/home/me/a sub folder with spaces/my%20repo`, you will escape the spaces in the subfolder. So a safer fix seems to require finding why `repo%2520with%2520spaces%2520in%2520its%2520name` ends up being decoded twice.

Issue #995, opened 2021-08-03
Switch from a detached head to a branch not working


fcollonval:
If the repository is in detached head (for example after switching to a tag), then switching to a branch will fail.

Error message:

```
fatal: ambiguous argument '(HEAD detached at v0.20.0)': unknown revision or path not in the working tree. Use '--' to separate paths from revisions, like this: 'git <command> [<revision>...] -- [<file>...]' 
```

![Peek 2021-08-03 19-06](https://user-images.githubusercontent.com/8435071/128057755-4437e620-6cba-4189-92a8-3ccb083b151d.gif)

Bl4ck-h00d:
Hi! I would like to work on this, can you help me how to get started?!

fcollonval:
Hey @Bl4ck-h00d 

Thanks for kicking in.

The extension has two parts: a front-end in Typescript and a back-end in Python. Whenever you trigger an action from the front-end, it will call a `git` command in the back-end.

In this case, to switch branch, the command call is done there:

https://github.com/jupyterlab/jupyterlab-git/blob/02c6d9aa8cf3344b37397e303c8d345efab12aff/jupyterlab_git/git.py#L880

The action is triggered in the frontend with that callback:

https://github.com/jupyterlab/jupyterlab-git/blob/02c6d9aa8cf3344b37397e303c8d345efab12aff/src/components/BranchMenu.tsx#L402

The workflow is the following:

1. `onClick` is triggered with the branch name
2. it calls the `model.checkout` method  https://github.com/jupyterlab/jupyterlab-git/blob/02c6d9aa8cf3344b37397e303c8d345efab12aff/src/model.ts#L426
> The model is handling the repository state and the call to the backend
3. That trigger a call to the backend https://github.com/jupyterlab/jupyterlab-git/blob/02c6d9aa8cf3344b37397e303c8d345efab12aff/src/model.ts#L469
4. The call is catch by a Python handler: https://github.com/jupyterlab/jupyterlab-git/blob/02c6d9aa8cf3344b37397e303c8d345efab12aff/jupyterlab_git/handlers.py#L443
5. That handler calls the git command mentioned above: https://github.com/jupyterlab/jupyterlab-git/blob/02c6d9aa8cf3344b37397e303c8d345efab12aff/jupyterlab_git/handlers.py#L456

The troubles is in the front-end because we want to deal with files that are different between two branches at

[...]

PR #1218, opened 2023-02-07
fix switch from a detached head to a branch not working

basokant:
Provides fix for #995 

If the repository is in detached head (for example after switching to a tag), then switching to a branch will no longer fail. Behaviour after changes:

https://user-images.githubusercontent.com/70033855/217340638-7ebe001e-f713-44dc-bcb4-65dc2c3bfced.mov

Transforms the fake branch name (HEAD detached at #####) to ##### using Regex.

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -562,6 +562,13 @@ export class GitExtension implements IGitExtension {
         let changes;
         if (!body.new_check) {
           if (body.checkout_branch && !body.new_check) {
+            const detachedHeadRegex = /\(HEAD detached at (.+)\)/;
+            let result = this._currentBranch.name.match(detachedHeadRegex);
+
+            if (result && result.length > 1) {
+              this._currentBranch.name = result[1];
+            }
+
             changes = await this._changedFiles(
               this._currentBranch.name,
               body.branchname
--- a/src/model.ts
+++ b/src/model.ts
@@ -562,13 +562,6 @@ export class GitExtension implements IGitExtension {
         let changes;
         if (!body.new_check) {
           if (body.checkout_branch && !body.new_check) {
-            const detachedHeadRegex = /\(HEAD detached at (.+)\)/;
-            const result = this._currentBranch.name.match(detachedHeadRegex);
-
-            if (result && result.length > 1) {
-              this._currentBranch.name = result[1];
-            }
-
             changes = await this._changedFiles(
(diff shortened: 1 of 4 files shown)


fcollonval (commented):
This looks great :rocket: 

I'm actually wondering if we should rather _clean_ the current branch name for consistency and to avoid other troubles.

The branch is set at

https://github.com/jupyterlab/jupyterlab-git/blob/c7e7363a5e749ea4590984f3a0303694f25212aa/src/model.ts#L1045

Would you mind trying applying your fix there to see if it works?

basokant:
> This looks great 🚀
> 
> I'm actually wondering if we should rather _clean_ the current branch name for consistency and to avoid other troubles.
> 
> The branch is set at
> 
> https://github.com/jupyterlab/jupyterlab-git/blob/c7e7363a5e749ea4590984f3a0303694f25212aa/src/model.ts#L1045
> 
> Would you mind trying applying your fix there to see if it works?

Yes I'll try moving the fix there!

src/model.ts
```diff
@@ -1042,6 +1042,15 @@ export class GitExtension implements IGitExtension {
       );
 
       this._branches = data.branches ?? [];
+
+      const detachedHeadRegex = /\(HEAD detached at (.+)\)/;
+      const result = data.current_branch.name.match(detachedHeadRegex);
+
+      if (result && result.length > 1) {
+        data.current_branch.tag = result[1];
```
fcollonval, line 1050: Let's not change the `tag` because the detached head case can occur when the git repository is checked out at a specific commit that is not a branch head.

```suggestion
```

fcollonval (changes requested):
Thanks @basokant 

Do you need help to fix the Jest tests?

fcollonval (changes requested):
Thanks for fixing the tests.

Are you ok with expanding this to add a new flag `detached` in `Git.IBranch` at

https://github.com/jupyterlab/jupyterlab-git/blob/master/src/tokens.ts#L816

And use that flag to tune the branch display at

https://github.com/jupyterlab/jupyterlab-git/blob/master/src/components/Toolbar.tsx#L267

basokant:
Relocated the translate of the branch toolbar title. Should be ready for a final review @fcollonval

fcollonval (approved):
Thanks @basokant

PR #1220, opened 2023-02-08
bug: modified password placeholder text depending if remote URI is github (Fix #1176)

### LINKED ISSUE #1176: Passwords no longer accepted by GitHub for pushing to https repos
## Description

When I try to push from a Git repo cloned with `https` I see this box:

image

While it says that I could enter my password, this is no longer supported by GitHub and if I do so I am shown this error message:

image

## Reproduce

Clone a repo via https, change something and commit it, try to push

## Expected behavior

Only mention PAT and include a link to https://docs.github.com/en/get-started/getting-started-with-git/about-remote-repositories#cloning-with-https-urls or https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token

## Context

- Extension version: 0.37.1

shawnesquivel:
This fixes #1176.

**Description of Changes**
- Check to see if the remote URI matches the GitHub expression
- Change the password placeholder text accordingly

**Git Push**

https://user-images.githubusercontent.com/94336773/219996988-e0dad60f-6e51-4e74-bfd3-6cfa22731d83.mov

**Git Clone**

https://user-images.githubusercontent.com/94336773/219996946-b3ecbd35-16a2-454e-b697-338f463d9555.mov

**Git Pull**

https://user-images.githubusercontent.com/94336773/220784704-a0246d05-3c29-48e2-a32c-a8a61d657980.mov

Collaborated with @basokant on this!

### DIFF
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -1713,7 +1713,6 @@ export async function showGitOperationDialog<T>(
           });
           remoteGitProvider = checkUrlGitProvider(remoteList[0]?.url);
           break;
-        // executed if none of the cases were matched
         default:
           break;
       }
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -397,7 +397,6 @@ export function addCommands(
       try {
         // copy how gitPush grabs remote - this needs to be fixed as it creates a Git Push UI
         // is there a better way to grab the remoteName?
-        let remote;
         const result = await showDialog({
           title: trans.__('Please select push options.'),
           body: new AdvancedPushForm(trans, gitModel),
@@ -406,7 +405,7 @@ export function addCommands(
             Dialog.okButton({ label: trans.__('Proceed') })
           ]
         });
-        remote = result.value.remoteName;
+        const remote = result.value.remoteName;
 
         if (args.force) {
           await discardAllChanges(gitModel, trans, args.fallback as boolean);
@@ -1610,7 +1609,7 @@ export async function showGitOperationDialog<T>(
    */
   function checkUrlGitProvider(remoteUrl: string): string {
     // Regex returns the word between "https" and "."
-    const re = /https:\/\/([^\.]+)\./;
+    const re = /https:\/\/([^.]+)\./;
     const result = remoteUrl.match(re) ?? [];
     const gitProvider = result[1];
     return gitProvider;
@@ -1666,7 +1665,7 @@ export async function showGitOperationDialog<T>(
     ) {
       // Change the placeholder message for GitHub
       let gitPasswordPlaceholder = 'password / personal access token';
-      let remoteGitProvider: string;
+      let remoteGitProvider: string = '';
 
       switch (operation) {
         case Operation.Clone:
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -1586,7 +1586,6 @@ export async function showGitOperationDialog<T>(
     if (currentRemote) {
       return currentRemote?.url;
     } else {
-      // if the current remote isn't inside, what is best to return? Could maybe throw an error?
       return '';
     }
   }
@@ -1668,8 +1667,8 @@ export async function showGitOperationDialog<T>(
           // If the remote is defined, check it against the remote URI list
           if (model.currentBranch.upstream) {
             // Compare the remote against the URI list
-            const remote = model.currentBranch.upstream.split('/')[0];
-            const currentRemoteUrl = await getCurrentRemote(remote);
+            const remoteName = model.currentBranch.upstream.split('/')[0];
+            const currentRemoteUrl = await getCurrentRemote(remoteName);
             remoteGitProvider = currentRemoteUrl
               ? getGitProviderHost(currentRemoteUrl)
               : '';
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -72,13 +72,6 @@ export interface IGitCloneArgs {
   submodules?: boolean;
 }
 
-// interface IGitPullArgs {
-//   remote?: string;
-//   name?: string;
-// }
-
-// interface IGitCloneOrPullArgs extends IGitCloneArgs, IGitPullArgs {}
-
 /**
  * Git operations requiring authentication
  */
@@ -1684,7 +1677,6 @@ export async function showGitOperationDialog<T>(
         case Operation.Push:
         case Operation.ForcePush:
         case Operation.Pull:
-          // Get the remote from the args
           const remote = (
             args as unknown as {
               remote: string;
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -1592,23 +1592,19 @@ export async function showGitOperationDialog<T>(
   authentication?: Git.IAuth,
   retry = false
 ): Promise<string> {
-  console.log('model', model);
-  console.log('operation', operation);
-  console.log('args', args);
-  console.log('authentication', authentication);
-
-  // Takes the remote URI and checks it against the list of git remotes
-  async function getCurrentRemote(remote: string): Promise<string> {
+  /**
+   * Returns the current remote's URL based on the current remote name and all the remotes
+   */
+  async function getCurrentRemote(currentRemoteName: string): Promise<string> {
     const remoteList = await model.getRemotes().then(remoteList => {
       return remoteList;
     });
 
     const currentRemote = remoteList.find(
-      remoteURI => remoteURI.name === remote
+      remoteURI => remoteURI.name === currentRemoteName
     );
 
     if (currentRemote) {
-      console.log(currentRemote?.url);
       return currentRemote?.url;
     } else {
       // if the current remote isn't inside, what is best to return? Could maybe throw an error?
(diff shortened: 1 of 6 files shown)


fcollonval (changes requested):
Thanks @shawnesquivel 

As mentioned in the issue, the trouble here is that this is specific to a Git server policy. For example other providers such GitLab or BitBucket may still allow that authentication system.

This means we need to condition the placeholder with the remote repository URI.

fcollonval:
The dialog is spinned there:

https://github.com/fcollonval/jupyterlab-git/blob/11258694f48262fd7542827f6f7e316d17cdfc45/src/commandsAndMenu.tsx#L1626

So we could pass the placeholder to the dialog constructor.

Now about the remote repository URI, this is gonna be more complex. In the case of a `clone` operation, we have the `url` in `args`. But for `pull` and `push`, we know the `remote` name but not the URI. You could find the URI by requesting the remotes with the following function:

https://github.com/fcollonval/jupyterlab-git/blob/11258694f48262fd7542827f6f7e316d17cdfc45/src/model.ts#L460

Does that make sense?

shawnesquivel:
Hi @fcollonval!

Added some changes + working screenshot.  

It is working for GitHub but remains as the default 'password / personal access token' otherwise.

@basokant also worked with me again on this one, he was a big help!

Let me know if there are any other changes to be made.

src/widgets/CredentialsBox.tsx
```diff
@@ -10,13 +10,16 @@ export class GitCredentialsForm
   extends Widget
   implements Dialog.IBodyWidget<Git.IAuth>
 {
+  passwordPlaceholder: string;
```
fcollonval, line 13: Purely styling comment:

Could you place this attribute at the bottom of the class with the other attributes and make it private (with the convention of starting with `_`)?

src/commandsAndMenu.tsx
```diff
@@ -1620,13 +1620,28 @@ export async function showGitOperationDialog<T>(
...
       )
     ) {
+      const remoteURI = await model.getRemotes().then(remote => {
+        return remote;
+      });
+
+      // Determine the Git Provider
+      const re = /https:\/\/(.+)\.com/;
+      const result = remoteURI[0]?.url.match(re);
+      const gitProvider = result[1];
```
fcollonval, line 1630: Result may be undefined so you should protect the code against that; something like:

```suggestion
      const result = remoteURI[0]?.url.match(re) ?? [];
      const gitProvider = result[1];
```

src/commandsAndMenu.tsx
```diff
@@ -1620,13 +1620,28 @@ export async function showGitOperationDialog<T>(
         errorMessage => (error as Error).message.indexOf(errorMessage) > -1
       )
     ) {
+      const remoteURI = await model.getRemotes().then(remote => {
+        return remote;
+      });
```
fcollonval, line 1625: Good this is the idea indeed. But now we need to be smart depending on the `operation` type. If the operation is `Operation.Clone`, we don't need to request the remotes as we can look directly at `args.url`. In the cases of push and pull, we should not take `remoteURI[0]` but we should find the item matching `args.remote`. Then for the fetch case, let's use your fallback of only testing `remoteURI[0]`.

fcollonval (changes requested):
Thanks @shawnesquivel 

This is looking good. We now have to improve the url search depending on the operation to be carried out.

src/widgets/CredentialsBox.tsx
shawnesquivel: Done. I'm guessing `_` is only for the private attribute, so I didn't change it on the parameter.

src/commandsAndMenu.tsx
shawnesquivel: @fcollonval  

**For Pull:**
Pull - I tried accessing args. When logging it, I noticed args is an empty object (no remote property), so I passed it in using the code below, which showed me that args.remote = 'origin'. But I don't think I fully understand what you mean by 'finding the item matching args.remote', do you mind explaining further? I guess I just don't see how the args.remote is going to tell us that the git provider is github (if that was what you were suggesting).

commandsAndMenu.tsx Line 383
```
  /** Add git pull command */
  commands.addCommand(CommandIDs.gitPull, {
   
    execute: async args => {
      try {
        // copy how commandIDS.gitPush grabs remote
        let remote;
        const result = await showDialog({
          title: trans.__('Please select push options.'),
          body: new AdvancedPushForm(trans, gitModel),
          buttons: [
            Dialog.cancelButton({ label: trans.__('Cancel') }),
            Dialog.okButton({ label: trans.__('Proceed') })
          ]
        });
        remote = result.value.remoteName;

        // added args as an additional param
        const details = await showGitOperationDialog(
          gitModel,
          Operation.Pull,
          trans,
          (args = { remote })
        );

```

**For Push**
- By default, args is an object with remote property, but it was undefined. Like for the pull, I'm not sure if I'm meant to pass in the remote by modifying

fcollonval: `git` uses name for remote to be more user friendly. Then it uses a db to get the URI from the name.

You can see that by executing in a terminal `git remote -vv`; the output will look like:

```sh
origin  git@github.com:fcollonval/jupyterlab.git (fetch)
origin  git@github.com:fcollonval/jupyterlab.git (push)
trungleduc      https://github.com/trungleduc/jupyterlab (fetch)
trungleduc      https://github.com/trungleduc/jupyterlab (push)
upstream        https://github.com/jupyterlab/jupyterlab.git (fetch)
upstream        https://github.com/jupyterlab/jupyterlab.git (push)
```

Name `origin` refers `git@github.com:fcollonval/jupyterlab.git` in the above example.

---

As you discover (:clap: for that), you could get the name passing the `remote` variable. But to determine the associated URI, you first need to call `getRemote` (as you have done already). It returns you an array of `Git.IGitRemote`. That interface has the following definition:

```typescript
interface IGitRemote {
    url: string;
    name: string;
  }
```

So now that you know `remote` you need to find the `IGitRemote` object in the array with that name. Then you will be able to test the URL.

Does it clarify what you need to do?

---

Regarding push, if it is undefined, you could still apply the same logic and fallback to `remoteURI[0].url` if it is undefined.

> We could improve this by having the backend tell use which of the remote is the default. But let's try not to increase the scope of this PR too much.

shawnesquivel: Hi @fcollonval !

Yes this was a very thorough explanation, thank you!

I was able to add the edits you requested for pull, push, clone, and fetch.

I have two pending issues/comments:
1. How do I test git fetch?
- I tried using `git fetch -all` which returned 'Fetching origin'
- However, it didn't throw any errors even though we are not authenticated, so I'm not sure how to proceed.
image

2. In git pull, I passed the args to the git-pull catch block by using the following code. It's wrong since it creates an unnecessary dialog box just to get the remote URI. Is there some other way of getting the current remote name that is better? If not, I'll just remove this section as there are fallbacks in place already.
```
        let remote;
        const result = await showDialog({
          title: trans.__('Please select push options.'),
          body: new AdvancedPushForm(trans, gitModel),
          buttons: [
            Dialog.cancelButton({ label: trans.__('Cancel') }),
            Dialog.okButton({ label: trans.__('Proceed') })
          ]
        });
        remote = result.value.remoteName;
```

3. I grouped git pull, push and force in the catch block because that made sense to me, but there may be some nuances here that I'm overlooking.

src/commandsAndMenu.tsx
```diff
@@ -1571,6 +1584,37 @@ export async function showGitOperationDialog<T>(
...
+    } else {
+      // if the current remote isn't inside, what is best to return? Could maybe throw an error?
+      return '';
+    }
+  }
+
+  /**
+   * Returns the Git provider based on the domain name of the url
+   */
+  function checkUrlGitProvider(remoteUrl: string): string {
```
fcollonval, line 1610: We could rename this to match better the purpose of the function:

```suggestion
  function getGitProviderHost(remoteUrl: string): string {
```

src/commandsAndMenu.tsx
```diff
@@ -1612,21 +1656,71 @@ export async function showGitOperationDialog<T>(
...
+      }
       // If the error is an authentication error, ask the user credentials
       const credentials = await showDialog({
         title: trans.__('Git credentials required'),
         body: new GitCredentialsForm(
           trans,
           trans.__('Enter credentials for remote repository'),
-          retry ? trans.__('Incorrect username or password.') : ''
+          retry ? trans.__('Incorrect username or password.') : '',
+          trans.__(gitPasswordPlaceholder)
```
fcollonval, line 1723: This unfortunately will not work as expected. To figure out all the string that requires translation, we cannot rely on executing the code as it may be very complex to get the application in all possible state. So the strings are extracted from a static analysis of the code; i.e. all string passed to `trans.__` (and similar) are extracted in the to be localized string database. This means that you need to wrap the constant strings you are setting `gitPasswordPlaceholder` with (i.e. lines 1667 and 1714).

```suggestion
          gitPasswordPlaceholder
```

src/widgets/CredentialsBox.tsx
```diff
@@ -10,13 +10,16 @@ export class GitCredentialsForm
   extends Widget
   implements Dialog.IBodyWidget<Git.IAuth>
 {
+  private _passwordPlaceholder: string;
   constructor(
     trans: TranslationBundle,
     textContent = trans.__('Enter credentials for remote repository'),
-    warningContent = ''
+    warningContent = '',
+    passwordPlaceholder = 'password / personal access token'
```
fcollonval, line 18: For the reason explain earlier.

```suggestion
    passwordPlaceholder = trans.__('password / personal access token')
```

fcollonval (commented):
Thanks @shawnesquivel 

Here are some comments and suggestions

src/commandsAndMenu.tsx
```diff
@@ -1571,6 +1571,36 @@ export async function showGitOperationDialog<T>(
...
+
+    const currentRemote = remoteList.find(
+      remoteURI => remoteURI.name === currentRemoteName
+    );
+
+    if (currentRemote) {
+      return currentRemote?.url;
+    } else {
+      return '';
+    }
```
fcollonval, line 1590: FYI you could have use a denser code:

```suggestion
    return currentRemote?.url ?? '';
```

src/commandsAndMenu.tsx
```diff
@@ -1612,21 +1642,65 @@ export async function showGitOperationDialog<T>(
...
+        case Operation.Fetch:
+          remoteGitProvider = await model
+            .getRemotes()
+            .then(remoteList => remoteList[0]?.url);
+          break;
+        default:
+          break;
+      }
+      // GitHub only verifies with personal access tokens
+      if (remoteGitProvider && remoteGitProvider.toLowerCase() === 'github') {
```
fcollonval, line 1693: FYI you do not need to test for `remoteGitProvider` being defined and not empty string. The second test is sufficient.

```suggestion
      if (remoteGitProvider.toLowerCase() === 'github') {
```

PR #1223, opened 2023-02-13
Adding support for standalone diffs of images

basokant:
A continuation of the work done in #1179 , including implementing an interactive image compare widget.
This PR intends to add the feature outlined in #664 .

Current progress:

https://user-images.githubusercontent.com/70033855/232121803-a08ed3a7-f0c9-4762-ae5c-ba2f6bdbcc51.mov

Tasks:

- [x] 2-up View
- [x] Onion Skin View
- [x] Swipe View
- [x] Fix colours (especially for dark theme)
- [x] Handle case where reference and challenger have different aspect ratios
- [x] Handle case where image diff is opened where the image didn't exist in the reference (it was created in the challenger commit). ie. the diff is opened from the commit where the image was created w/o compare
- [x] Add support for multiple image formats
- [x] Fix failing tests
- [x] Write ui tests


basokant:
@fcollonval I ended up using [react-compare-image](https://www.npmjs.com/package/react-compare-image) instead of the component you suggested for the image comparing functionality. I'm wondering if there's a suggested method to adjust in the height of the Panel widget. Currently, the widget fits the entire width of the panel.

src/components/diff/ImageDiff.tsx
```diff
@@ -0,0 +1,128 @@
...
+        this._model.reference.content(),
+        this._model.challenger.content()
+        // this._model.base?.content() ?? Promise.resolve(null)
+      ]);
+
+      const compareImageWidget = ReactWidget.create(
+        <CompareImage reference={reference} challenger={challenger} />
+      );
+
+      this.addWidget(compareImageWidget);
```
basokant, line 100: @fcollonval I ended up using [react-compare-image](https://www.npmjs.com/package/react-compare-image) instead of the component you suggested for the image comparing functionality. I'm wondering if there's a suggested method to adjust in the height of the Panel widget. Currently, the widget fits the entire width of the panel (as shown by the video in the PR description).

src/components/diff/ImageDiff.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+export class ImageDiff extends Panel implements Git.Diff.IDiffWidget {
+  constructor(model: Git.Diff.IModel, translator?: TranslationBundle) {
+    super();
+    const getReady = new PromiseDelegate<void>();
+    this._container = this.node as HTMLElement;
+    this._isReady = getReady.promise;
+    this._model = model;
+    this._trans = translator ?? nullTranslator.load('jupyterlab_git');
+
```
fcollonval, line 51: In here you can add a class to the container:

```typescript
this.addClass('jp-git-imagediff');
```

Then you could use it to set some CSS rules in a css file with the `style` folder.

fcollonval (commented):
This looks awesome :tada: 

Do you think you could add a dropbox to switch between viewers  _2-up_ | _Slider_ | _Onion_ (and of course add those :wink:) ?

basokant:
> This looks awesome 🎉
> 
> Do you think you could add a dropbox to switch between viewers _2-up_ | _Slider_ | _Onion_ (and of course add those 😉) ?

This could be difficult, perhaps if I try using the component that you recommended instead this would be quite easy 🤔.

fcollonval:
For an example of the image diff viewer in GitHub: https://github.com/jupyterlab/jupyterlab/pull/13861/files#diff-9f79052c539324287f8df1779707857e4e41bc23fc336393de60dcbadf0b8cd9

basokant:
Hi @fcollonval! I attempted to replace the react image compare component with the react image diff component (https://github.com/cezary/react-image-diff), which would give us the views that you suggested for free. I was able to add an external TypeScript declaration file, but after integrating it into our component, it broke the extension with a runtime error:

```
TypeError: Cannot read properties of undefined (reading 'string')
    at Object.<anonymous> (react-image-diff.js:277:1)
```

This is caused by the use of PropTypes to validate the types of props. Additionally, many depreciated react class component methods were used in the component, such as `componentWillMount`. I think that the right course of action is to either

1. Create a completely new react component from scratch using the GitHub image diff as an inspiration
2. Use the react-image-compare component and try and build the other desired views on top of it.

How should I progress from here?

fcollonval:
> Hi @fcollonval! I attempted to replace the react image compare component with the react image diff component (https://github.com/cezary/react-image-diff), which would give us the views that you suggested for free. I was able to add an external TypeScript declaration file, but after integrating it into our component, it broke the extension with a runtime error:
> 
> ```
> TypeError: Cannot read properties of undefined (reading 'string')
>     at Object.<anonymous> (react-image-diff.js:277:1)
> ```
> 
> This is caused by the use of PropTypes to validate the types of props. Additionally, many depreciated react class component methods were used in the component, such as `componentWillMount`. I think that the right course of action is to either
> 
>     1. Create a completely new react component from scratch using the GitHub image diff as an inspiration
> 
>     2. Use the react-image-compare component and try and build the other desired views on top of it.
> 
> 
> How should I progress from here?

Thanks for trying it out @basokant - I would go for the second path. Looking at the code of react-image-diff, it is quite short. So I guess it will be possible to understand how they do the onion view.

fcollonval:
> * I assumed that the image is a PNG. **Should we add support for multiple image formats** (like JPEG)? This would require some work on the back-end i believe (unless I'm missing something)

Yes we should support more formats. But actually, it is in the frontend when you set the source

```
src={`data:image/png;base64,${reference}`}
```

That function may be helpful as it allows to get the filetype that got an attribute `mimeTypes` that should provide you the image type:

https://github.com/jupyterlab/jupyterlab-git/blob/dda8e82b9b2751e1b2006f71cb101accef261d10/src/model.ts#L1614

> * (Less important) Should I **change the slider width to be the same (largest) image width** for Swipe and Onion Skin?

This seems like a nice improvement

> * Is the **border a good enough indicator** for which image/side is the reference vs the challenger? I was thinking of either changing the colour of the slider depending on the slider value and/or adding coloured circles to the left and right of the slider.

I let you test and propose what you think more appropriate

jupyterlab_git/git.py
```diff
@@ -110,12 +112,20 @@ def call_subprocess(
...
         output, error = process.communicate()
-        return (process.returncode, output.decode("utf-8"), error.decode("utf-8"))
+        if is_binary == False:
+            return (process.returncode, output.decode("utf-8"), error.decode("utf-8"))
+        else:
+            return (
+                process.returncode,
+                base64.encodebytes(output).decode("ascii"),
+                error.decode("utf-8"),
+            )
```
fcollonval, line 128: let's make it simpler :wink: 

```suggestion
        if is_binary:
            return (
                process.returncode,
                base64.encodebytes(output).decode("ascii"),
                error.decode("utf-8"),
            )
        else:
            return (process.returncode, output.decode("utf-8"), error.decode("utf-8"))
```

src/components/diff/ImageDiff.tsx
```diff
@@ -0,0 +1,545 @@
...
+    height = Math.min(event.currentTarget.height, 400);
+    setImageDimensions([width, height]);
+  };
+
+  useEffect(() => {
+    if (!reference) {
+      setReferDimensions([...challDimensions]);
+    }
+
+    console.log('effect');
```
fcollonval, line 210: Debug print

```suggestion
```

src/components/diff/ImageDiff.tsx
```diff
@@ -0,0 +1,545 @@
...
+  }, [challDimensions]);
+
+  useEffect(() => {
+    if (!challenger) {
+      setChallDimensions([...referDimensions]);
+    }
+
+    console.log('effect');
```
fcollonval, line 218: Debug print

```suggestion
```

src/components/diff/ImageDiff.tsx
```diff
@@ -0,0 +1,545 @@
...
+  get isFileResolved(): boolean {
+    return null;
+  }
+
+  async getResolvedFile(): Promise<Partial<Contents.IModel>> {
+    return null;
+  }
+
+  async refresh(): Promise<void> {
+    // await this.ready;
```
fcollonval, line 474:
```suggestion
```

fcollonval (changes requested):
Thanks @basokant and sorry for the late review.

Overall it looks good. Would you be able to finish it up and add integration tests? Otherwise I'm happy to finish the PR.

PR #1250, opened 2023-06-20
Add git_command_timeout_s for allowing >20 seconds for git operations

mdietz94:
Some large repositories may take more than 20 seconds, especially for the initial git clone. Adds a configuration option to make this configurable.


package.json
```diff
@@ -1,218 +1,218 @@
 {
-    "name": "@jupyterlab/git",
```
mdietz94, line 2: This file is unchanged other than prettier running on it

mdietz94:
I believe this PR passes outside of existing breaks, please let me know if there is anything else I can do to help.

fcollonval (changes requested):
Thanks @mdietz94 

I have minor suggestions (mainly about the naming) and we should use a float as we can rather than an integer.

Would you mind rebasing your branch too as I fixed the CI error in #1252

mdietz94:
Thanks! Rebased and applied your comments. Let me know if you need anything else.

fcollonval (approved):
Thanks @mdietz94 and sorry for the late review

PR #1251, opened 2023-06-21
Add option to ask user identity on every commit

eyusupov:
This is useful for situation when multiple people collaborate in the single Jupyter Lab instance so that people can commit PRs under their own name  and specify their own emails.


src/components/GitPanel.tsx
```diff
@@ -794,19 +794,30 @@ export class GitPanel extends React.Component<IGitPanelProps, IGitPanelState> {
    *
    * @param path - repository path
    */
-  private async _hasIdentity(path: string): Promise<void> {
-    // If the repository path changes, check the user identity
-    if (path !== this._previousRepoPath) {
+  private async _hasIdentity(path: string): Promise<string> {
```
fcollonval, line 797: let's return `null` if no transient identity is set.

```suggestion
  private async _hasIdentity(path: string): Promise<string | null> {
```

src/components/GitPanel.tsx
```diff
@@ -794,19 +794,30 @@ export class GitPanel extends React.Component<IGitPanelProps, IGitPanelState> {
...
+    // If the repository path changes or explicitly configured, check the user identity
+    if (
```
fcollonval, line 799:
```suggestion
    let authorOverride: string | null = null;
    if (
```

src/components/GitPanel.tsx
```diff
@@ -794,19 +794,30 @@ export class GitPanel extends React.Component<IGitPanelProps, IGitPanelState> {
...
+      this.props.settings.composite['promptUserIdentity']
+    ) {
       try {
         const data: JSONObject = (await this.props.model.config()) as any;
         const options: JSONObject = data['options'] as JSONObject;
         const keys = Object.keys(options);
-
-        // If the user name or e-mail is unknown, ask the user to set it
-        if (keys.indexOf('user.name') < 0 || keys.indexOf('user.email') < 0) {
+        let authorOverride: string = null;
```
fcollonval, line 807:
```suggestion
```

src/components/GitPanel.tsx
```diff
@@ -794,19 +794,30 @@ export class GitPanel extends React.Component<IGitPanelProps, IGitPanelState> {
...
-    // If the repository path changes, check the user identity
-    if (path !== this._previousRepoPath) {
+  private async _hasIdentity(path: string): Promise<string> {
+    // If the repository path changes or explicitly configured, check the user identity
+    if (
+      path !== this._previousRepoPath ||
+      this.props.settings.composite['promptUserIdentity']
+    ) {
       try {
         const data: JSONObject = (await this.props.model.config()) as any;
```
fcollonval, line 804: I'ld like to avoid requesting the server every time.

Could you a new attribute to the `GitModel` - `lastAuthor: {name: string, email: string} | null`?

So calling `model.config()` will only happen if `lastAuthor` is `null`?

src/components/GitPanel.tsx
```diff
@@ -816,12 +827,23 @@ export class GitPanel extends React.Component<IGitPanelProps, IGitPanelState> {
           }
 
           const { name, email } = result.value;
-          await this.props.model.config({
-            'user.name': name,
-            'user.email': email
-          });
+          if (
+            keys.indexOf('user.name') < 0 ||
+            keys.indexOf('user.email') < 0
```
fcollonval, line 832: Let's store that condition to reduce keys inspection.

src/widgets/AuthorBox.ts
```diff
@@ -9,8 +9,10 @@ export class GitAuthorForm
   extends Widget
   implements Dialog.IBodyWidget<Git.IIdentity>
 {
-  constructor() {
+  constructor(userName = '', userEmail = '') {
```
fcollonval, line 12: Could we use an object here so it will be easier to extend and avoid the weird call if extended later as the one you saw to call `model.commit` that force to explicit argument that use their default values.

```suggestion
  constructor(options: {userName?: string, userEmail?: string} = {}) {
```

src/widgets/AuthorBox.ts
```diff
@@ -9,8 +9,10 @@ export class GitAuthorForm
...
     super();
+    this._userNameInitialValue = userName;
+    this._userEmailInitialValue = userEmail;
```
fcollonval, line 15:
```suggestion
    const { userName, userEmail } = options;
    this._userNameInitialValue = userName ?? '';
    this._userEmailInitialValue = userEmail ?? '';
```

fcollonval (changes requested):
Thanks @eyusupov this looks great

I have a couple of suggestions.

> Would you mind also rebasing your branch to get the CI fixes I pushed in #1252

src/widgets/AuthorBox.ts
eyusupov: I reused Git.IIdentity here and in some other places instead of (?) anonymous class. Let me know if you have any objections. I'm not TS experts, but it seems to be pretty clean now

eyusupov:
Integration tests failures seem to be unrelated - I see them [failing in another PR](https://github.com/jupyterlab/jupyterlab-git/actions/runs/5330220045/jobs/9656774891?pr=1250) too.

fcollonval (approved):
Thanks a lot @eyusupov 

I pushed a commit to improve slightly the look of the author form.

PR #1259, opened 2023-08-03
Switch back to using the Jupyter Releaser actions

jtpio:
Using the trusted publisher workflow for PyPI and npm are now available in the releaser: https://github.com/jupyter-server/jupyter_releaser/pull/511


fcollonval (approved):
Thanks @jtpio

PR #1271, opened 2023-10-02
Add constrain in install instructions

### LINKED ISSUE #1270: Installation guide leads to incompatibility
Hi!

I would like to tell you that the installation guide needs to be adjusted.
The extension requires a Jupyterlab version < 4.

Executing the command from the readme.md `pip install --upgrade jupyterlab jupyterlab-git`, you automatically update Jupyter to 4.X which disables jupyterlab-git due to an incompatibility.

fcollonval:
Fixes #1270

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,13 +26,13 @@ To see the extension in action, open the example notebook included in the Binder
 To install perform the following steps, with _pip_:
 
 ```bash
-pip install --upgrade jupyterlab jupyterlab-git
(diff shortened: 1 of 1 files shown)


markusReinert:
It seems like the [Requirements section of README](https://github.com/jupyterlab/jupyterlab-git/#requirements) does not agree with the installation guide anymore. While the installation guide asks for "jupyterlab<4", the Requirements say "JupyterLab >= 4.0". This sounds contradictory to me.

PR #1272, opened 2023-10-05
Show tags in history sidebar

DenisaCG:
In the history panel, show the tags (if they exist) next to the commits they are pointing to. 

For this, the implementation of an `ITag` interface was required, such that for each tag we keep not only its name, but also the id of the commit it points to. 

```
export interface ITag {
    name: string;
    baseCommitId?: string;
}
```


fcollonval (approved):
Thanks @DenisaCG

PR #1277, opened 2023-10-19
Add context menu on commits in the HistorySidebar with add tag command

DenisaCG:
Created a context menu for the commits showing inside the HistorySidebar and added a new command `Add Tag`. 

Users can now right click on a commit and use the `Add Tag` command to directly create a tag pointing to that specific commit. Once clicking on the command, the `NewTagDialogBox` pops up, showing only the selected commit. 

[Screencast-Add-Tag-from-Context-Menu.webm](https://github.com/jupyterlab/jupyterlab-git/assets/91504950/b38fd172-7910-4f27-b2a1-11b930b85d91)

### DIFF
--- a/src/commandsAndMenu.tsx
+++ b/src/commandsAndMenu.tsx
@@ -1569,7 +1569,6 @@ export function addCommands(
           trans={trans}
           open={tagDialog}
           onClose={(tagName?: string) => {
-            // () => {
             dialog.dispose();
             waitForDialog.resolve(tagName ?? null);
           }}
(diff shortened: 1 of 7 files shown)


fcollonval (changes requested):
Thanks Denisa, I have only a single request. Would you mind not rendering the filter div when `isSingleCommit` == true in the `NewTagDialog`?

DenisaCG:
Thank you for the review, Frédéric! And great point, I made the changes such that the filter input box disappears when adding a tag from the context menu. I'm also working on creating an integration test.

fcollonval (approved):
Thanks @DenisaCG

PR #1278, opened 2023-10-20
Deny git access to nested server root

### LINKED ISSUE #1205: Files do not open from side panel if lab started from subdirectory of git
## Description

When working on jupyterlab repo if I start jupyterlab from `packages` sub-directory and try to open a modified file from the git sidebar it shows:

```
Uncaught (in promise) Error: No such file or directory: packages/completer/style/base.css
```

The path here is wrong: it starts at the git root, but it should be relative to jupyterlab root.

## Context

- Python package version: 0.41.0
- Extension version: `@jupyterlab/git v0.41.0 enabled OK (python, jupyterlab-git)`
- Git version: 2.37.2
- Operating System and its version: Ubuntu, latest

Zxun2:
Fixes #1205 

# Changes
1. `IShowPrefixResult` now returns an additional attribute `relative_path` which determines the current path relative to the server root directory.
2. Using `relative_path`, i compared the server root to the git root. The main logic is that if `relative_path == "."` then you simply check if `path === ""`. This would be an indication that server root is equal to the git root since path is relative to git root. Otherwise, determine if `normalize(relative_path) === normalize(path)`. If git root is indeed equal to server root, then relative path must be equal to the path itself.


jupyterlab_git/git.py
```diff
@@ -955,8 +955,13 @@ async def show_prefix(self, path):
...
             cwd=path,
         )
+
         if code == 0:
-            result = {"code": code, "path": my_output.strip("\n")}
+            result = {
+                "code": code,
+                "path": my_output.strip("\n"),
+                "relative_path": os.path.relpath(path, contents_manager.root_dir),
+            }
```
fcollonval, line 964: Thanks @Zxun2 

This is going in the right direction. However I would encapsulate all the logic here on the server side.

So the code will be something like:
```suggestion
            relative_path = os.path.relpath(path, contents_manager.root_dir)
            result = {
                "code": code,
                "path": None if relative_path == "." else my_output.strip("\n"),
            }
```

fcollonval:
@Zxun2 the main branch has been updated to JupyterLab 4 - to ease the work here, I changed the target to the branch `jlab-3` to avoid complex rebasing.

Zxun2:
I have made the code ever slightly more compact by shifting all the relevant logic to server side. Let me know if there are any more changes to be made!

src/model.ts
```diff
@@ -1793,7 +1793,6 @@ export class GitExtension implements IGitExtension {
    * @throws {ServerConnection.NetworkError} If the request cannot be made
    */
   async tags(): Promise<Git.ITagResult> {
-    const path = await this._getPathRepository();
```
fcollonval, line 1796: This looks like a mistake when rebasing. Would you mind restoring that line?

jupyterlab_git/git.py
```diff
@@ -955,9 +955,23 @@ async def show_prefix(self, path):
...
+                "path": relative_git_path,
+                "relative_path": relative_server_path,
+            }
+
+            if relative_server_path == "." and relative_git_path == "":
+                return result
+            elif relative_git_path.split("/")[:-1] == relative_server_path.split("\\"):
+                return result
+            else:
+                return {"code": code, "path": None}
```
fcollonval, line 974: Ok I was playing around a bit also to try  to avoid the `split("\\")` that I think will only work for windows.

Here is an alternative suggestion

```suggestion
            relative_git_path = my_output.strip("\n")
            repository_path = path[:-len(relative_git_path)]

            try:
                # Raise an error is the repository_path is not a subpath of root_dir
                Path(repository_path).absolute().relative_to(Path(contents_manager.root_dir).absolute())
            except ValueError:
                return {"code": code, "path": None}
            else:
                result = {
                    "code": code,
                    "path": relative_git_path,
                }
```

What do you think?

fcollonval (changes requested):
Thanks @Zxun2 

I looked a bit at the new `pathlib` library that is much better than `os.path`. So here is an alternative proposal.

Would you mind adding a test for this? You can look at [that test](https://github.com/jupyterlab/jupyterlab-git/blob/a901350985c00bd532fc78c56a8ab5f8b13cef0c/jupyterlab_git/tests/test_handlers.py#L133) as an example to start a new one.

Zxun2:
Hey @fcollonval! Thanks for the review. I tried looking into the test case that you linked to, but i am a little confused how the function works. Both `test_git_show_prefix` and `test_git_show_prefix_not_a_git_repo` appears to have the similar method calls and code logic but has different assertions (expected return value). It'll be great if you can give some insights on how these method calls work!

For example, 
```
local_path = jp_root_dir / "test_path"
```
Changing "test_path" to any other string causes the testcase to fail, but i am unsure why. Are these strings predetermined (from the codebase, i only see "test_path" and "ignored_path")?

fcollonval:
Thanks @Zxun2 

The difference between the two tests is the mocked result from the git command:

* Returning a valid answer in `test_git_show_prefix`

https://github.com/jupyterlab/jupyterlab-git/blob/a901350985c00bd532fc78c56a8ab5f8b13cef0c/jupyterlab_git/tests/test_handlers.py#L85

* Returning an error in `test_git_show_prefix_not_a_git_repo`

https://github.com/jupyterlab/jupyterlab-git/blob/a901350985c00bd532fc78c56a8ab5f8b13cef0c/jupyterlab_git/tests/test_handlers.py#L138

So in your case you will need to return a valid answer as the error must be raised from the relative path computation.

https://github.com/Zxun2/jupyterlab-git/blob/eee28dd958ea4d775bf11e6af5f62e72fc875e1a/jupyterlab_git/git.py#L967

Zxun2:
Hey @fcollonval, thanks for the quick response! I updated the repository to `juypter4`, hence the huge LoC changes

fcollonval:
Kicking the CI following targeted branch change

jupyterlab_git/tests/test_handlers.py
```diff
@@ -112,6 +114,52 @@ async def test_git_show_prefix(mock_execute, jp_fetch, jp_root_dir):
...
+        ]
+    )
+
+    try:
+        Path(path).absolute().relative_to(Path(local_path).absolute())
+    except ValueError:
+        # pass the test if the path is not a subdirectory of the root directory
+        pass
+    else:
+        assert False, "The path should not be a subdirectory of the root directory"
```
fcollonval, line 160: Thanks for starting the test @Zxun2 

This does not test the new case you have introduce. What should be expected is that the `payload["path"] is None`.

So the git command `--show-prefix` is returning the relative path from the current folder to the repository root.
The server root directory is given by `jp_root_dir`. Therefore if we return `jp_root_dir.name + "/"` when the endpoint call is `NAMESPACE, "show_prefix"`, your new code should be triggered.

Does that make sense?

The code would look similar to that:

```suggestion
    mock_execute.return_value = maybe_future((0, f"{jp_root_dir.name}/", ""))

    # When
    response = await jp_fetch(
        NAMESPACE,
        "show_prefix",
        body="{}",
        method="POST",
    )

    # Then
    assert response.code == 200
    payload = json.loads(response.body)
    assert payload["path"] is None

    mock_execute.assert_has_calls(
        [
            call(
                ["git", "rev-parse", "--show-prefix"],
                cwd=str(jp_root_dir),
                timeout=20,
                env=None,
                username=None,
                password=None,
                is_binary=False,
            ),
        ]
    )
```

Zxun2: I have replaced the test case. Thank you!

jupyterlab_git/tests/test_handlers.py
```diff
@@ -112,6 +114,35 @@ async def test_git_show_prefix(mock_execute, jp_fetch, jp_root_dir):
...
+    )
+    # Then
+    assert response.code == 200
+    payload = json.loads(response.body)
+    assert payload["path"] is None
+    mock_execute.assert_has_calls(
+        [
+            call(
+                ["git", "rev-parse", "--show-prefix"],
+                cwd=str(jp_root_dir),
```
fcollonval, line 135: Here we are missing the trailing slash:

```suggestion
                cwd=str(jp_root_dir) + "/",
```

jupyterlab_git/git.py
```diff
@@ -959,9 +959,26 @@ async def show_prefix(self, path):
             cmd,
             cwd=path,
         )
+
         if code == 0:
-            result = {"code": code, "path": my_output.strip("\n")}
-            return result
+            relative_git_path = my_output.strip("\n")
+            repository_path = path[: -len(relative_git_path)]
```
fcollonval, line 965: My fault but this actually does not work if `len(relative_git_path) == 0` as `"hello"[:-0] == ""`. Hence some of the test failure we are seeing. That suggestion should fix it

```suggestion
            repository_path = path[: -len(relative_git_path)] if relative_git_path else path
```

fcollonval (changes requested):
Almost through, here are some suggestions to fix the CI

Zxun2:
@fcollonval I am done with this PR!

fcollonval (approved):
Thanks @Zxun2

fcollonval:
@all-contributors please add @Zxun2 for code

Issue #1323, opened 2024-05-14
Unable to set user name and email


Zsailer:
## Description

(Huge thanks to @jzhang20133 for finding the root cause of this issue—I'm merely reporting her findings :)) 

We noticed a regression from previous versions of this package.

When initializing a repo, if the user name and email is not set, the extension no longer raises a modal prompting a user for name/email. Instead, it raises an error dialog to the user:

![image](https://github.com/jupyterlab/jupyterlab-git/assets/2791223/b76f4bd0-ddec-43ff-bc63-0c72db29ebe6)

The root of the issue appears related to related to: https://github.com/jupyterlab/jupyterlab-git/issues/1294

There is one more place where we should explicitly check if `path == null`: 
https://github.com/jupyterlab/jupyterlab-git/blob/b878c0b28376e7d847a095d6e5205596508421a5/src/components/GitPanel.tsx#L845

When we make this change, the prompt appears:

![image](https://github.com/jupyterlab/jupyterlab-git/assets/2791223/aeeff44e-fad7-4065-bd41-730b9fad859f)

## Reproduce

The easiest way to trigger this issue is to remove any git config that sets the user's email or name. Then, turn off git's "auto-detection" for user details/email.

```
git config --unset user.name
git config --unset user.email

git config --local user.useConfigOnly true
```

After adding these settings, use jupyterlab-git to make an commit and you'll see the error mentioned above.

## Expected behavior

Prompt the user for a user name and email if not set.

PR #1324, opened 2024-05-14
Unable to set email/username on a remote server

jzhang20133:
Fixes https://github.com/jupyterlab/jupyterlab-git/issues/1323

### DIFF
--- a/src/components/GitPanel.tsx
+++ b/src/components/GitPanel.tsx
@@ -842,7 +842,7 @@ export class GitPanel extends React.Component<IGitPanelProps, IGitPanelState> {
    * @param path - repository path
    */
   private async _hasIdentity(path: string | null): Promise<string | null> {
-    if (!path) {
+    if (path === null) {
       return null;
     }
 
--- a/src/__tests__/test-components/GitPanel.spec.tsx
+++ b/src/__tests__/test-components/GitPanel.spec.tsx
@@ -274,6 +274,72 @@ describe('GitPanel', () => {
       expect(commitSpy).toHaveBeenCalledWith(commitSummary, false, null);
     });
 
+    it('should prompt for user identity if user.name is not set when pathRepository is empty string', async () => {
+      props.model.pathRepository = '';
+      renderResult.rerender(<GitPanel {...props} />);
+
+      configSpy.mockImplementation(mockConfigImplementation('user.email'));
+      mockUtils.showDialog.mockResolvedValue(dialogValue);
+
+      await userEvent.type(screen.getAllByRole('textbox')[0], commitSummary);
+      await userEvent.click(screen.getByRole('button', { name: 'Commit' }));
+
+      await waitFor(() => {
+        expect(configSpy).toHaveBeenCalledTimes(2);
+      });
+      expect(configSpy.mock.calls[0]).toHaveLength(0);
+      expect(configSpy.mock.calls[1]).toEqual([commitUser]);
+      expect(commitSpy).toHaveBeenCalledTimes(1);
+      expect(commitSpy).toHaveBeenCalledWith(commitSummary, false, null);
+    });
+
+    it('should prompt for user identity if user.email is not set when pathRepository is empty string', async () => {
+      props.model.pathRepository = '';
+      renderResult.rerender(<GitPanel {...props} />);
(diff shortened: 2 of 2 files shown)


Zsailer (approved):

src/components/GitPanel.tsx
```diff
@@ -842,7 +842,7 @@ export class GitPanel extends React.Component<IGitPanelProps, IGitPanelState> {
    * @param path - repository path
    */
   private async _hasIdentity(path: string | null): Promise<string | null> {
-    if (!path) {
+    if (path == null) {
```
krassowski, line 845: The liner suggests to use triple equals here. The argument annotation says it cannot be undefined so it should be equivalent:


```suggestion
    if (path === null) {
```

Unless this fix depends on it, then we need to change linter eqeqeq rule to smart and add undefined in the argument.

Also, can you add a test?

krassowski (changes requested):

Zsailer:
Nice! Thanks @jzhang20133! 

This LGTM. @fcollonval I don't have access on this repo. Would you mind helping me merge and release a patch here? I can pick this up too if you want to give me access here.

Zsailer:
> Also, can you add a test?

@krassowski I dived into this a bit. It appears that we're skipping most unit tests for this panel since the JupyterLab 4 release. (see [here](https://github.com/jupyterlab/jupyterlab-git/blob/b878c0b28376e7d847a095d6e5205596508421a5/src/__tests__/test-components/GitPanel.spec.tsx#L179), [here](https://github.com/jupyterlab/jupyterlab-git/blob/b878c0b28376e7d847a095d6e5205596508421a5/src/__tests__/test-components/GitPanel.spec.tsx#L215), [here](https://github.com/jupyterlab/jupyterlab-git/blob/b878c0b28376e7d847a095d6e5205596508421a5/src/__tests__/test-components/GitPanel.spec.tsx#L234), [here](https://github.com/jupyterlab/jupyterlab-git/blob/b878c0b28376e7d847a095d6e5205596508421a5/src/__tests__/test-components/GitPanel.spec.tsx#L250), and [here](https://github.com/jupyterlab/jupyterlab-git/blob/b878c0b28376e7d847a095d6e5205596508421a5/src/__tests__/test-components/GitPanel.spec.tsx#L266)). I believe this is the reason the current bug wasn't detected in the first place. Removing and of the `.skip` calls causes these tests to fail, so it seems that the API changed somewhere along the way and the tests were never updated? 

I think it's beyond the scope of this current PR to fix all of these failing unit tests. For the sake of addressing the current bug, I would propose that we merge the current PR without additional unit testing. 

Further, it seems that the workflows for this repo have been failing for quite awhile due to some dependency resolution errors.

Zsailer:
I've opened the following issue to track work on re-enabling the skipped unit tests: https://github.com/jupyterlab/jupyterlab-git/issues/1325

krassowski:
> I don't have access on this repo. Would you mind helping me merge and release a patch here? I can pick this up too if you want to give me access here.

I've added you as a collaborator, but let's get the CI green before merging - thank you for opening https://github.com/jupyterlab/jupyterlab-git/pull/1326!

Zsailer:
> but let's get the CI green before merging 

Done, @krassowski 🚀 

I'll try to get the (skipped) unit tests working for the Git Panel. If I can't get that done today, I don't think we should block this PR.

Zsailer:
Amazing work, @jzhang20133! Thank you so much for taking it the extra mile to fix all unit tests and adding a few new ones for this PR! 🚀

Zsailer (approved):

Zsailer:
We'll need help with releasing this bug fix, since we don't have admin access to this repo. 

I think it would be worth updating the Github workflows for releasing to use the latest work happening in Jupyter-releaser with project level PyPI tokens.

fcollonval:
Thanks a lot @jzhang20133 @Zsailer and @krassowski for the latest work; [v0.50.1 has been released](https://github.com/jupyterlab/jupyterlab-git/releases/tag/v0.50.1) with the latest fixes.

Zsailer:
Thank you, @fcollonval! We appreciate you tremendously! :heart:

PR #1329, opened 2024-05-15
Remove hybridcontents from test dependencies.

Zsailer:
Fixes #1328.

hybridcontents is incompatible with the latest Jupyter Server (2) and JupyterLab.

PR #1327, opened 2024-05-15
Fix integration tests

krassowski:
Follow-up to https://github.com/jupyterlab/jupyterlab-git/pull/1326

PR #1330, opened 2024-05-15
Fix integration tests

Zsailer:
Working in parallel to #1327. 

* Rebased on #1329 so that the tests run fast.
* Rebased on #1327 to pick up @krassowski's changes.


Zsailer:
All green! 🎉 🎉

Zsailer:
Since everything is green here, merging to unblock #1324

Issue #1353, opened 2024-10-10
Git config dialog is only displayed once


jtpio:
## Description

If you try to create a new commit and haven't set your git config (name and email address) yet, the dialog will be displayed on commit.

If you click on "OK" and then try to commit again, the dialog will not be shown.

## Reproduce

First, make sure the git config is not set.

1. Make a change
2. Create a commit
3. Click on "OK"
4. The commit should fail
5. Try to commit again
6. The dialog is **not** displayed and the commit fails.

## Expected behavior

The dialog should be displayed the second time.

## Context

- Python package version: 0.50.1

- Extension version: 0.50.1

- Git version: 2.43.0

- Operating System and its version: All

PR #1360, opened 2024-10-22
Confirm identity validity

gjmooney:
Fix for #1353. Adds a check for the validity of the `lastAuthor` information, and if the information isn't valid the author form is displayed again.


jtpio (approved):
Thanks!

Just checked on Binder and it seems to be working well :+1: 

[jupyterlab-git-config-dialog.webm](https://github.com/user-attachments/assets/4a675ac0-b6d5-4b43-928b-01866a33746d)

Looks like the notification not going away is not related and could be tracked separately.

Issue #1354, opened 2024-10-10
Creating a new branch in a new repo (initialized from the UI) fails


jtpio:
## Description

Initializing a repo from the `jupyterlab-git` UI and trying to create a new branch fails.

## Reproduce

[jupyterlab-git-init-branch-issue.webm](https://github.com/user-attachments/assets/c5dc0446-e412-4ac8-adfa-8e825cea6385)

## Expected behavior

It should be possible to create a new branch directly after initializing the git repo.

## Context

- Python package version: 0.50.1

- Extension version: 0.50.1

- Git version: 2.43.0

- Operating System and its version:

PR #1363, opened 2024-10-22
Create new branch in new repo

gjmooney:
Fix for #1354

The issue was caused because in the command for creating a new branch `git checkout -b <branch_name> <start_point>`, git tries to resolve `<start_point>` to a commit hash, but there isn't one until a commit is made, so I added an empty commit to the `init` endpoint. 

new_branch


jtpio (approved):
Thanks!

Issue #1356, opened 2024-10-10
Clicking on the refresh icon quickly creates many notifications


jtpio:
## Description

Clicking on the refresh icon quickly creates many notifications.

## Reproduce

1. Go to the git panel
2. Click on the refresh icon multiple times
3. Many notifications are created

[jupyterlab-git-refresh-notifications.webm](https://github.com/user-attachments/assets/a55e91dc-2d4a-4d57-83e1-24e0aed9be5d)

## Expected behavior

One notification should be enough. Maybe the previous one should be disposed and a new one recreated.

## Context

- Python package version: 0.50.1

- Extension version: 0.50.1

- Git version: 2.43.0

- Operating System and its version:

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Web Browser Output</summary>
<pre>
Paste the output from your browser web console here.
</pre>
</details>

PR #1364, opened 2024-10-24
Disable refresh button while refresh is in progress

gjmooney:
Fix for #1356.

Adds a boolean to check if a refresh is currently in progress and disable the refresh button if so. 

less_spam


jtpio (approved):
Thanks!

Issue #286, opened 2018-12-04
Empty AWS CodeCommit repo fails to get upstream branch during push


jaipreet-s:
### Steps to repro
* Create an empty CodeCommit repo
* Clone locally
* Touch a file, stage, push.
* `Push failed` with an empty error message

### Details

On observing the server side logs, I saw the following logs which comes from the `get_upstream_branch` path

```
[E 21:47:22.008 NotebookApp] Uncaught exception POST /git/push?1543960041903 (10.0.1.116)
    HTTPServerRequest(protocol='https', host='10.0.46.183:8443', method='POST', uri='/git/push?1543960041903', version='HTTP/1.1', remote_ip='10.0.1.116')
    Traceback (most recent call last):
      File "/home/ec2-user/anaconda3/envs/JupyterSystemEnv/lib/python3.6/site-packages/tornado/web.py", line 1590, in _execute
        result = method(*self.path_args, **self.path_kwargs)
      File "/home/ec2-user/anaconda3/envs/JupyterSystemEnv/lib/python3.6/site-packages/jupyterlab_git/handlers.py", line 365, in post
        current_upstream_branch = self.git.get_upstream_branch(current_path, current_local_branch)                                                                                  s/JupyterSystemEnv/lib/python3.6/site-packages/jupyterlab
      File "/home/ec2-user/anaconda3/envs/JupyterSystemEnv/lib/python3.6/site-packages/jupenvs/JupyterSystemEnv/share/jupyter/labyterlab_git/git.py", line 589, in get_upstream_branch
        ' '.join(command)
    Exception: Error [fatal: ambiguous argument 'master@{upstream}': unknown revision or pkerath not in the working tree.
    Use '--' to separate paths from revisions, like this:
    'git <command> [<revision>...] -- [<file>...]'                                        twice to skip confirmation).
    ] occurred while executing [git rev-parse --abbrev-ref master@{upstream}] command to get upstream branch.
[W 21:47:22.058 NotebookApp] Unhandled error   
```

On `git status`, I see this state of the local repo, and plain `git push` works
```
On branch master
Your branch is based on 'origin/master', but the upstream is gone.
  (use "git branch --unset-upstream" to fixup)

Untracked files:
  (use "git add <file>..." to include in what will be committed)

        .ipynb_checkpoints/

nothing added to commit but untracked files present (use "git add" to track)
sh-4.2$ git push
Counting objects: 3, done.
[...]

PR #1365, opened 2024-10-24
Enable push/pull buttons for empty repos

gjmooney:
This addresses the issue mentioned in https://github.com/jupyterlab/jupyterlab-git/issues/286#issuecomment-2406107190 by updating how the `Toolbar` checks for remotes.

### DIFF
--- a/src/__tests__/test-components/Toolbar.spec.tsx
+++ b/src/__tests__/test-components/Toolbar.spec.tsx
@@ -224,11 +224,15 @@ describe('Toolbar', () => {
         />
       );
 
-      await waitFor(async () => {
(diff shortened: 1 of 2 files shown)


mikeknack86 (approved):

jtpio (approved):
Thanks!

jtpio:
When I tried to reproduce the rest of the https://github.com/jupyterlab/jupyterlab-git/issues/286, it looks like it was working fine, except for these disabled buttons.

So we should be able to close https://github.com/jupyterlab/jupyterlab-git/issues/286 as fixed by this PR :+1:

Issue #1352, opened 2024-10-10
Commit is lost when switching from the local branch to the origin branch


jtpio:
## Description

Using the branches UI to switch between branches:

![image](https://github.com/user-attachments/assets/142e6d9f-a0bc-4c7d-821e-ec38692cc127)

Seems to be reverting local changes.

## Reproduce

1. Clone a repo
2. Make a change to a file
3. Stage, commit
4. Switch to the corresponding remote branch, for example `origin/main`
5. Switch back to the local branch
6. Changes are gone (the commit disappeared)

[jupyterlab-git-branch-switch-issue.webm](https://github.com/user-attachments/assets/367437ce-ce65-4476-8382-c92fbecab71e)

## Expected behavior

I would expect to be able to temporarely switch to the remote branch, and then switch back to the local branch and see the new changes again.

## Context

- JupyterLab: 4.2.5
- Python package version: 0.50.1

- Extension version: 0.50.1

- Git version:

- Operating System and its version:

gjmooney:
So this issue is caused by the `-B` option [here](https://github.com/jupyterlab/jupyterlab-git/blob/main/jupyterlab_git/git.py#L1141). When you checkout a remote branch, you're not really checking out the remote, it creates a local copy based on the remote, and if a local copy already exists, it gets reset to the remote, hence the vanishing commit. 

I've been trying to find a fix that maintains the current behavior of creating a local copy of the remote branch, but without losing any commits. I think the best so far is doing `git checkout <local_branch_name>` when clicking the remote, which will create a local copy based on the remote if it doesn't exist, and then doing a `rebase` after the checkout to sync the local branch with the remote. Obviously this could lead to conflicts, so I'm not sure if that's the way to go, but I guess it's better than losing commits.

jtpio:
To add more context, this is what happens when using the `git` command in a terminal directly, with the local `main` branch being one commit ahead of the remote branch:

```
$ git checkout upstream/main
Note: switching to 'upstream/main'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at b09e477 Create deploy.yml (#2)
```

[...]

PR #1366, opened 2024-10-25
Add confirmation dialog when checking out remote branch

gjmooney:
Addresses #1352. 

Based on @jtpio's suggestion I added a confirmation dialog when checking out a remote branch that could potentially cause a loss of changes on a users local branch. 

confirm_checkout

### DIFF
--- a/src/components/BranchMenu.tsx
+++ b/src/components/BranchMenu.tsx
@@ -6,22 +6,23 @@ import {
 } from '@jupyterlab/apputils';
 import { TranslationBundle } from '@jupyterlab/translation';
 import { CommandRegistry } from '@lumino/commands';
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

jtpio:
Wondering if it could be worth reconsidering the use of the `-B` flag for the `git checkout` command? To be able to have a similar workflow as when using the `git` command directly (detached mode, and no commit lost).

Something we could track in a separate issue, as it may require more work and would likely change the current behavior, as discussed in #1352.

jtpio:
> Something we could track in a separate issue

Opened https://github.com/jupyterlab/jupyterlab-git/issues/1368

Issue #1355, opened 2024-10-10
Handle multiple repos and/or git modules


jtpio:
## Is your feature request related to a problem? Please describe.

JupyterLab git does not show any information about other repos listed in a  `.gitmodules` file.

For example https://github.com/denoland/deno defines a `.gitmodules` file with the following content:

```git
[submodule "tests/util/std"]
	path = tests/util/std
	url = https://github.com/denoland/deno_std
	shallow = true
[submodule "tests/wpt/suite"]
	path = tests/wpt/suite
	url = https://github.com/web-platform-tests/wpt.git
	shallow = true
[submodule "tests/node_compat/runner/suite"]
	path = tests/node_compat/runner/suite
	url = https://github.com/denoland/node_test.git
	shallow = true
[submodule "cli/bench/testdata/lsp_benchdata"]
	path = cli/bench/testdata/lsp_benchdata
	url = https://github.com/denoland/deno_lsp_benchdata.git
	shallow = true
```

If I do the following:

- `git clone https://github.com/denoland/deno`
- `cd deno`
- `git submodule init`
- `git submodule update`

The JupyterLab Git UI does not show anything about these modules from the top-level:

![image](https://github.com/user-attachments/assets/444e0af3-e7bb-4bf0-87e1-88a0efa05bee)

But if you navigate to one of the submodules then the UI detects the repo properly:

![image](https://github.com/user-attachments/assets/75d53560-de35-459d-ae45-8b983a58fc3e)

## Describe the solution you'd like

Something like what VS Code does could be useful:

![image](https://github.com/user-attachments/assets/1c2be507-bc04-4229-b3d3-22ceeba8831a)

## Describe alternatives you've considered

N/A

## Additional context

- Python package version: 0.50.1

- Extension version: 0.50.1

- Git version: 2.43.0

- Operating System and its version: All

krassowski:
@jtpio is this resolved by #1374?

PR #1374, opened 2024-10-30
Add submodules display

gjmooney:
Addresses #1355.

Adds a display for submodules in the current repo inspired by the branch menu. Currently this just displays the submodules that are present and doesn't have any functionality. I figure how to handle the submodules and which functionality should be exposed should be discussed first. 

Repo with no submodules:
no_subs

Repo with submodules - collapsed:
subs-collapse

Repo with submodules - expanded:
subs

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -2195,13 +2195,13 @@ class Git:
         results = []
 
         for line in output.splitlines():
-            name = line.strip().split(" ")[-1]
+            name = line.strip().split(" ")[1]
             submodule = {
                 "name": name,
             }
             results.append(submodule)
 
-        return {"code": code, "subModules": results, "error": error}
+        return {"code": code, "submodules": results, "error": error}
 
     @property
     def excluded_paths(self) -> List[str]:
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -2202,9 +2202,12 @@ class Git:
 
         results = []
         for prefix, name, postfix in (line.split(".") for line in output.splitlines()):
-            results.append(name)
+            submodule = {
+                "name": name,
+            }
+            results.append(submodule)
 
-        return {"code": code, "submodules": results, "error": error}
+        return {"code": code, "subModules": results, "error": error}
 
     @property
     def excluded_paths(self) -> List[str]:
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -2183,6 +2183,29 @@ class Git:
 
         return {"code": code, "message": output.strip()}
 
+    async def submodule(self, path):
+        """
+        Execute git config --file .gitmodules --name-only --get-regexp path
+        """
+
+        cmd = [
+            "git",
+            "config",
+            "--file",
+            ".gitmodules",
+            "--name-only",
+            "--get-regexp",
+            "path",
+        ]
+
+        code, output, error = await self.__execute(cmd, cwd=path)
+
+        results = []
+        for prefix, name, postfix in (line.split(".") for line in output.splitlines()):
+            results.append(name)
+
+        return {"code": code, "submodules": results, "error": error}
+
     @property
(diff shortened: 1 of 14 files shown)


jupyterlab_git/git.py
```diff
@@ -2183,6 +2183,26 @@ async def apply_stash(self, path: str, stash_index: Optional[int] = None) -> dic
...
+        """
+        Execute git submodule status --recursive
+        """
+
+        cmd = ["git", "submodule", "status", "--recursive"]
+
+        code, output, error = await self.__execute(cmd, cwd=path)
+
+        results = []
+        for sha, name, commit in (
```
jtpio, line 2196: Trying locally with https://github.com/denoland/deno, this seems to be giving the following error:

```
File "/home/jtp/tmp/jupyterlab_git-releaser-dist-321/.pixi/envs/default/lib/python3.13/site-packages/jupyterlab_git/git.py", line 2196, in submodule
        for sha, name, commit in (
            ^^^^^^^^^^^^^^^^^
    ValueError: not enough values to unpack (expected 3, got 2)
```

Which could be because the `git submodule status --recursive` command returns only two elements per line?

```git submodule status --recursive
-af4c6a1eee825f19d3b3cce74cfdd03ebe1a3b92 cli/bench/testdata/lsp_benchdata
-d12a68fc4930062c0bed26447a6b5245697e77c1 tests/node_compat/runner/suite
-c5d7930d5700dbbbedea607f9cf3e50acdc33e2d tests/util/std
-e78446e34a1921371658a5df08c71d83f50a2a2f tests/wpt/suite
```

gjmooney: So it looks like the output changes once `git submodule update` has been ran.

Output before running `git submodule update`:
```
-af4c6a1eee825f19d3b3cce74cfdd03ebe1a3b92 cli/bench/testdata/lsp_benchdata
-d12a68fc4930062c0bed26447a6b5245697e77c1 tests/node_compat/runner/suite
-c5d7930d5700dbbbedea607f9cf3e50acdc33e2d tests/util/std
-e78446e34a1921371658a5df08c71d83f50a2a2f tests/wpt/suite
```

output after:
```
 af4c6a1eee825f19d3b3cce74cfdd03ebe1a3b92 cli/bench/testdata/lsp_benchdata (af4c6a1e)
 d12a68fc4930062c0bed26447a6b5245697e77c1 tests/node_compat/runner/suite (heads/main)
 c5d7930d5700dbbbedea607f9cf3e50acdc33e2d tests/util/std (c5d7930d)
 e78446e34a1921371658a5df08c71d83f50a2a2f tests/wpt/suite (e78446e34a)
```

I just pushed a quick fix.

jtpio:
> how to handle the submodules and which functionality should be exposed should be discussed first.

Yes that makes sense :+1: This would probably require integrating other `submodule` commands such as `git submodule init` and `git submodule update`.

jupyterlab_git/handlers.py
```diff
@@ -1069,6 +1069,24 @@ async def post(self, path: str = ""):
             self.finish(json.dumps(response))
 
 
+class GitSubModulesHandler(GitHandler):
```
jtpio, line 1072: Minor nitpick: should `SubModule` be spelled `Submodule` instead?

Looks like this is what the git docs use: https://git-scm.com/book/en/v2/Git-Tools-Submodules

Wondering about this, if we start exposing classes and other interfaces which include the word "submodule" in their name.

src/__tests__/test-components/GitPanel.spec.tsx
```diff
@@ -372,6 +372,7 @@ describe('GitPanel', () => {
     beforeEach(() => {
       props.model = {
         branches: [],
+        subModules: [],
```
jtpio, line 375: Maybe we'll want to apply the casing change to other places like this one?

jtpio (approved):
Thanks!

Issue #1318, opened 2024-02-28
Branch name conflicts with existing directory name


Paul2708:
## Description

If there exists a directory with the same name as a branch, switching to this branch fails.
The following error message is shown:

![image](https://github.com/jupyterlab/jupyterlab-git/assets/12039677/1cd7ff71-9336-4896-8608-88c40e8c4925)

*Failed to switch branch: fatal: ambiguous argument 'b': both revision and filename Use '--' to separate paths from revisions, like this: 'git <command> [<revision>...] -- [<file>...]'*

## Reproduce

1. Initialize a new repository and create an initial commit (e.g., creating a README file).
2. Create a new branch "test". Now, switching branches works as expected.
3. Switch into branch "main" and create a new directory called "test".
4. Now, switching back to the branch "test" results in the error above.

## Expected behavior

Due to the error, the branch could not be switched.
Thus, the expected behavior is that switching branches works even if there is a directory with the same name.

## Context

- Python package version: `0.50.0`

- Extension version: `@jupyterlab/git v0.50.0 enabled OK (python, jupyterlab-git)`

- Git version:  `2.42.0`

- Operating System and its version: Ubuntu 20.04 (WSL)

<details><summary>Command Line Output</summary>
<pre>
[D 2024-02-28 20:40:06.211 ServerApp.jupyterlab_git] Execute ['git', 'diff', 'main', 'test', '--name-only', '-z'] in /home/paul/cammp/del-me-git-extension-test/myrepo.
[D 2024-02-28 20:40:06.216 ServerApp.jupyterlab_git] Code: 128
    Output:
    Error: fatal: ambiguous argument 'test': both revision and filename
    Use '--' to separate paths from revisions, like this:
    'git <command> [<revision>...] -- [<file>...]'

[E 2024-02-28 20:40:06.217 ServerApp] {
      "Host": "localhost:8888",
      "Accept": "*/*",
      "Referer": "http://localhost:8888/lab/tree/myrepo/Untitled.ipynb",
      "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/121.0.0.0 Safari/537.36 OPR/107.0.0.0"
    }
[...]

PR #1390, opened 2025-02-03
Fix switching branches by adding double dash separator

Paul2708:
This PR addresses #1318 by treating the parameters as Git objects and not files (e.g., [Double Hyphens in Git Diff](https://vincenttam.github.io/blog/2014/08/07/double-hyphens-in-git-diff/)). 

I'm happy to provide some tests to ensure the correctness of the functionality (testing it manually works, though).
However, the test base contains a lot of mocks. It would be awesome if you could point out some tests that actually rely on the created Git repositories and the underlying files.

### DIFF
--- a/jupyterlab_git/git.py
+++ b/jupyterlab_git/git.py
@@ -312,7 +312,7 @@ class Git:
             elif base == "INDEX":
                 cmd = ["git", "diff", "--staged", remote, "--name-only", "-z"]
             else:
-                cmd = ["git", "diff", base, remote, "--name-only", "-z"]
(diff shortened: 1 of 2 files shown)


krassowski:
@Paul2708 can you rebase on `main` to make sure tests pass?

krassowski (approved):
Thank you @Paul2708!

Verified the fix on Binder, all looks good.

krassowski:
Released in https://github.com/jupyterlab/jupyterlab-git/releases/tag/v0.51.0

Issue #1342, opened 2024-08-01
Cloning from new host via ssh causes spurious error rather than prompting for confirmation and succeeding


ctcjab:
## Description

Using jupyterlab-git to clone a repo over ssh when the server is not yet in ~/.ssh/known_hosts causes the following "failed to clone" error:
<img width="318" alt="Screenshot 2024-08-01 at 1 00 48 PM" src="https://github.com/user-attachments/assets/ebb84854-d386-42ac-a346-73779a38b4a2">
Clicking "Show" produces the following somewhat spurious error:
<img width="494" alt="Screenshot 2024-08-01 at 1 00 57 PM" src="https://github.com/user-attachments/assets/bce7a0f2-05fa-4dda-99b6-9a57ed042b54">
The failure is due to jupyterlab-git not handling the case when the host key is not yet in ~/.ssh/known_hosts, so the error message should list that as a potential cause, in addition to the user not having access rights or the repo not existing.

Better yet, jupyterlab-git should handle this case and provide a graphical UI for the user to confirm, similar to how the command line `git` client exposes this via text UI:
```
$ git clone ssh://git@bitbucket.example.com:7999/foo/bar.git
Cloning into 'bar'...
The authenticity of host '[bitbucket.example.com]:7999 ([X.Y.Z.W]:7999)' can't be established.
RSA key fingerprint is SHA256:<fingerprint>.
This key is not known by any other names
Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
Warning: Permanently added '[bitbucket.example.com]:7999' (RSA) to the list of known hosts.
remote: Enumerating objects: 18, done.
remote: Counting objects: 100% (18/18), done.
remote: Compressing objects: 100% (17/17), done.
remote: Total 18 (delta 0), reused 0 (delta 0), pack-reused 0
Receiving objects: 100% (18/18), 560.80 KiB | 24.38 MiB/s, done.
```
This way the user doesn't have to switch away from jupyterlab-git to the terminal in order to clone the repo, which some less technical users may be unable to do by themselves.

## Reproduce

1. Click "Clone a Repository"
2. Paste the URL of a git repo that is hosted via SSH and whose host's fingerprint does not appear in ~/.ssh/known_hosts 
3. Click "Clone"

## Expected behavior

[...]

PR #1408, opened 2025-05-28
Fix cloning from new host via ssh by prompting for confirmation

jesuino:
This is a possible fix for #1342

When a host is unknown then user will be prompted to add it to the know hosts file:

image

If the users clicks OK then the hostname is checked and added to the known hosts and it is proceeded by the clone operation, otherwise the clone operation will proceed and will likely fail.

Please bear in mind that I didn't check this on Windows, but I am opening the PR anyway for discussion and getting reviews because this is my first contribution for this plugin, so I may be missing some important details.

Thanks!


jupyterlab_git/ssh.py
```diff
@@ -0,0 +1,48 @@
...
+
+GIT_SSH_HOST = re.compile(r"git@(.+):.+")
+
+
+class SSH:
+    """
+    A class to perform ssh actions
+    """
+
+    def is_known_host(self, hostname):
```
jiridanek, line 19: Can this be implemented in a "ask for forgiveness, not for permission" manner? That is, run clone, and if it fails on host not known, prompt user to add host?

jupyterlab_git/ssh.py
```diff
@@ -0,0 +1,48 @@
...
+        """
+        Check if the given git clone URL contains a known host
+        """
+        cmd = ["ssh-keygen", "-F", hostname.replace(" ", "")]
```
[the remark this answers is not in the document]
jesuino: Thanks for the suggestion, changed to `.strip()`

jupyterlab_git/ssh.py
jesuino: That could be done as well, the issue for me doing this at first is to know if the error message remains consistent in all systems so I can check it on client then prompt user, hence  I did added the check before attempting to clone. Should I go ahead and do the approach you suggested anyway?

jiridanek: My reason for disliking the original implementation is that

> The host is unknown, would you like to add it to the list of known hosts?

does not say what I'm actually adding to my known hosts by clicking the button. Also it does not reveal the file location being modified, which IMO there should be, somewhere, possibly in a smaller grayer font, like a postscript, something like that.

Also, I dislike that the fingerprint is examined twice, first during the initial check and then during actual connection attempt. I do not see how this could be turned into something like time-of-check/time-of-use vulnerability, maybe it cannot, but if there is another better way found, I would prefer that one.

jesuino: >> does not say what I'm actually adding to my known hosts by clicking the button. Also it does not reveal the file location being modified, which IMO there should be, somewhere, possibly in a smaller grayer font, like a postscript, something like that.

Thanks for the suggestion, changing the dialog body would require us to create a form for the dialog. Let me know if that way is a feasible one.Or  I could also emit a notification after the host is added:

image


>> Also, I dislike that the fingerprint is examined twice, first during the initial check and then during actual connection attempt. I do not see how this could be turned into something like time-of-check/time-of-use vulnerability, maybe it cannot, but if there is another better way found, I would prefer that one.

I could simply check the `known_hosts` file and see if a fingerprint is present there for a given host, this was my first try, but I want to avoid referent to this file due possible Windows incompatible, but I ended having to refer to this file directly anyway :/  Let me know if you that's a better approach and we can switch to it. Thanks!

jiridanek: > The host %1 is not known

I love that, thanks!

jesuino:
Hello @krassowski 

Do you know if someone from `jupyterlab-git` would be able to review this PR? 

Thanks!

jupyterlab_git/handlers.py
```diff
@@ -1096,6 +1105,28 @@ async def get(self, path: str = ""):
...
+    async def get(self):
+        """
+        GET request handler, check if the host is known by SSH
+        """
+        hostname = self.get_query_argument("hostname")
+        is_known_host = self.ssh.is_known_host(hostname)
+        self.set_status(200)
+        self.finish(json.dumps(is_known_host))
+
+    @tornado.web.authenticated
```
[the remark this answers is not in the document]
jesuino: Thank you so much for your review. I added the decorator. I will be glad for any other review

jupyterlab_git/ssh.py
```diff
@@ -0,0 +1,48 @@
...
+        """
+        Check if the provided hostname is a known one
+        """
+        cmd = ["ssh-keygen", "-F", hostname.strip()]
+        try:
+            code = subprocess.call(
+                cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL
+            )
+            return code == 0
+        except subprocess.CalledProcessError as e:
```
ctcjab, line 29: I think this entire `except` block is dead code. This is the difference between `subprocess.call()` and [`subprocess.check_call()`](https://docs.python.org/3/library/subprocess.html#subprocess.check_call): The latter raises `CalledProcessError` when the exit code is nonzero, whereas the former does not.

Maybe better to actually use `check_call()` above (without capturing the return value), then return False in this `except` block, and return `True` in an [`else` block](https://docs.python.org/3/tutorial/errors.html#:~:text=has%20an%20optional%20else%20clause).

Probably also worth widening the `except` target from `CalledProcessError` to `Exception`, to catch e.g. `FileNotFoundError` in the case that there is no `ssh-keygen` on the PATH.

ctcjab (commented):
Took a quick look and caught some minor code improvement opportunities, but otherwise this looks like a great improvement, thanks @jesuino!

jupyterlab_git/ssh.py
jesuino: Thank you so much for your review and for spotting this issue, I really appreciate. I made the changes as suggested, thanks!

krassowski (approved):
Thank you!

PR #1412, opened 2025-06-11
Ignore github.com links for now, fix some other links

jtpio:
Looking into fixing the check links on CI.

Similar to https://github.com/jupyterlab/jupyterlab/pull/17544, it looks like GitHub tightened the rate limits.


jtpio:
cc @krassowski this is following the same approach as https://github.com/jupyterlab/jupyterlab/pull/17544 unfortunately.

Hopefully we can get back to this at some point as part of https://github.com/jupyterlab/jupyterlab/issues/16837.

krassowski (approved):
Thank you @jtpio!

PR #1417, opened 2025-07-16
Add environment variable inheritance to git status and branch API calls

### LINKED ISSUE #1416: Git Status and Branch APIs Don't Use Current Environment Variables - Missing os.environ.copy()
## Description

The status() and branch() methods in git.py use static environment inheritance instead of capturing current environment state, causing them to miss runtime environment variable changes that could be critical for Git operations.

## Reproduce

In `/jupyterlab_git/git.py`, the status and branch APIs execute git commands without explicitly passing environment variables:

  ```
Status API (git.py:494):
  code, status, my_error = await self.__execute(cmd, cwd=path)
```

```
Branch APIs (git.py:842, git.py:908):
### In branch_heads() and branch_remotes()
code, heads, error = await self.__execute(cmd, cwd=path)
```

 ```
When these calls reach call_subprocess() (git.py:151):
process = subprocess.Popen(
    cmdline, stdout=subprocess.PIPE, stderr=subprocess.PIPE, cwd=cwd, env=env
)
```
Since env=None, the subprocess inherits the static environment from server startup time, not the current os.environ state.

Current Broken Behavior:

#### Later during runtime:
```
os.environ['GIT_CONFIG'] = '/custom/.gitconfig'
os.environ['GIT_SSH_COMMAND'] = 'ssh -i /custom/key'
os.environ['GIT_TRACE'] = '1'  # For debugging
```

#### These environment changes are IGNORED by:
```
git_instance.status()   # Still uses startup environment
git_instance.branch()   # Still uses startup environment
```

## Expected behavior
Both status and branch APIs should start with a clean environment like other APIs
```
env = os.environ.copy()
code, status, my_error = await self.__execute(cmd, cwd=path, env=env)
```

nsingl00:
Several git subprocess calls in the status and branch APIs were not inheriting the parent process environment variables, which could cause git commands to fail when custom environment configurations are needed (e.g., custom PATH, GIT_*
  variables, proxy settings, or authentication tokens).

Added env=os.environ.copy() parameter to the following subprocess calls:

-   Status API (git.py:status() method):
-   Branch API:

Issue: https://github.com/jupyterlab/jupyterlab-git/issues/1416

### DIFF
--- a/packages/core/jupyterlab_git_core/git.py
+++ b/packages/core/jupyterlab_git_core/git.py
@@ -522,7 +522,7 @@ class Git:
         Execute git status command & return the result.
         """
         cmd = ["git", "status", "--porcelain", "-b", "-u", "-z"]
-        code, status, my_error = await self.__execute(cmd, cwd=path)
+        code, status, my_error = await self.__execute(cmd, cwd=path, env=os.environ.copy())
 
         if code != 0:
             return {
@@ -877,7 +877,7 @@ class Git:
             "refs/heads/",
         ]
 
-        code, output, error = await self.__execute(cmd, cwd=path)
+        code, output, error = await self.__execute(cmd, cwd=path, env=os.environ.copy())
         if code != 0:
             return {"code": code, "command": " ".join(cmd), "message": error}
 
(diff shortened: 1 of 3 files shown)


krassowski (approved):
In general the pattern we use in the codebase is:

```python
env = os.environ.copy()
env["GIT_TERMINAL_PROMPT"] = "0"
```

but I think these commands should not prompt so I think it is fine to merge as-is.

nsingl00:
@jtpio @krassowski @Zsailer Could anyone help merge this PR for me?

Zsailer (approved):
Thanks @nsingl00!

PR #1453, opened 2026-03-26
refactor: remove Tornado dependency from git.py

djangoliv:
Remove Tornado dependency from git.py

Replaced tornado.web.HTTPError with a custom GitError and switched IOLoop calls to asyncio.

This decouples git.py from Tornado, making it framework-agnostic, easier to test, and keeps HTTP handling in the handlers. Behavior for handlers remains unchanged.


davidbrochart (commented):
If we want to make it framework-agnostic, we could also use AnyIO instead of asyncio?

packages/core/jupyterlab_git_core/git.py
jtpio: Should this use the following?

```suggestion
        code, output, error = await self.__execute(cmd, cwd=path)
```

jtpio: Similar comment as above:


```suggestion
        code, output, error = await self.__execute(cmd, cwd=path, env=env)
```

packages/jupyterlab/pyproject.toml
```diff
@@ -0,0 +1,70 @@
...
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
+]
+dependencies = [
+    "jupyterlab-git-core[nbdime]",
```
jtpio: Since `jupyterlab-git` imports from `jupyterlab-git-core`, should this dependency have bounds?

README.md
jtpio, line 70: Should these instructions match the install instructions above?

jtpio (commented):
Thanks @djangoliv for working on this!

Had a first pass on the diff and left a couple of comments.

We should probably also check the releaser properly bumps the version of the two Python packages and the npm package.

packages/core/jupyterlab_git_core/git.py
djangoliv: Good catch! These should indeed use self.__execute to benefit from the timeout and execution lock mechanism. I'll apply both suggestions.

packages/jupyterlab/pyproject.toml
djangoliv: Since both packages live in the same monorepo and are released together, they will always be in sync.   
 However, we could add a lower bound like jupyterlab-git-core[nbdime]>=<current_version> to be safe.

README.md
djangoliv: Good point! The uninstall section should cover both installation modes. I'll add instructions for       uninstalling jupyterlab-git-core and fps-jupyterlab-git in the jupyverse case.

djangoliv:
> We should probably also check the releaser properly bumps the version of the two Python packages and the npm package.

 I verified this locally by simulating what jupyter_releaser does under the hood:
                                                                                                   
  # Python packages (what jupyter_releaser calls via version_cmd = "hatch version")
  cd packages/core && hatch version 0.52.1                
  cd packages/jupyterlab && hatch version 0.52.1 
  npm version 0.52.1 --no-git-tag-version                            
                                                                                                            
  All three files were correctly bumped:
  - packages/core/package.json (@jupyterlab/git-core): 0.52.0 → 0.52.1 ✓                                    
  - packages/jupyterlab/package.json (@jupyterlab/git): 0.52.0 → 0.52.1 ✓                                   
  - package.json (npm): 0.52.0 → 0.52.1 ✓
   
@jtpio Do you have another way to verify this in mind ?

jtpio:
Looking at the artifacts built by the releaser on CI: https://github.com/jupyterlab/jupyterlab-git/actions/runs/24345435413?pr=1453

It seems that one of them (the npm package for the JupyterLab frontend extension) does not get the same version:

image

djangoliv:
@jtpio  Thanks for catching this! Fixed in the latest commit : https://github.com/jupyterlab/jupyterlab-git/pull/1453/changes/1c7fc2548447e2c94e8f1ba6a82484cf042a22f9

packages/jupyterlab/pyproject.toml
jtpio: A lower bound would probably make sense yes 👍 

In case `jupyterlab_git_core` changes some of the modules and functions that `jupyterlab_git` relies on.

packages/core/install.json
```diff
@@ -0,0 +1,5 @@
+{
+  "packageManager": "python",
+  "packageName": "jupyterlab-git-core",
+  "uninstallInstructions": "Use your Python package manager (pip, conda, etc.) to uninstall the package jupyterlab-git-core"
```
jtpio, line 4: Wondering if the fact that `jupyterlab-git-core` now "owns" the JupyterLab extension may create some issues with how packages are installed, especially with existing installations of `jupyterlab-git`?

Just thinking out loud here. Maybe highlighting this packaging change in the README could help users handle potential issues, even though the upgrade path will likely handle things properly.

packages/jupyterlab/pyproject.toml
jtpio: Although on second thought, adding such lower bound may also increase the complexity of the version bump script, as we would likely want to keep that lower bound in sync with the version of the `jupyterlab-git-core` package.

packages/core/install.json
djangoliv: I had a note in the README

.github/workflows/check-release.yml
```diff
@@ -13,11 +13,18 @@ jobs:
         uses: actions/checkout@v4
       - name: Base Setup
         uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
+      - name: Install workspace deps
```
jtpio: Oh just noticed this.

We should likely avoid custom steps in this workflow, because they will not run when triggering the proper workflows for release: https://github.com/jupyterlab/jupyterlab-git/blob/main/.github/workflows/prep-release.yml

And instead have the necessary steps be specified as part of the jupyter-releaser config if needed, so they run in both workflows. Although the multi python packages should be supported normally.

jtpio: @djangoliv if you want I can take a look and push to the branch

djangoliv: Good point! The custom install step was redundant since python-packages in pyproject.toml already tells 
 jupyter-releaser how to handle both packages. Removed it to keep both workflows consistent.

jtpio (approved):
Thanks!

I'll follow-up with a `0.53.0a0` pre-release first, and will do the final release after if everything goes well.

PR #1454, opened 2026-03-30
refactor: remove unused GitAllHistoryHandler endpoint

djangoliv:
The GitAllHistoryHandler and related frontend methods/types are no longer used (see e7ab244). This cleans up dead code from backend handlers, tests, and TypeScript interfaces.


jtpio (approved):
Thanks!

For consistency, maybe this section could also be removed?

https://github.com/jupyterlab/jupyterlab-git/blob/28eb84383b4ebbb26346066828a71f84149a079e/specification/Git_REST_API.yml#L33

PR #1460, opened 2026-04-27
fix: read version from package.json instead of hatch at workspace root

djangoliv:
hatch version fails at the uv workspace root (no build backend) causing jupyter-releaser to fall back to 0.0.0 for the release tag.
Add a minimal hatchling + hatch-nodejs-version config at the root so hatch version works from there, reading the version from package.json.
The root package is not published; python-packages controls what gets released.


jtpio (approved):
Thanks, let's try it!

PR #1461, opened 2026-04-27
Fix check release

jtpio:
- Remove the custom logic in the check release workflow to avoid differences in behavior with the actual release workflows.
- Skip `check-python` for now

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,9 @@ addopts = "--import-mode=importlib"
 [tool.check-wheel-contents]
 ignore = ["W002"]
 
+[tool.jupyter-releaser]
+skip = ["check-python"]
+
 [tool.jupyter-releaser.options]
 version_cmd = "hatch version"
 python-packages = ["packages/core", "packages/jupyterlab"]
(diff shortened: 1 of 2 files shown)


jtpio:
OK it's failing as expected now.

pyproject.toml
```diff
@@ -28,6 +28,9 @@ addopts = "--import-mode=importlib"
 [tool.check-wheel-contents]
 ignore = ["W002"]
 
+[tool.jupyter-releaser]
+skip = ["check-python"]
```
jtpio, line 32: Needed for now, tracked in https://github.com/jupyter-server/jupyter_releaser/issues/499

djangoliv: You will revert this change after the first release of jupyterlab-git-core, right ?

jtpio: Maybe, but not sure it will help since the releaser will then likely pull `jupyterlab-git-core` from PyPI and not use the local version.

jtpio: In other words we should look into fixing the Python monorepo use case in `jupyter-releaser` directly.

PR #1462, opened 2026-04-27
Avoid 500 errors when the repo does not have commits yet

jtpio:
The server would otherwise log the following error when the git repo exists but does not have any commit yet:

```
[E 2026-04-27 14:48:05.973 ServerApp] 500 POST /git/log?1777294085878
```


djangoliv:
LGTM. Thanks @jtpio

PR #1465, opened 2026-05-04
Move the diff functionality to separate plugins

jtpio:
This allows other extensions to more easily swap how diffs are being rendered, while keeping the functionalities from the base git extension.


krassowski (approved):
LGTM, thanks!

PR #1466, opened 2026-05-05
Fix root `pyproject.toml`

jtpio:
Follow-up to #1453 

Since `uv` is mentioned in https://github.com/jupyterlab/jupyterlab-git#development-install


jtpio:
cc @djangoliv if you want to have a quick look

djangoliv:
Perfect ! Thanks @jtpio

PR #1477, opened 2026-05-17
Add inline edit for diffs

### LINKED ISSUE #1464: Feature Request: Add inline editing to the Git diff view (VS Code-style)
## Is your feature request related to a problem? Please describe.
When reviewing file diffs (Working Tree vs. HEAD) within the `jupyterlab-git` extension, the diff view is currently read-only. To make changes or correct mistakes, I am forced to switch to the standard file editor, manually locate the relevant line numbers, and edit the file there. This breaks the workflow, requires context switching between tabs, and is inefficient for quickly iterating on changes while reviewing the diff. It is difficult to maintain focus on the Git view when minor edits are needed.

## Describe the solution you'd like
I would like to see an **Editable Diff Mode** introduced, similar to the functionality available in VS Code and VSCodium.

Specifically:
- When viewing the diff of a file, there should be a way to enable an "Edit Mode" or toggle that makes the diff content interactive.
- In this mode, users should be able to click directly into the diff lines, modify the text inline, and have those changes immediately reflected in the underlying file.
- This allows for quick corrections and iterations without leaving the Git tab or losing the visual context of the diff.
- Ideally, this mirrors the VS Code experience where the diff view supports inline editing, making the Git integration feel seamless and powerful within JupyterLab.

## Describe alternatives you've considered
- **Manual Navigation:** Currently, I have to open the file in the Jupyter editor and navigate to the specific line numbers shown in the diff. This is error-prone and slow.
- **External Tools:** Using an external Git GUI or command-line tools to make changes, which defeats the purpose of having an integrated extension in JupyterLab.
- **"Open at Line" Button:** A button that opens the file at the diff line number would be a minor improvement, but it still requires switching tabs and does not provide the seamless inline editing experience that keeps the workflow contained.

MUFFANUJ:
Closes #1464

This change adds inline editing support for WORKING vs HEAD plain text diffs and keeps the UX stable by saving edits with a 1-second debounce, refreshing diff content safely, and restoring cursor/focus after refresh. It also adds focused tests for edit button visibility, save behavior, safe refresh handling, selection restoration, and dispose-time save to ensure the flow is deterministic and reliable.


MUFFANUJ:
Hmm, so right now if the file is already open it does not reflect the edit mode changes

MUFFANUJ:
For quick look up:


https://github.com/user-attachments/assets/7d3eacf3-c40b-4fdc-a5f2-437c19572d90

MUFFANUJ:
So earlier, when we had unsaved changes, and we used the edit mode to make changes in the same file with unsaved changes, it overwrote the changes. I feel this is maybe not something a user would want ( to lose their changes ). So, we have now added a notification `Inline edit is unavailable while this file has unsaved changes in another editor. `


https://github.com/user-attachments/assets/30ebfad8-cab0-4d05-9e66-1c01d8118174

PR #1517, opened 2026-08-05
Improve Git panel UX

### LINKED ISSUE #1513: Move commit box to the top of the panel
Currently the input areas to specify the commit message and description are located at the bottom of the panel, while the other buttons to pull, push and refresh are at the top, which does not make it practical to use.

Also the text area for the optional description takes quite a bit of space. Maybe it would make sense to just have a single multi-line text area for both the commit message and description instead?

Image

jtpio:
Fixes https://github.com/jupyterlab/jupyterlab-git/issues/1513

- [x] Move the commit box to the top of the panel
- [x] Keep only one input since it can be used to specify both the message and the description
- [x] Make the toolbar configurable via the settings, so items could be reordered or even removed (for example hiding the branch)
- [x] Update tests

| Before | After |
| :---: | :---: |
| before  | after |

**Toolbar changes**

https://github.com/user-attachments/assets/d3ed6955-e44a-4b9c-9180-e8f084d3e727

**Commit box hint**

Show a hint when there are multiple lines, since the box now combines the message and the description.

image

### DIFF
--- a/schema/plugin.json
+++ b/schema/plugin.json
@@ -106,6 +106,71 @@
       "title": "Clear outputs before commit",
       "description": "If true, notebook outputs will be cleared before committing. If false, outputs are kept. If null, ask each time.",
       "default": null
+    },
+    "toolbar": {
+      "title": "Git panel toolbar items",
+      "description": "Note: To disable a toolbar item,\ncopy it to User Preferences and add the\n\"disabled\" key. The following example will disable the refresh button:\n{\n  \"toolbar\": [\n    {\n      \"name\": \"refresh\",\n      \"disabled\": true\n    }\n  ]\n}\n\nToolbar description:",
+      "items": {
+        "$ref": "#/definitions/toolbarItem"
+      },
+      "type": "array",
+      "default": []
+    }
+  },
+  "definitions": {
+    "toolbarItem": {
+      "properties": {
+        "name": {
+          "title": "Unique name",
+          "type": "string"
+        },
+        "args": {
+          "title": "Command arguments",
+          "type": "object"
+        },
+        "command": {
+          "title": "Command id",
+          "type": "string",
+          "default": ""
+        },
+        "disabled": {
+          "title": "Whether the item is ignored or not",
+          "type": "boolean",
+          "default": false
+        },
+        "icon": {
+          "title": "Item icon id",
+          "description": "If defined, it will override the command icon",
+          "type": "string"
+        },
+        "label": {
+          "title": "Item label",
+          "description": "If defined, it will override the command label",
+          "type": "string"
+        },
+        "caption": {
+          "title": "Item caption",
+          "description": "If defined, it will override the command caption",
+          "type": "string"
+        },
+        "type": {
+          "title": "Item type",
+          "type": "string",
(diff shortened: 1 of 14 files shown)


krassowski:
Is it easy to make the commit box position customizable? If not, I am not requesting it, just noting that I prefer it at the bottom. It is closer to my work area which is generally at the bottom of the screen (as is the last cell in the notebook).

nakul-py:
Now UI looks quite good ❤️ and i think familiar to more users as well.

 > Is it easy to make the commit box position customizable? If not, I am not requesting it, 

Yes there will be an option in future somehow if we can make it draggable in sidebar.

jtpio:
The last commit introduces a new setting to set the position:

https://github.com/user-attachments/assets/5948ab95-46bf-4a93-8c37-9bd1796281c8

Should we keep the previous default and have the commit box at the bottom? It would be fine if it's what most users prefer. Otherwise also happy to take the opportunity to move it to the top by default (current state of the PR).

krassowski:
Thank you1 I think changing the default to top is fine if that the more common convention.

krassowski (approved):
Looks fine. I played around on Binder and feed that through Opus for review, I think one user-facing find might be worth addressing:

A message whose first line is empty leaves the box visibly full of text while the Commit button is disabled - a dedicated message would be nice

Image

Image


Some nitpicks from AI review for reference:
- Toolbar.tsx:284 - `hasUpstream = activeBranch[0]?.upstream !== null` is true when there is no current-branch entry, since optional chaining yields undefined and `undefined !== null`. On a detached HEAD, or before `refreshBranch()` populates `model.branches`, the push button reads "Push committed changes" instead of "Publish branch" and the ahead badge is hidden. Needs `!= null`
- `commitFiles` unconditionally injects a blank line after the first line, so the message git records differs from the message the box shows
- GitWidget.tsx:82 - the submodule menu can be stranded open with no control able to dismiss it
- Toolbar.ts:9 `gap: '4px'` on the toolbar host has no effect
- item names are registered unprefixed into an application-global registry, and `factoryAdded` matches by item name only. Core ships items literally named refresh in filebrowser-extension and htmlviewer-extension, so registering Git's items tears down and rebuilds those toolbars' refresh buttons as a side effect of activation
- first line is the summary" is implemented twice: `commitFiles` splits and re-joins, `CommitBox._summary` does `split('\n', 1)[0]` and drives both `_canCommit()` and the error border. `src/utils.ts` already exists for this kind of parsing

jtpio:
> A message whose first line is empty leaves the box visibly full of text while the Commit button is disabled - a dedicated message would be nice

Ah good catch. After https://github.com/jupyterlab/jupyterlab-git/pull/1517/commits/425cb11385c85ca9db77657e79ac5d12cb35fa26 this should now show as the following:

image

For the other nitpicks maybe we can track them in a separate issue and have a look at them individually or all at once?

jtpio:
Just fixed the conflicts, but otherwise it looks like it's ready to go when CI passes.

Opened https://github.com/jupyterlab/jupyterlab-git/issues/1536 to potentially look into addressing the points raised above.

PR #1527, opened 2026-08-23
Add UI branch comparison

### LINKED ISSUE #1512: Ability to diff branches
Currently it does not seem to be possible to diff branches via the UI directly.

However the API may already support that via refs.

A typical use case would be to check all the changes against `main` while working on a feature branch, without having to check the diffs for each of the underlying commits.

MUFFANUJ:
Closes #1512 

This PR adds a UI action to compare any non-current branch against the current branch from the Branches and Tags section. It addresses #1512 by letting users inspect all changes between, for example, main and a feature branch without checking each commit individually.

The implementation reuses the existing diff API, which already accepts Git refs. `CommitComparisonBox` now works with generic Git refs instead of commit-only objects, branch rows expose a “Compare with current branch” diff action, and GitPanel renders the resulting changed-file summary below the branch list. Existing commit comparison behaviour is preserved by converting commits into the same ref-based shape.

Tests were added and updated for the branch compare button, GitPanel branch comparison flow, and generic ref diff handling in `CommitComparisonBox`.

### DIFF
--- a/src/__tests__/test-components/BranchMenu.spec.tsx
+++ b/src/__tests__/test-components/BranchMenu.spec.tsx
@@ -101,6 +101,7 @@ describe('BranchMenu', () => {
         execute: jest.fn()
       } as any,
       trans: trans,
+      onCompareWithCurrent: jest.fn(),
       ...props
     };
   }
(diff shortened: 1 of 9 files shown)


MUFFANUJ:
While implementing branch comparison, I noticed a separate case in the existing backend diff parsing.

The new UI reuses the existing `model.diff(base, target)` path, which already supports Git refs such as commits and branches. That part works for normal branch diffs.

However, the backend parser in `packages/core/jupyterlab_git_core/git.py` currently parses `git diff --numstat -z` output by splitting each entry on whitespace. That can misread filenames that contain spaces, and it may also mishandle renamed files because Git emits extra path information for renames.

This is not caused by the branch-comparison UI change, but the new feature may make the existing parser limitation easier to hit when comparing whole branches.

Would you prefer that I keep this PR scoped to exposing branch comparison in the UI, or should I also include a backend parser fix for paths with spaces / renamed files here?

PR #1534, opened 2026-09-16
Skip check release when running CI on forks

telamonian:
Small bug fix. When I just now updated the main branch of my fork to match upstream, after the push it started running a bunch of new CI (fine) and `check-release` (not so fine). I was scared for a sec that it would automatically create at least some kind of artifact that it shouldn't be, but thankfully it failed on

```
ValueError: package.json for '@jupyterlab/git' does not define a 'repository.url' matching the cloned repository 'telamonian/jupyterlab-git'.
```

This PR stops the `check-release` action from running on forks at all in the first place. For proof that this change actually does skip `check-release` on forks, see https://github.com/telamonian/jupyterlab-git/actions/runs/35066171431


krassowski (approved):
Thanks!

It should probably be done this way in https://github.com/jupyterlab/extension-template too.

PR #1535, opened 2026-09-16
Update `jupyter/eslint-plugin`, improve initial load performance

krassowski:
## References

- Sibling to https://github.com/jupyterlab/jupyterlab/pull/19751
- Enables `jupyter/prefer-lazy-imports` rule from `@jupyter/eslint-plugin` 1.2 and tackles easy wins

## Code changes

- Bumps `@jupyter/eslint-plugin` to 1.2 and enables `jupyter/prefer-lazy-imports` as an error; the modules needed while the application starts are listed under `ignoreImports` with the reason.
- The Git sidebar builds its toolbar and its Changes, History and Branches sections the first time it is shown, so the panel components and their MUI and react-window dependencies stay out of the startup bundle.
- The image diff module is loaded the first time an image diff is opened.

Extension JavaScript loaded at startup, production build:

| | Before | After |
|---|---|---|
| Extension JS fetched before `restored` | 654 KB in 15 files | 135 KB in 5 files |
| 4G (10 Mbps, 50 ms latency) | 6,208, 6,250, 6,386 ms | 6,126, 6,012, 6,041 ms (82 to 344 ms less) |
| Slow 3G (400 kbit/s, 2 s latency) | 148.7, 148.8, 148.9 s | 138.2, 138.1, 137.9 s (10.6 to 11.0 s less) |

Note: the totals include time of the JupyterLab core load before optimizations from https://github.com/jupyterlab/jupyterlab/pull/19751.

## User-facing changes

The sidebar content is built when the panel is first shown instead of at startup.

## Backwards-incompatible changes

None.

## AI usage

- **Yes**: Some or all of the content of this PR was generated by AI.
- **Yes**: The human author has carefully reviewed this PR and run this code (keep this PR "draft" until the answer is YES)
- AI tools and models used: Fable 5.1


src/index.ts
```diff
@@ -24,7 +24,6 @@ import {
   addFileBrowserContextMenu,
   createGitMenu
 } from './commandsAndMenu';
-import { createImageDiff } from './components/diff/ImageDiff';
 import { createNotebookDiff } from './components/diff/NotebookDiff';
 import { createPlainTextDiff } from './components/diff/PlainTextDiff';
```
krassowski: We could get down from 381 KB to 254 KB by doing the same with `createNotebookDiff` and `createPlainTextDiff`; the problem is that they get exported just below:

https://github.com/jupyterlab/jupyterlab-git/blob/d5fb6a7e4971d1d8db73a3a0cb07af15324274a8/src/index.ts#L36-L38

We could still export the types.

krassowski: The only public use (on GitHub) of the exported class (not type) is in [`jupyterlab/pull-requests`](https://github.com/jupyterlab/pull-requests) and that is vesion-pinned to an older version so just a task to handle on upgrade.

Deep imports would continue working so it will still be importable from `@jupyterlab/git/lib/components/diff/NotebookDiff`.

Thoughts?

krassowski: Based on the above I am proposing inclusion of that here too.

src/index.ts
```diff
@@ -24,19 +24,15 @@ import {
...
 import { getServerSettings } from './server';
 import { gitIcon } from './style/icons';
 import { CommandIDs, Git, IGitExtension } from './tokens';
 import { GitWidget } from './widgets/GitWidget';
 
 export { DiffModel } from './components/diff/model';
-export { NotebookDiff } from './components/diff/NotebookDiff';
-export { PlainTextDiff } from './components/diff/PlainTextDiff';
+export type { NotebookDiff } from './components/diff/NotebookDiff';
+export type { PlainTextDiff } from './components/diff/PlainTextDiff';
```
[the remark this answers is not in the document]
jtpio: Discussed above

src/index.ts
jtpio: Not sure about the state of `jupyterlab/pull-requests` and if it will be updated at some point, so probably such change is fine if that's the only known extension impacted.

src/index.ts
```diff
@@ -332,9 +333,19 @@ async function activate(
...
+    // status round trip.
     if (statusBar) {
-      addStatusBarWidget(statusBar, gitExtension, settings, trans);
+      const pluginSettings = settings;
+      app.restored
+        .then(async () => {
+          const { addStatusBarWidget } = await import(
+            './components/StatusWidget'
+          );
+          addStatusBarWidget(statusBar, gitExtension, pluginSettings, trans);
```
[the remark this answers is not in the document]
krassowski: Addressed in 8113a26

jtpio (approved):
Thanks!

